PR #38, opened 2025-12-26
fix(appkit): update llms.txt file

MarioCadenas:
Update llms.txt and setup script


ditadi (approved):

MarioCadenas:
> Do we need the CSS on llms.txt? I also added it in the beginning, but removed it later

I found interesting while creating apps that it does change the theming, so I guess adding some instructions on how it should do it should be good?

PR #39, opened 2025-12-27
fix(appkit): obo logic and api usage

MarioCadenas:
### Summary
Introduces a clean asUser(req) API for executing Databricks operations with user credentials. Replaces the middleware-based context management with a declarative pattern.

### What Changed
**New**: plugin.asUser(req).method()` pattern for user-scoped execution.

**New**: `ServiceContext` singleton initialized at app startup.

**New**: `ExecutionContext` using `AsyncLocalStorage` for transparent context propagation

**Removed**: `databricksClientMiddleware` (no longer needed)

**Removed**: `requiresDatabricksClient` plugin flag

### Architecture

```
┌─────────────────────────────────────────────────────────────┐
│                      App Startup                            │
│  • createApp() → ServiceContext.initialize()                  │
│  • Creates service principal WorkspaceClient                │
│  • Fetches warehouseId, workspaceId                         │
└─────────────────────────────────────────────────────────────┘
                              │
                              ▼
┌─────────────────────────────────────────────────────────────┐
│                     Request Handling                        │
├─────────────────────────────────────────────────────────────┤
│                                                             │
│  Service Principal Path       User Context Path             │
│  ───────────────────────      ─────────────────────────     │
│                                                             │
│  analytics.query(...)         analytics.asUser(req)         │
│         │                            │                      │
│         ▼                            ▼                      │
│  ServiceContext.get()         Creates IUserContext from     │
│         │                     x-forwarded-access-token      │
│         ▼                            │                      │
│  Uses service principal       Returns Proxy that wraps      │
│  WorkspaceClient              methods in AsyncLocalStorage  │
│                                      │                      │
│                                      ▼                      │
│                               .query(...) executes with     │
│                               user's WorkspaceClient        │
│                                                             │
└─────────────────────────────────────────────────────────────┘
```

### Usage
```typescript
// Service principal (default)
const data = await analytics.query("SELECT ...");

// User context (on behalf of user)
const userData = await analytics.asUser(req).query("SELECT ...");
```

### Frontend

Charts/tables now support `asUser` prop:

```typescript
<LineChart queryKey="spend" parameters={params} asUser />
```

### DIFF
--- a/packages/appkit/src/index.ts
+++ b/packages/appkit/src/index.ts
@@ -1,6 +1,7 @@
 // Types from shared
 export type {
   BasePluginConfig,
+  CacheConfig,
   IAppRouter,
   StreamExecutionSettings,
 } from "shared";
@@ -19,9 +20,11 @@ export { CacheManager } from "./cache";
 export {
   SeverityNumber,
   SpanStatusCode,
+  type TelemetryConfig,
   type Counter,
   type Histogram,
   type Span,
+  type ITelemetry,
 } from "./telemetry";
 
 // Vite plugin
--- a/docs/docs/api/appkit/Function.createApp.md
+++ b/docs/docs/api/appkit/Function.createApp.md
@@ -22,10 +22,10 @@ Bootstraps AppKit with the provided configuration.
 
 | Parameter | Type |
 | ------ | ------ |
-| `config` | \{ `cache?`: `CacheConfig`; `plugins?`: `T`; `telemetry?`: `TelemetryConfig`; \} |
-| `config.cache?` | `CacheConfig` |
+| `config` | \{ `cache?`: [`CacheConfig`](Interface.CacheConfig.md); `plugins?`: `T`; `telemetry?`: [`TelemetryConfig`](Interface.TelemetryConfig.md); \} |
+| `config.cache?` | [`CacheConfig`](Interface.CacheConfig.md) |
 | `config.plugins?` | `T` |
-| `config.telemetry?` | `TelemetryConfig` |
+| `config.telemetry?` | [`TelemetryConfig`](Interface.TelemetryConfig.md) |
 
 ## Returns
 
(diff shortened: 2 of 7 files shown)


packages/appkit/src/context/service-context.ts
```diff
@@ -0,0 +1,283 @@
...
+ * Execution context can be either service or user context.
+ */
+export type ExecutionContext = IServiceContext | IUserContext;
+
+/**
+ * Check if an execution context is a user context.
+ */
+export function isUserContext(ctx: ExecutionContext): ctx is IUserContext {
+  return "isUserContext" in ctx && ctx.isUserContext === true;
+}
```
pkosiec: Shouldn't we move those types, functions and also interfaces above the highlighted snippet to the `execution-context.ts`?

Actually, this is a general comment to the `service-context.ts`. IMO there is some mixed stuff which isn't about service context at all, but more about user context or `execution-context` 🤔 I'd suggest extracting those to `execution-context.ts` and maybe even create `user-context.ts` file too.

packages/appkit/src/context/service-context.ts
```diff
@@ -0,0 +1,283 @@
...
+
+function getClientOptions(): ClientOptions {
+  const isDev = process.env.NODE_ENV === "development";
+  const normalizedVersion = productVersion
```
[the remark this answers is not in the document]
MarioCadenas: good catch

packages/appkit/src/index.ts
```diff
@@ -1,27 +1,28 @@
...
+// Telemetry (for advanced custom telemetry)
 export {
+  SeverityNumber,
+  SpanStatusCode,
   type Counter,
   type Histogram,
-  SeverityNumber,
   type Span,
-  SpanStatusCode,
-  TelemetryConfig,
```
pkosiec: Let's keep the export as it is used for the `createApp` function - then the docs nicely link to it:

image

I'd also keep the `ITelemetry` export (as it contains user-facing api for telemetry handling as a part of the plugin but we should rename it 🤔 )

PR #26, opened 2025-12-18
docs: generate API reference docs from TS packages

pkosiec:
https://databricks.atlassian.net/browse/LKB-7605

API reference docs are generated on `pnpm docs:dev` or `pnpm docs:build`.

## Demo

https://github.com/user-attachments/assets/2c3b551d-da93-4126-91b0-2afb62d4c834

PR #41, opened 2026-01-07
docs: generate API reference docs from TS packages

pkosiec:
https://databricks.atlassian.net/browse/LKB-7605

API reference docs are generated on `pnpm docs:dev` or `pnpm docs:build`.

Related PR with previous conversation: https://github.com/databricks/appkit/pull/26 

## Demo

Outdated:
https://github.com/user-attachments/assets/2c3b551d-da93-4126-91b0-2afb62d4c834

Current:
Screenshot 2026-01-07 at 15 55 17
Screenshot 2026-01-07 at 15 55 22


docs/docs/api/appkit/Variable.sql.md
pkosiec: In theory we could postprocess those "variables" and rename it with https://typedoc-plugin-markdown.org/docs/options/display#pagetitletemplates but not sure if it is worth to do it at this point. After a second thought, the "Helper" name is too wide and confusing as well.

Maybe we should convert `sql` constant to a class instead?

PR #61, opened 2026-01-22
feat: add .obo conventions on sql files

ditadi:
## Summary

- Replace `asUser` API option with `.obo.sql` file naming convention for user-scoped query execution
- Simplify analytics endpoints from two routes to one unified route
- Remove `asUser` prop from all UI components (hooks, charts, tables)

### Convention-based execution context

Query execution context is now determined by file naming:
- `config/queries/<query_key>.sql` → Executes as service principal (shared cache)
- `config/queries/<query_key>.obo.sql` → Executes as user (OBO = On-Behalf-Of, per-user cache)

This is more declarative than passing `asUser: true` through the UI layer, as the query file itself declares its execution semantics.

### API changes

**Removed:**
- `POST /api/analytics/users/me/query/:query_key`
- `GET /api/analytics/users/me/arrow-result/:jobId`
- `asUser` option from `useAnalyticsQuery`, `useChartData`, `TableWrapper`, and chart components

**Unchanged:**
- `POST /api/analytics/query/:query_key` (now handles both contexts)
- `GET /api/analytics/arrow-result/:jobId`
- `Plugin.asUser(req)` method (still available for custom plugins)

### Cache behavior

- `.sql` queries use global cache key (shared across all users)
- `.obo.sql` queries use per-user cache key (isolated per user)

## Test plan

- [x] Verify `.sql` queries execute as service principal
- [x] Verify `.obo.sql` queries execute as user
- [x] Verify cache isolation between `.sql` and `.obo.sql` queries
- [x] Verify global cache sharing for `.sql` queries across users
- [x] Verify per-user cache isolation for `.obo.sql` queries
- [x] Run `pnpm typecheck` - no type errors
- [x] Run `pnpm test` - all tests pass

### DIFF
--- a/apps/clean-app/server.ts
+++ b/apps/clean-app/server.ts
@@ -1,5 +1,5 @@
 import { createApp, server } from "@databricks/appkit";
 
 createApp({
-  plugins: [server({ port: 8001 })],
+  plugins: [server()],
 });


pkosiec:
Could you please add a sentence about the .obo extension in the docs? `docs/docs/plugins.md#analytics-plugin`

MarioCadenas (approved):

PR #64, opened 2026-01-26
feat: embed `appkit` CLI and AI-targeted docs into the `appkit` and `appkit-ui` packages

pkosiec:
This PR:
- consolidates all CLI tooling under a single `npx @databricks/appkit` command
- embeds AI-targeted documentation directly into the published packages.
- fixes double docs building on packing `appkit` and `appkit-ui` packages, which caused some race conditions

Resolves https://databricks.atlassian.net/browse/LKB-8734
Resolves https://databricks.atlassian.net/browse/LKB-8323

## Testing

```
# From repo root
pnpm i
pnpm clean
pnpm pack:sdk

# Then install the local tarball in your test location
cd apps/dev-playground
npm install ../../packages/appkit/tmp/*.tgz --force

# Now, test the commands
npx @databricks/appkit docs
npx @databricks/appkit lint
npx @databricks/appkit generate-types
npx @databricks/appkit setup --write
npx @databricks/appkit --version
```

### DIFF
--- a/packages/shared/src/cli/index.ts
+++ b/packages/shared/src/cli/index.ts
@@ -1,14 +1,24 @@
 #!/usr/bin/env node
 
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
 import { Command } from "commander";
 import { setupCommand } from "./commands/setup.js";
 import { generateTypesCommand } from "./commands/generate-types.js";
 import { lintCommand } from "./commands/lint.js";
 import { docsCommand } from "./commands/docs.js";
 
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const pkgPath = join(__dirname, "../../package.json");
+const pkg = JSON.parse(readFileSync(pkgPath, "utf-8"));
+
 const cmd = new Command();
 
-cmd.name("appkit").description("CLI tools for Databricks AppKit");
+cmd
+  .name("appkit")
+  .description("CLI tools for Databricks AppKit")
+  .version(pkg.version);
 
 cmd.addCommand(setupCommand);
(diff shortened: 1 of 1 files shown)


packages/shared/scripts/postinstall.js
```diff
@@ -2,5 +2,5 @@
 console.log("");
 console.log("[@databricks/appkit] To setup AI assistant instructions, run:");
 console.log("");
-console.log("  npx appkit-setup --write");
+console.log("  npx appkit setup --write");
```
MarioCadenas: this should be the same as the main command

packages/shared/src/cli/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { Command } from "commander";
+import { setupCommand } from "./commands/setup.js";
+import { generateTypesCommand } from "./commands/generate-types.js";
+import { lintCommand } from "./commands/lint.js";
+import { docsCommand } from "./commands/docs.js";
+
+const cmd = new Command();
+
+cmd.name("appkit").description("CLI tools for Databricks AppKit");
```
MarioCadenas: super small detail, but maybe we can also put in the description the current version from the package.json file, so if there are any issues we can easily see on which version?

template/package.json
```diff
@@ -12,16 +12,16 @@
     "typecheck": "tsc -p ./tsconfig.server.json --noEmit && tsc -p ./tsconfig.client.json --noEmit",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
-    "lint:ast-grep": "appkit-lint",
+    "lint:ast-grep": "appkit lint",
```
MarioCadenas: I'm confused, this is `appkit` and on top it said `npx @databricks/appkit [command]` which one is the one?

PR #76, opened 2026-02-02
test: add Playwright integration tests for dev-playground

calvarjorge:
- Add frontend integration tests using Playwright
- Tests use mocked API responses to verify UI behavior without backend dependencies
- We will have them run in the CI in a future PR

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,13 @@ The following command will compile all the packages and app in watch mode.
 pnpm dev
 ```
 
+> **Note:** To avoid port collisions with the `clean-app` example, you should create a `.env` file in `apps/dev-playground` and set another port for this app:
+>
+> ```
+> DATABRICKS_APP_PORT=8001
+> ```
+
+
 ## Running the project in production mode
 
 Running the following command
--- a/apps/dev-playground/playwright.config.ts
+++ b/apps/dev-playground/playwright.config.ts
@@ -1,4 +1,5 @@
 import { defineConfig, devices } from "@playwright/test";
+import "dotenv/config";
 
 export default defineConfig({
   testDir: "./tests",
@@ -8,7 +9,7 @@ export default defineConfig({
   workers: process.env.CI ? 1 : undefined,
   reporter: "html",
   use: {
-    baseURL: "http://localhost:8000",
+    baseURL: `http://localhost:${process.env.DATABRICKS_APP_PORT || 8000}`,
     trace: "on-first-retry",
   },
   projects: [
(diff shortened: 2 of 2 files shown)


apps/dev-playground/tests/data-visualization.spec.ts
pkosiec: This test already has a value with mocked backend, but it would be multiplied if we converted it to E2E test. That is, we'd mock the SDK on backend, run the `dev-playground` app, and run this test. What do you think?

apps/dev-playground/package.json
```diff
@@ -25,6 +28,7 @@
     "zod": "^4.1.13"
   },
   "devDependencies": {
+    "@playwright/test": "^1.50.1",
```
MarioCadenas: should we add a more recent version? latest is 1.58, is there any reason we have 1.50.1?

apps/dev-playground/tests/fixtures/test-utils.ts
```diff
@@ -0,0 +1,140 @@
...
+export async function waitForPageLoad(page: Page) {
+  await page.waitForLoadState("networkidle");
+}
+
+export async function waitForChartsToLoad(page: Page) {
+  await waitForPageLoad(page);
+  await page.waitForFunction(
+    () => document.querySelectorAll(".animate-pulse").length === 0,
+    { timeout: 10000 },
+  );
```
MarioCadenas: these two should be wrapped in a Promise.all

apps/dev-playground/tests/utils/test-utils.ts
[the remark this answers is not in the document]
calvarjorge: Yes, this makes a lot of sense, moved!

apps/dev-playground/tests/utils/test-utils.ts
```diff
@@ -0,0 +1,140 @@
...
+export async function waitForPageLoad(page: Page) {
+  await page.waitForLoadState("networkidle");
+}
+
+export async function waitForChartsToLoad(page: Page) {
+  await waitForPageLoad(page);
+  await page.waitForFunction(
+    () => document.querySelectorAll(".animate-pulse").length === 0,
+    { timeout: 10000 },
+  );
```
calvarjorge: If they are wrapped they will run in parallel. If the page has not loaded, the second promise will evaluate to true immediately no?

apps/dev-playground/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Integration Tests
+
+Integration tests use Playwright to verify the application works correctly with mocked backend responses.
+
+**Note:** These are frontend-only integration tests. API calls are intercepted at the browser level and return mock data, so the AppKit backend plugins are not tested. They focus on verifying UI behavior, navigation, data rendering, and client-side interactions.
+
+### Running Tests
+
```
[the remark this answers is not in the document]
pkosiec: @calvarjorge I think you missed that PR, PTAL - let's configure playwright to point to the app working on 8001 port as our `.env.dist` configures, alright?

PR #77, opened 2026-02-02
chore: setup import ordering with Biome

calvarjorge:
- Update lint-staged to use `biome check --write` for combined linting, formatting, and import sorting on pre-commit
- Update CI to use `pnpm check` and fail in case of unordered imports. Example behaviour here: https://github.com/databricks/appkit/pull/78
- Most files are auto-generated changes, please look at `ci.yml` and `package.json`.


.github/workflows/ci.yml
```diff
@@ -50,10 +50,8 @@ jobs:
           cache: 'pnpm'
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
-      - name: Run Lint
-        run: pnpm run lint
-      - name: Run Format Check
-        run: pnpm run format:check
+      - name: Run Biome Check
```
pkosiec: We should do the same with docs - please add docs:check and docs:check:fix and use them for CI pipelines. Thank you!

PR #82, opened 2026-02-03
feat(appkit): plugin manifest definition

MarioCadenas:
## Summary

This PR introduces a **plugin manifest definition** and **resource registry** so plugins can declare their Databricks resource requirements in a machine-readable way. Tooling can then derive app configuration (e.g. `app.yaml` / bundle resources) from those manifests.

### Plugin manifest

- **Manifest is required**: Every plugin must define a static `manifest` with `name`, `displayName`, `description`, and `resources` (required/optional).
- **Manifest format**: JSON Schema in `plugin-manifest.schema.json`; built-in plugins (analytics, server) ship a `manifest.json` next to their code and a typed `manifest.ts` used at runtime.
- **Runtime requirements**: Plugins can implement static `getResourceRequirements(config)` so optional resources can become required based on config (e.g. warehouse only when analytics is enabled).

### Resource registry

- **`ResourceRegistry`**: Collects resource requirements from all plugins at app creation, deduplicates by `type` + `resourceKey`, and merges permissions (e.g. takes the most permissive when several plugins need the same resource).
- **Resource types**: Typed enum (`ResourceType`) and per-type permissions (e.g. `sql_warehouse` → `CAN_USE` / `CAN_MANAGE`) in `packages/appkit/src/registry/types.ts`.
- **Validation**: `createApp()` calls `registry.collectResources(rawPlugins)` then `registry.enforceValidation()` so missing or invalid manifests fail fast.

### Manifest loader

- **`getPluginManifest(plugin)`**: Loads and validates the manifest from a plugin constructor; normalizes string type/permission into `ResourceType` / `ResourcePermission`.
- **`getResourceRequirements(plugin, config?)`**: Returns the flattened list of requirements (manifest + optional runtime from `getResourceRequirements(config)`).

### CLI: `appkit plugins sync`

- New command in `@databricks/shared` that discovers plugins from the app's server entrypoint (AST parsing of imports and plugin usage), reads their manifests from installed packages or local paths, and builds/updates **`appkit.plugins.json`**.
- Template uses this file (and the new `template-plugins.schema.json`) so the template's plugin list and resource metadata stay in sync with the app.
- Options: `--write` to write the file, `--output <path>` for custom path.

### Template and docs

- **Template**: `appkit.plugins.json` added; analytics bundle config simplified (fewer redundant YAML files); `databricks.yml.tmpl` and server wiring updated to use the new plugin manifest story.
- **Docs**: Plugin docs and API reference updated (Plugin class, `PluginManifest`, `ResourceRegistry`, `ResourceType`, new functions), plus new schema docs and copy script for static schemas.

### Other

- **Dev-playground**: Reconnect and telemetry-example plugins updated to define a manifest.
- **Removals**: `env-validator` and its exports removed from appkit.
- **Tests**: New/updated tests for manifest loader, resource registry (including `getResourceRequirements` and permission merging), integration with `createApp`, and `plugins-sync` CLI.


packages/appkit/src/registry/manifest-loader.ts
```diff
@@ -0,0 +1,164 @@
...
+ */
+export function getPluginManifest(plugin: PluginConstructor): PluginManifest {
+  const pluginName = plugin.name || "unknown";
+
+  try {
+    // Check for static manifest property
+    if (!plugin.manifest) {
+      throw new ConfigurationError(
+        `Plugin ${pluginName} is missing a manifest. All plugins must declare a static manifest property.`,
+      );
```
pkosiec: Must? maybe we'll keep it optional for now? I can imagine some simple plugins that just inject some server routes. Like the telemetry examples plugin from the dev playground 🤔 

(later we should enforce it, sure, but I'm just not sure if now is the best timing)

PR #85, opened 2026-02-04
chore: resource registry

MarioCadenas:
## 📚 Stack (3/3)

```
🏠 main (base)
│
└──▶ ✅ refactor-move-core-plugins
     │
     └──▶ ✅ plugin-manifest-definition
          │
          └──▶ 👉 resource-registry ◀ this PR
```

## 📝 Changes

- `b7f9199` chore: resource registry

---

<details>
<summary>📖 About this stack</summary>

This PR is part of a stack of changes. Each PR in the stack builds on the one below it.

- ⬇️ **Base**: This PR is based on `plugin-manifest-definition`
- ⬆️ **Next**: This is the top of the stack

</details>

### DIFF
--- a/packages/appkit/src/core/appkit.ts
+++ b/packages/appkit/src/core/appkit.ts
@@ -1,6 +1,5 @@
 import type {
   BasePlugin,
-  BasePluginConfig,
   CacheConfig,
   InputPluginMap,
   OptionalConfigPluginDef,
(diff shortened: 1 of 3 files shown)


packages/appkit/src/core/appkit.ts
pkosiec: Can we extract resource collection and validation to some separate methods? The logic here a bit makes it harder to read the high-level logic for registering plugins.


Like it was before:
```ts
    pluginInstance.validateEnv();
```

so now I can imagine, we'd have two calls: validation + collecting resources

PR #91, opened 2026-02-09
fix: make warehouseId optional in ServiceContext when no plugin requires it

pffigueiredo:
## Description

Makes `warehouseId` optional in `ServiceContext` when no registered plugin declares a SQL Warehouse in its manifest. Previously, `ServiceContext.createContext()` always called `getWarehouseId(client)`, which in production throws if `DATABRICKS_WAREHOUSE_ID` is unset, and in dev triggers an unnecessary API call.

The need for a warehouse is now derived from the existing **plugin manifest / ResourceRegistry**: if any plugin declares a required `sql_warehouse` resource in its manifest, `ServiceContext` resolves the warehouse; otherwise it is skipped. No new plugin API (e.g. `requiredResources`) was added — the manifest is the single source of truth.

**FEATURES**
- `ServiceContext.initialize(options?, client?)` accepts `options?.warehouseId` to gate warehouse resolution.
- `AppKit._createApp()` uses `ResourceRegistry`: collect resources from manifests, derive `needsWarehouse` from `registry.getRequired()` (any `ResourceType.SQL_WAREHOUSE`), then call `ServiceContext.initialize({ warehouseId: needsWarehouse })` before `registry.enforceValidation()`.

**FIXES**
- `createApp({ plugins: [server()] })` no longer fails when `DATABRICKS_WAREHOUSE_ID` is missing.
- Apps without warehouse-dependent plugins (e.g. analytics) avoid unnecessary warehouse ID resolution and API calls.

**BREAKING CHANGES**
- None. Existing plugins that declare `sql_warehouse` in their manifest (e.g. Analytics) keep current behavior; apps using only `server()` no longer need the env var.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,3 @@ coverage
 *.tsbuildinfo
 
 .turbo
-
-# AI generated files
-.serena
-docs/plans/
--- a/packages/appkit/src/context/service-context.ts
+++ b/packages/appkit/src/context/service-context.ts
@@ -24,7 +24,7 @@ export interface ServiceContextState {
   client: WorkspaceClient;
   /** The service principal's user ID */
   serviceUserId: string;
-  /** Promise that resolves to the warehouse ID (only present when a plugin requires it) */
+  /** Promise that resolves to the warehouse ID (only present when a plugin requires `SQL_WAREHOUSE` resource) */
   warehouseId?: Promise<string>;
   /** Promise that resolves to the workspace ID */
   workspaceId: Promise<string>;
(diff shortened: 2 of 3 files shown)


packages/appkit/src/core/appkit.ts
```diff
@@ -149,11 +150,13 @@ export class AppKit<TPlugins extends InputPluginMap> {
     TelemetryManager.initialize(config?.telemetry);
     await CacheManager.getInstance(config?.cache);
 
-    // Initialize ServiceContext for Databricks client management
-    // This provides the service principal client and shared resources
-    await ServiceContext.initialize(config?.client);
-
+    // Collect required resources from all plugins before initializing context
     const rawPlugins = config.plugins as T;
+    const requiredResources = AppKit.collectRequiredResources(rawPlugins);
```
pkosiec: Could you please rebase your PR? There's resource registry now, where all Plugins have their manifests with required resources. Maybe we could use it to mark plugin as the one which needs the warehouse ID? 🤔

PR #105, opened 2026-02-16
fix: sync template versions on release

MarioCadenas:
this pr creates bumps the template version after the package has been published, and publishes a new tag for it since we can't properly update the package-lock file if the package is not published

### DIFF
--- a/tools/publish-template-tag.ts
+++ b/tools/publish-template-tag.ts
@@ -73,20 +73,17 @@ if (installExit !== 0) {
 }
 
 // 3. Git add, commit, tag, push
-if (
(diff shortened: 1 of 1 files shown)


tools/sync-template-versions.ts
pkosiec: As a part of https://databricks.atlassian.net/browse/LKB-9008 we agreed that we will run some validation with the `apps init` command, to verify that the template works before releasing it.

Will it come in a separate PR? It doesn't need to be here but I just want to make sure we do at least a smoke test to avoid issues like before, which broke Arseni's demo.

PR #108, opened 2026-02-17
feat(appkit): add Genie plugin for AI/BI space integration

calvarjorge:
## Summary

- Add new Genie plugin
- Single SSE endpoint (`POST /api/genie/:alias/messages`) handles both new and follow-up conversations
- Space alias abstraction: users configure URL aliases, which map to a Genie Space ID in the backend.
- Always executes as user (OBO) via `asUser(req)`
- SSE event flow: `message_start` → `status` (×N) → `message_result` → `query_result` (×N per query attachment)
- Configurable timeout (default 2min, `0` for indefinite)
- No cache/retry (chat is stateful and non-idempotent)
- Programmatic `sendMessage` API exposed via `exports()`
- Conversation history SSE endpoint (`GET /api/genie/:alias/conversations/:conversationId`) replays full conversation using the same event types, enabling page refresh without losing chat state
  - `includeQueryResults` query param (default `true`) controls whether query result data is fetched
  - Messages streamed first, then query results fetched in parallel
  - Server-side pagination with 200-message safety cap
  - Programmatic `getConversation` API exposed via `exports()`
- Fix pre-existing ajv type resolution issue in shared package

### DIFF
--- a/packages/appkit/src/connectors/genie/poll-waiter.test.ts
+++ b/packages/appkit/src/connectors/genie/poll-waiter.test.ts
@@ -1,9 +1,5 @@
 import { describe, expect, test, vi } from "vitest";
-import {
-  type Pollable,
-  type PollEvent,
-  pollWaiter,
-} from "../../../connectors/genie";
+import { type Pollable, type PollEvent, pollWaiter } from "./poll-waiter";
 
 function createMockWaiter<P>(opts: {
   progressValues?: P[];


packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -0,0 +1,516 @@
+import type {
+  GenieMessage,
+  GenieStartConversationResponse,
+} from "@databricks/sdk-experimental/dist/apis/dashboards";
+import Time, {
+  TimeUnits,
+} from "@databricks/sdk-experimental/dist/retries/Time";
+import type { Waiter } from "@databricks/sdk-experimental/dist/wait";
```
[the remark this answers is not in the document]
calvarjorge: I renamed the time import, which does seem to be exported through the index, but not the rest. Also simplified the `createLogger` import changing from `../../logging/logger` to `../../logging`.

docs/docs/api/appkit/Variable.genie.md
pkosiec: we should add
```ts
/**
 * @internal
 */
```
comment to the genie const to avoid generating this

packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -0,0 +1,514 @@
...
+      query: att.query
+        ? {
+            title: att.query.title,
+            description: att.query.description,
+            query: att.query.query,
+            statementId: att.query.statement_id,
+          }
+        : undefined,
+      text: att.text ? { content: att.text.content } : undefined,
+      suggestedQuestions: att.suggested_questions?.questions,
```
pkosiec: do you have a screenshot of how the suggested questions look like? (I mean the user point of view, so it's probably a question to this PR: https://github.com/databricks/appkit/pull/116

packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -0,0 +1,517 @@
...
+    });
+
+    this.route(router, {
+      name: "getConversation",
+      method: "get",
+      path: "/:alias/conversations/:conversationId",
+      handler: async (req: express.Request, res: express.Response) => {
+        await this.asUser(req)._handleGetConversation(req, res);
+      },
+    });
```
MarioCadenas: This is only supporting OBO, but what about service principal?

packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  async _handleSendMessage(
```
MarioCadenas: we are not handling reconnections, or at least I don't see it? am I missing something?

packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -0,0 +1,517 @@
...
+            resultMessageId = result.message_id;
+            return result;
+          })()
+            .catch((err: Error) => {
+              waiterError = err;
+            })
+            .finally(() => {
+              waiterDone = true;
+              notifyGenerator();
+            });
```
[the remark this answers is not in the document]
calvarjorge: i implemented the wrapper as we discussed, it makes the code easier to follow. the length of the func is reduced, although i don't think as much as we expected 😅

PR #109, opened 2026-02-18
fix: template sync

MarioCadenas:
github action to sync the template version after a release is published

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,13 +68,12 @@ jobs:
         id: version
         if: steps.mode.outputs.dry_run != 'true'
         run: |
-          VERSION=$(pnpm exec release-it --release-version --ci)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -61,6 +64,19 @@ jobs:
             echo "dry_run=${{ inputs.dry-run }}" >> $GITHUB_OUTPUT
           fi
 
+      - name: Determine version
+        id: version
+        if: steps.mode.outputs.dry_run != 'true'
+        run: |
+          VERSION=$(pnpm exec release-it --release-version --ci)
+          LATEST=$(git tag -l 'v*' --sort=-version:refname | head -1 | sed 's/^v//')
```
pkosiec: Do we need do print the latest version? I'd skip it because all of that magic with sed etc. 😄  Especially that we can track the actually latest version manually by pushed tags timestamp. 

So my suggestion is: Let's remove it 😄

PR #114, opened 2026-02-18
chore: add files route and UI to dev-playground

atilafassina:
Adds the new File Plugin to Dev Playground.

Add a new env_var to your `.env` file.
For example:

```
DATABRICKS_DEFAULT_VOLUME='/Volumes/main/info-atila/atila-volume'
```
(you can use that one, nothing critical there 😉 )

### DIFF
--- a/packages/appkit/src/plugins/files/index.ts
+++ b/packages/appkit/src/plugins/files/index.ts
@@ -1,5 +1,4 @@
 export * from "./defaults";
-export * from "./helpers";
 export * from "./manifest";
 export * from "./plugin";
 export * from "./types";


apps/dev-playground/client/src/appKitTypes.d.ts
```diff
@@ -14,46 +14,28 @@ declare module "@databricks/appkit-ui/react" {
           endDate: SQLDateMarker;
         };
         result: Array<{
-          /** @sqlType STRING */
-          app_name: string;
-          /** @sqlType STRING */
-          day_of_week: string;
-          /** @sqlType DECIMAL(35,2) */
-          spend: number;
+          ;
```
atilafassina: 💥 need to fix this in my typegen

happens also out of playground

PR #127, opened 2026-02-23
chore: generate template dynamically for specific enabled plugins

pkosiec:
First phase of the dynamic template, that works with analytics and/or lakebase plugins. 

Resolves https://databricks.atlassian.net/browse/LKB-9681

## Video

https://github.com/user-attachments/assets/fb54e03d-8a05-4aa4-a605-9b7ed273cc2a

## All PRs

1. CLI: https://github.com/databricks/cli/pull/4549
2. https://github.com/databricks/appkit/pull/123
3. https://github.com/databricks/appkit/pull/126
4. https://github.com/databricks/appkit/pull/127

## Scope

We agreed on the following scope:
1. remove harcoded paths on CLI - maybe just trigger an npm command as a part of the template at the end (if necessary, let's see)
1. from the template, I'll use Go templates to conditionally add/remove files content

### DIFF
--- a/template/server/routes/lakebase/todo-routes.ts
+++ b/template/server/routes/lakebase/todo-routes.ts
@@ -1,10 +1,14 @@
 {{if .plugins.lakebase -}}
 import { z } from 'zod';
-import type { LakebasePlugin, ServerPlugin } from '@databricks/appkit';
+import { Application } from 'express';
 
 interface AppKitWithLakebase {
-  lakebase: ReturnType<LakebasePlugin['exports']>;
-  server: ReturnType<ServerPlugin['exports']>;
+  lakebase: {
+    query(text: string, params?: unknown[]): Promise<{ rows: Record<string, unknown>[] }>;
+  };
+  server: {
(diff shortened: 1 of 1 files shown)


template/server/server.ts
```diff
@@ -1,7 +1,26 @@
...
+import { createApp{{range $name, $_ := .plugins}}, {{$name}}{{end}} } from '@databricks/appkit';
+{{- if .plugins.lakebase}}
+import { setupSampleLakebaseRoutes } from './routes/lakebase/todo-routes';
+{{- end}}
 
 createApp({
   plugins: [
-    {{.plugin_usages}}
+{{- if .plugins.lakebase}}
+    server({ autoStart: false }),
```
MarioCadenas: since the server is already defined as "requiredByTemplate" it will be added automatically, we don't have to hard code it, what would be nice is to get the config from the manifest and inject it here instead, but I guess that can be added later. But this seems very specific to lakebase so it can extend the routes, maybe we can do this in a different way?

PR #146, opened 2026-03-02
feat(genie): add automatic chart visualization for query results

calvarjorge:
## Summary

- **Chart type inference**: Automatically determines the best chart type (bar, line, pie, scatter, table) based on query column metadata (types, cardinality, column count)
- **Inline chart rendering**: `GenieQueryVisualization` component renders charts directly within Genie chat messages alongside SQL query details
- **Dev playground page**: Adds a `/chart-inference` route for testing chart inference with sample datasets

TODO in a subsequent PR:
- Adding a dropdown to change the default chart

### DIFF
--- a/docs/docs/api/appkit-ui/genie/GenieQueryVisualization.mdx
+++ b/docs/docs/api/appkit-ui/genie/GenieQueryVisualization.mdx
@@ -1,0 +1,34 @@
+# GenieQueryVisualization
+
+Renders a chart + data table for a Genie query result.
+
+
+## GenieQueryVisualization
+
+Renders a chart + data table for a Genie query result.
+
+- When a chart type can be inferred: shows Tabs with "Chart" (default) and "Table"
+- When no chart fits: shows only the data table
+- When data is empty/malformed: renders nothing
+
+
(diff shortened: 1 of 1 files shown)


packages/appkit-ui/src/react/charts/utils.ts
```diff
@@ -77,6 +77,35 @@ export function createTimeSeriesData(
...
+  indices.sort((a, b) => Number(xData[a]) - Number(xData[b]));
+
+  const sortedXData = indices.map((i) => Number(xData[i]));
+  const sortedYDataMap: Record<string, (string | number)[]> = {};
+  for (const key of yFields) {
+    const original = yDataMap[key];
+    sortedYDataMap[key] = indices.map((i) => Number(original[i]));
+  }
+
+  return { xData: sortedXData, yDataMap: sortedYDataMap };
```
MarioCadenas: this function converts all Y-values to Number when sorting, but should preserve their original types like `sortTimeSeriesAscending` does. This causes string Y-values to become NaN, breaking charts with numeric X-values but non-numeric Y-data.

PR #174, opened 2026-03-11
chore: build template artifact

MarioCadenas:
Creates an artifact with the template and appkit versions created from this pr


keugenek (approved):
LGTM\! This is exactly what we need for the eval pipeline — pre-built template artifacts per PR branch. We'll consume these artifacts from the evals-monitor Custom Run panel to let users test against specific AppKit branches.

PR #197, opened 2026-03-18
feat(files): per-volume in-app policy enforcement

atilafassina:
## Summary

- **Per-volume policy system**: `FilePolicy` — a function `(action, resource, user) → boolean` — gates every file operation before the Databricks API call. Ships with built-in helpers (`policy.publicRead()`, `policy.allowAll()`, `policy.denyAll()`) and combinators (`all`, `any`, `not`) for composition.
- **Service-principal execution model**: HTTP routes now always execute as the service principal. The policy layer becomes the app-level gate — user identity is extracted from `x-forwarded-user` and passed to the policy function.
- **Safe defaults**: Volumes without an explicit policy default to `publicRead()` (read-only). A startup warning encourages setting an explicit policy.

## Motivation

Previously the files plugin relied entirely on OBO token forwarding and Unity Catalog grants. This meant:
1. No app-level way to restrict per-user operations (e.g. allow reads but deny uploads).
2. The SP's credentials were always used on HTTP routes regardless of UC grants.

Policies close this gap with a composable, per-volume authorization layer evaluated before any API call.

## Key changes

| Area | Details |
|------|---------|
| **`policy.ts`** (new) | `FileAction`, `FileResource`, `FilePolicyUser` types · `PolicyDeniedError` · `policy` namespace with `publicRead()`, `allowAll()`, `denyAll()`, `all()`, `any()`, `not()` |
| **`plugin.ts`** | Policy enforcement wired into every route handler and programmatic API · Removed `isInUserContext`/OBO gating · Added `_enforcePolicy()`, `_checkPolicy()`, `_extractUser()` · `_executeOrThrow()` for proper 4xx propagation |
| **`types.ts`** | `policy?: FilePolicy` on `VolumeConfig` · Updated JSDoc to reflect SP-first execution model |
| **`index.ts`** | Re-exports `policy` from top-level barrel |
| **dev-playground** | Updated to use `policy.allowAll()` on the default volume |
| **docs** | Full plugin docs (`files.md`) with permission model, built-in policies, combinators, custom policies · Generated API reference for `policy` namespace |

## Test coverage

- **Policy combinators** (`policy.test.ts`): all helpers, async policies, `PolicyDeniedError`, edge cases
- **Plugin enforcement** (`plugin.test.ts`): route-level 401/403 responses, SDK API with user/SP identity, upload size in resource, cache invalidation under SP model
- **Integration tests** updated for the policy-first model

## Test plan

- [x] Unit tests for policy combinators and helpers
- [x] Unit tests for plugin policy enforcement across all route handlers
- [x] Existing plugin tests updated for new execution model
- [x] Linting and type checks pass
- [ ] Manual verification against a Databricks workspace via `pnpm dev`


pkosiec (commented):
Great work! Overall LGTM, pls see my small comments on the proposal and here - they might affect some methods naming etc.

**Disclaimer**: unfortunately I wasn't able to test it, I'll run it next week 👍 Unless you could record a short, raw demo video showcasing how the policies work ?

Also, I think it would be worth if we had approval for at least the RFC from Fabian / Mario who were much more involved in the files discussions. Thanks!

packages/appkit/src/context/execution-context.ts
```diff
@@ -81,11 +81,3 @@ export function getWarehouseId(): Promise<string> {
 export function getWorkspaceId(): Promise<string> {
   return getExecutionContext().workspaceId;
 }
-
-/**
- * Check if currently running in a user context.
- */
-export function isInUserContext(): boolean {
```
MarioCadenas, line 88: why?

apps/dev-playground/server/index.ts
```diff
@@ -1,5 +1,12 @@
 import "reflect-metadata";
-import { analytics, createApp, files, genie, server } from "@databricks/appkit";
+import {
+  analytics,
+  createApp,
+  files,
+  genie,
+  policy,
```
MarioCadenas: can we change the export? instead of exporting the policy as an object here, can we put it as a property of the files plugin? so its used like 
```ts
import { files } from "@databricks/appkit"

files({ volumes: { default: { policy: files.policy.allowAll() } } }),
```

packages/appkit/src/context/user-context.ts
```diff
@@ -20,13 +20,13 @@ export interface UserContext {
 }
 
 /**
- * Execution context can be either service or user context.
- */
-export type ExecutionContext = ServiceContextState | UserContext;
```
MarioCadenas, line 25: why again? haha

packages/appkit/src/plugins/files/plugin.ts
```diff
@@ -389,16 +467,63 @@ export class FilesPlugin extends Plugin {
...
+        }
+      },
+      options,
+      userKey,
+    );
+
+    if (result === undefined && capturedClientError) {
+      throw capturedClientError;
+    }
+
```
MarioCadenas: `_executeOrThrow` is defined (~30 lines) but never called by any route handler. Was this intended to replace `this.execute()` in the handlers? Either wire it in or remove it to avoid confusion.

docs/docs/plugins/files.md
```diff
@@ -97,6 +100,115 @@ files({
...
+};
+
+files({
+  volumes: { reports: { policy: adminOnly } },
+});
+```
+
+#### Enforcement
+
+- **HTTP routes**: Policy checked before every operation. Denied → `403` JSON response with `"Action denied by volume policy"`.
```
MarioCadenas: The docs say:

> Denied → `403` JSON response with `"Action denied by volume policy"`.

But `PolicyDeniedError` actually produces `Policy denied "{action}" on volume "{volumeKey}"`. These should match.

packages/appkit/src/plugins/files/policy.ts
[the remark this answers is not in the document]
atilafassina: > Not sure if we shouldn't update the AppKit template to utilize the policies for the sample path?

I think we shouldn't because it would imply a configuration is needed. We have a default for policies

> And the same with an agent skill for files? Agent might be confused without additional guidance - we need to verify that.

100% need to adjust the skills for this before releasing

packages/appkit/src/context/user-context.ts
```diff
@@ -20,13 +20,13 @@ export interface UserContext {
...
+ * Type guard to check if a context is a UserContext.
  */
 export function isUserContext(ctx: ExecutionContext): ctx is UserContext {
   return "isUserContext" in ctx && ctx.isUserContext === true;
 }
+
+/**
+ * Execution context can be either service or user context.
+ */
+export type ExecutionContext = ServiceContextState | UserContext;
```
MarioCadenas: we can probably undo this change no? as its the same as we already had unless I'm missing something, seems like the type is just moved to another line

packages/appkit/src/plugins/files/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./defaults";
 export * from "./plugin";
+export * from "./policy";
```
MarioCadenas: do we need this export?

packages/appkit/src/plugins/files/plugin.ts
```diff
@@ -390,7 +471,7 @@ export class FilesPlugin extends Plugin {
       : "__root__";
     const listKey = this.cache.generateKey(
       [`files:${volumeKey}:list`, cachePathSegment],
-      userId,
+      getCurrentUserId(),
```
MarioCadenas, line 430: I think this would always be the service principal, so I'm not sure if its what we want or not, but I would take a look

packages/appkit/src/plugins/files/plugin.ts
```diff
@@ -148,60 +225,61 @@ export class FilesPlugin extends Plugin {
...
+   * Not used internally — `_createPolicyWrappedAPI` handles all current
+   * call sites. Kept as a `protected` extension point so subclasses can
+   * override `exports()` or build custom APIs with raw connector access,
+   * e.g. background jobs or migrations that should bypass user-facing policies.
+   *
+   * @security This method skips all policy enforcement. Do not expose its
+   * return value to HTTP routes or end-user-facing code paths — use
+   * `_createPolicyWrappedAPI` for anything that serves user requests.
    */
   protected createVolumeAPI(volumeKey: string): VolumeAPI {
```
MarioCadenas: do we need to keep this? given that is exactly the same as the other except for the whole bypass? if we actually have a use case where we need the bypass, couldn't we just put a flag in the other method to build a "bypassed" api? just wondering 😄

packages/appkit/src/plugins/files/plugin.ts
atilafassina: This is by design.  Both the cache write (in _handleList) and the invalidation run in SP context.
so they should match the SP id for invalidation to work consistently.


Does it make sense?

packages/appkit/src/plugins/files/plugin.ts
```diff
@@ -732,6 +777,13 @@ export class FilesPlugin extends Plugin {
       ? parseInt(rawContentLength, 10)
       : undefined;
 
+    if (
+      !(await this._enforcePolicy(req, res, volumeKey, "upload", path, {
+        size: contentLength,
```
MarioCadenas, line 782: Small security/correctness issue worth closing before merge: if `Content-Length` is malformed (e.g. `Content-Length: abc`), `parseInt(rawContentLength, 10)` returns `NaN`, and that `NaN` gets passed here as `resource.size` to the policy.

Any numeric comparison against `NaN` evaluates to `false`, so a documented size-gating policy like:

```ts
(_, resource) => !resource.size || resource.size < 10_000_000
```

...will silently **allow** the upload (because `!NaN === true`). Since `Content-Length` is a client-controlled header, this is effectively a policy bypass on attacker input. Blast radius is bounded by the volume-level `maxUploadSize` (the 413 check already guards with `!Number.isNaN(contentLength)`), but custom size-based policies — which the docs explicitly advertise as a supported use case via `FileResource.size` — are bypassable.

One-line fix: normalize once above, so both the policy and the 413 check see a clean value:

```ts
const parsed = rawContentLength ? parseInt(rawContentLength, 10) : NaN;
const contentLength = Number.isFinite(parsed) ? parsed : undefined;
```

And the 413 block below can drop its `!Number.isNaN` guard since `contentLength` is now `number | undefined`.

Would also be worth a quick test case: `Content-Length: abc` + a size-gating policy that denies > N → expect 403 (not 200).

atilafassina: nice finding!

I fixed and added test cases for 
- fully bogus cases (e.g.:`abc`) =>`NaN`
- partially valid that may fool `parseInt` (123abc`) => `123`
- negative numbers (`-1`) => `-1` valid, but we don't want that either

MarioCadenas (approved):

PR #200, opened 2026-03-20
feat: add Vector Search plugin

adamgurary:
## Summary

- Adds `@databricks/appkit-vector-search` — a plugin that gives Databricks Apps built with AppKit instant vector search query capabilities
- Ships backend (Express routes, VS REST API client, service principal + OBO auth) and frontend (React hook, styled components with Radix UI)
- Developer experience target: ~45 lines for a full search page with search box, results, filters, and keyword highlighting
- 82 tests included; validated against real VS index on dogfood

## What's included

**Backend plugin** (`src/plugin/`):
- `VectorSearchPlugin.ts` — plugin class with lifecycle, config, route injection
- `VectorSearchClient.ts` — REST API client for VS endpoints
- `routes.ts` — Express route handlers
- `auth.ts` — service principal default, OBO opt-in per index

**Frontend UI** (`src/ui/`):
- `useVectorSearch` React hook
- `SearchBox`, `SearchResults`, `SearchResultCard`, `SearchLoadMore` components

## Design decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Package structure | Single package (backend + UI) | Shared types, single dependency, matches Lakebase plugin pattern |
| Default search mode | Hybrid (ANN + keyword) | Best out-of-the-box quality |
| Reranking | Off by default, opt-in per index | Adds latency; too slow for interactive search by default |
| Auth | Service principal default, OBO opt-in | Simple default, secure option when needed |

## Test plan

- [ ] Review plugin structure against existing AppKit plugin patterns (Lakebase)
- [ ] Run test suite (`vitest run` in `packages/vector-search/`)
- [ ] Validate against live VS index on dogfood
- [ ] Review API surface and types

### DIFF
--- a/packages/appkit/src/plugins/index.ts
+++ b/packages/appkit/src/plugins/index.ts
@@ -4,4 +4,3 @@ export * from "./genie";
 export * from "./lakebase";
 export * from "./server";
 export * from "./serving";
-export * from "./vector-search";
--- a/apps/dev-playground/client/src/routes/vector-search.route.tsx
+++ b/apps/dev-playground/client/src/routes/vector-search.route.tsx
@@ -84,7 +84,10 @@ function VectorSearchRoute() {
               onKeyDown={handleKeyDown}
               className="flex-1"
             />
-            <Button onClick={() => void handleSearch()} disabled={loading || !query.trim()}>
+            <Button
+              onClick={() => void handleSearch()}
+              disabled={loading || !query.trim()}
+            >
               {loading ? (
                 "Searching..."
               ) : (
@@ -105,15 +108,18 @@ function VectorSearchRoute() {
           {response && (
             <div className="flex flex-col gap-4">
               <div className="text-sm text-muted-foreground">
-                {response.totalCount} result{response.totalCount !== 1 ? "s" : ""} &middot;{" "}
+                {response.totalCount} result
+                {response.totalCount !== 1 ? "s" : ""} &middot;{" "}
                 {response.queryTimeMs}ms &middot; {response.queryType}
               </div>
 
               {response.results.length === 0 ? (
-                <p className="text-sm text-muted-foreground">No results found.</p>
+                <p className="text-sm text-muted-foreground">
+                  No results found.
+                </p>
               ) : (
                 response.results.map((result, index) => (
-                  <Card key={index}>
+                  <Card key={`${result.score}-${index}`}>
                     <CardHeader className="pb-2">
                       <CardTitle className="text-sm font-medium flex items-center justify-between">
                         <span>Result {index + 1}</span>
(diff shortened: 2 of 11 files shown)


atilafassina (changes requested):
Great work!

I left a few comments, some are minor but I think there are 3 critical ones

1. not wrapping the plugin connector calls in the executor
2. returning plain objects as error
3. not wrapping the `embeddingFn` in a try/catch

The AbortControl in the connector would be nice to have too, but that's not a big issue if we have the plugin calls wrapped in the executor.

The "hybrid" hardcode in the response is mostly a question from me, maybe it's a non-issue.

---

Once we land in the right API, we need to add docs and a route in the dev-playground. Btu I'd defer that for after a the reviews to avoid re-work. Ofc, you do what you prefer :)

packages/appkit/src/plugins/index.ts
pkosiec: Please also remember about the docs (docs/docs/plugins) - we need to ensure the plugin is documented. Thank you!

adamgurary:
Addressed all review feedback from @atilafassina, @calvarjorge, and @pkosiec. Rebased onto current main.

### Blockers (Atila)
- **`this.execute()` wrapping** — All connector calls now go through the interceptor chain (retry/cache/timeout/telemetry). `vectorSearchDefaults` feeds into `execute()`, no longer dead code.
- **Error instances** — Programmatic `query()` throws `new Error(...)` instead of plain objects.
- **`embeddingFn` try/catch** — User-provided embedding function wrapped in try/catch with clear error message.

### Pattern alignment
- Error responses now use `{ error, plugin }` shape matching Files/Genie
- Extracted `_prepareQuery()` helper to DRY shared logic between `query()` and route handlers
- Removed `streamManager.abortAll()` from `shutdown()` (plugin has no streams)
- Removed `fromCache` from response type (execute handles caching transparently)
- Replaced `endpointName!` non-null assertion with runtime guard
- Next-page response uses `indexConfig.queryType` instead of hardcoded `"hybrid"`
- Removed unused `Context` import from connector; AbortSignal passed from `execute()` to connector

### New content
- **manifest.json** — Added `vector_search_index` required resource with env fields for `apps init` flow
- **Dev playground** — Server config, nav link, search route with form + results display
- **Docs** — `docs/docs/plugins/vector-search.md` with usage, config options, API reference
- **Template** — `App.tsx` routes, `VectorSearchPage.tsx`, `appkit.plugins.json` entry
- **Package export** — `vectorSearch` added to top-level `index.ts`

### Skipped
- **Zod request body validation** (Jorge's AI review #4) — No existing plugin uses Zod for `req.body` validation. All use simple `if (!field)` guards. Matched existing patterns.

### Note on CI
CI doesn't appear to run on fork PRs (protected runner group). Could not run lint/typecheck/tests locally either — the Databricks npm proxy (`npm-proxy.dev.databricks.com`) is returning empty replies / ECONNRESET. If a maintainer can trigger CI or approve the workflow run, that would help verify. Happy to fix anything CI catches.

PR #240, opened 2026-04-03
feat: add serving type generator, Vite plugin, and UI hooks

pkosiec:
## Summary
- Add Vite plugin that auto-generates TypeScript types from serving endpoint OpenAPI schemas
- Add AST-based server file extraction (`@ast-grep/napi`) for automatic endpoint discovery
- Add schema-to-TypeScript conversion with OpenAI-compatible streaming chunk inference
- Add `useServingInvoke` and `useServingStream` React hooks with type-safe registry support

## Demo

https://github.com/user-attachments/assets/a8ee32f4-1c10-4600-9e1c-5481f20ef2c2

## PR Stack — Model Serving
| # | PR | Description |
|---|-----|------------|
| 1 | #239 | Serving connector & plugin |
| 2 | **this PR** | Type generator, Vite plugin & UI hooks |
| 3 | #241 | Dev-playground, template & docs |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,4 @@ coverage
 
 .turbo
 
-# AppKit type generator caches
 .databricks
--- a/packages/appkit/src/type-generator/serving/vite-plugin.ts
+++ b/packages/appkit/src/type-generator/serving/vite-plugin.ts
@@ -2,7 +2,7 @@ import path from "node:path";
 import type { Plugin } from "vite";
 import { createLogger } from "../../logging/logger";
 import type { EndpointConfig } from "../../plugins/serving/types";
-import { generateServingTypes } from "./generator";
+import { generateServingTypes } from "../index";
 import {
   extractServingEndpoints,
   findServerFile,
--- a/packages/appkit/src/type-generator/index.ts
+++ b/packages/appkit/src/type-generator/index.ts
@@ -2,7 +2,7 @@ import fs from "node:fs/promises";
 import dotenv from "dotenv";
 import { createLogger } from "../logging/logger";
 import { generateQueriesFromDescribe } from "./query-registry";
-import { generateServingTypes } from "./serving/generator";
+import { generateServingTypes as generateServingTypesImpl } from "./serving/generator";
 import type { QuerySchema } from "./types";
 
 dotenv.config();
@@ -88,4 +88,7 @@ export async function generateFromEntryPoint(options: {
   logger.debug("Type generation complete!");
 }
 
-export { generateServingTypes };
+// Rolldown tree-shaking only preserves "own exports" (locally defined) — not re-exports.
+// A local binding ensures the serving vite plugin's import keeps this in the dependency graph,
+// mirroring how generateFromEntryPoint (also defined here) is preserved via the analytics vite plugin.
+export const generateServingTypes = generateServingTypesImpl;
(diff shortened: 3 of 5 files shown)


packages/appkit/src/type-generator/serving/vite-plugin.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Vite plugin to generate TypeScript types for AppKit serving endpoints.
+ * Fetches OpenAPI schemas from Databricks and generates a .d.ts with
+ * ServingEndpointRegistry module augmentation.
+ *
+ * Endpoint discovery order:
+ * 1. Explicit `endpoints` option (override)
+ * 2. AST extraction from server file (server/index.ts or server/server.ts)
+ * 3. DATABRICKS_SERVING_ENDPOINT_NAME env var (single default endpoint)
+ */
+export function appKitServingTypesPlugin(
```
MarioCadenas, line 30: This plugin is not being added into the vite-dev-server along with `appKitTypesPlugin`, so we should add it too

packages/appkit/src/type-generator/serving/fetcher.ts
```diff
@@ -0,0 +1,158 @@
...
+          "Failed to fetch schema for '%s' (HTTP %d), skipping%s",
+          endpointName,
+          res.status,
+          body ? `: ${body}` : "",
+        );
+      }
+      return null;
+    }
+
+    const rawSpec: unknown = await res.json();
```
MarioCadenas: can we use the workspace client here as in the other pr or it doesn't work for this one?

packages/appkit/src/type-generator/serving/generator.ts
```diff
@@ -0,0 +1,276 @@
...
+  outFile: string;
+  endpoints?: Record<string, EndpointConfig>;
+  noCache?: boolean;
+}
+
+/**
+ * Generates TypeScript type declarations for serving endpoints
+ * by fetching their OpenAPI schemas and converting to TypeScript.
+ */
+export async function generateServingTypes(
```
MarioCadenas, line 37: can we maybe simplify this method a little bit? aren't there parts we can extract outside to make it more readable?

packages/appkit-ui/src/react/hooks/use-serving-invoke.ts
```diff
@@ -0,0 +1,111 @@
...
+  const { alias, autoStart = false } = options;
+
+  const [data, setData] = useState<TResponse | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const abortControllerRef = useRef<AbortController | null>(null);
+
+  const urlSuffix = alias
+    ? `/api/serving/${encodeURIComponent(String(alias))}/invoke`
+    : "/api/serving/invoke";
```
MarioCadenas, line 65: nit: this can probably be even easier if we export a clientConfig on the serve plugin and consume the data here

MarioCadenas:
also almost forgot, this should be hooked into the `packages/shared/src/cli/commands/generate-types.ts` command too

packages/appkit-ui/src/react/hooks/use-serving-invoke.ts
pkosiec: clientConfig actually can help here with validating the passed alias by user - will do it 👍

.gitignore
```diff
@@ -9,3 +9,6 @@ coverage
 *.tsbuildinfo
 
 .turbo
+
+# AppKit type generator caches
+.databricks
```
MarioCadenas, line 13: this is from DABs AFAIK

PR #256, opened 2026-04-07
feat: add INLINE + ARROW_STREAM format support for analytics plugin

### LINKED ISSUE #242: Analytics plugin incompatible with ARROW_STREAM-only warehouses
## Summary

AppKit's analytics plugin defaults to `disposition: "INLINE"` + `format: "JSON_ARRAY"` for SQL queries. Some serverless warehouses only support `ARROW_STREAM` for inline disposition, causing all analytics queries to fail at runtime.

The ARROW code path (`format: "ARROW"`) uses `disposition: "EXTERNAL_LINKS"`, which some warehouses also reject with `NOT_IMPLEMENTED: ExternalLinks disposition is not yet implemented`.

This means certain warehouses are fundamentally incompatible with AppKit's analytics plugin — there is no format/disposition combination that works.

## Reproduction

1. Deploy an AppKit app using a serverless SQL warehouse that enforces ARROW_STREAM inline (e.g. warehouse `000000000000000d` on appsdev)
2. Use `useAnalyticsQuery` with default JSON format → fails with `INVALID_PARAMETER_VALUE: Inline disposition only supports ARROW_STREAM format`
3. Use `useAnalyticsQuery` with `format: "ARROW"` → fails with `NOT_IMPLEMENTED: ExternalLinks disposition is not yet implemented`
4. Use `BarChart` with `format="arrow"` → same EXTERNAL_LINKS error

## Root cause

`@databricks/appkit/dist/connectors/sql-warehouse/defaults.js`:
```js
const executeStatementDefaults = {
    disposition: "INLINE",
    format: "JSON_ARRAY",
};
```

ARROW path in `analytics.js` line 104-108:
```js
formatParameters: {
    disposition: "EXTERNAL_LINKS",
    format: "ARROW_STREAM"
}
```

Neither path supports `INLINE` + `ARROW_STREAM`, which is the only combination some warehouses accept.

## Suggested fix

Add an `INLINE` + `ARROW_STREAM` code path:
1. Detect when `JSON_ARRAY` is rejected and auto-retry with `ARROW_STREAM` inline
2. Or add a config option: `analytics({ format: "ARROW_STREAM" })`
3. Or probe the warehouse at startup with a `SELECT 1` test query to determine supported formats

## Environment

- `@databricks/appkit` v0.14.1
- Warehouse: serverless PRO on AWS dev workspace

jamesbroadhead:
## Summary

Serverless warehouses return `ARROW_STREAM` + `INLINE` results as base64 Arrow IPC in `result.attachment` rather than `result.data_array`. The previous code path discarded inline data for any `ARROW_STREAM` response (designed for `EXTERNAL_LINKS`), so these warehouses silently returned empty results.

This PR makes the analytics plugin work across classic and serverless warehouses by handling both dispositions for `ARROW_STREAM`, decoding inline Arrow IPC attachments server-side, and falling back to `JSON_ARRAY` when a warehouse rejects `ARROW_STREAM` + `INLINE`.

## Changes

- **Format model collapsed to API enum names**: two formats — `JSON_ARRAY` and `ARROW_STREAM` — replacing the prior `JSON` / `ARROW` / `ARROW_STREAM` triple. `ARROW_STREAM` now supports both `INLINE` and `EXTERNAL_LINKS` dispositions.
  - Note: these strings (`JSON_ARRAY`, `ARROW_STREAM`, `INLINE`, `EXTERNAL_LINKS`) are the verbatim values defined by the Statement Execution API — we no longer translate to/from local aliases. Confirmed against the Databricks SDK Go source: [`Format`](https://github.com/databricks/databricks-sdk-go/blob/main/service/sql/model.go#L2526-L2530) (`JSON_ARRAY` / `ARROW_STREAM` / `CSV`) and [`Disposition`](https://github.com/databricks/databricks-sdk-go/blob/main/service/sql/model.go#L1831-L1833) (`INLINE` / `EXTERNAL_LINKS`). See also the [API reference](https://docs.databricks.com/api/workspace/statementexecution/executestatement#format).
- **Inline Arrow IPC decoding**: `_transformDataArray` detects `result.attachment` and decodes via `apache-arrow`'s `tableFromIPC`, producing the same row-object shape as `JSON_ARRAY` regardless of warehouse backend. `apache-arrow@21.1.0` added as a server dependency.
- **Format fallback**: `ARROW_STREAM` + `INLINE` requests automatically fall back to `JSON_ARRAY` if a classic warehouse rejects them. Explicit format requests are respected without fallback.
- **Default remains `JSON_ARRAY`** for compatibility.
- **UI hooks updated** to use `json_array` / `arrow_stream` naming in `useAnalyticsQuery`, `useChartData`, and chart types.

## Tests

Added ~147 unit tests covering coverage gaps surfaced during the rework:
- `connectors/genie/tests/client.test.ts` (new, 786 lines)
- `connectors/sql-warehouse/tests/client.test.ts` (new, 286 lines) — uses real base64 Arrow IPC captured from a live serverless warehouse
- `context/tests/service-context.test.ts` (new, 457 lines)
- `plugins/files/tests/upload-and-write.test.ts` (new, 1245 lines)
- `stream/tests/stream-registry.test.ts` (new, 582 lines)
- `plugins/analytics/tests/analytics.test.ts` (+217)

Coverage deltas: service-context 7%→100%, stream-registry 32%→100%, genie connector 61%→~97%, files plugin 69%→~89%.

## Test plan

- [ ] Deploy an app against a serverless warehouse that returns `ARROW_STREAM` + `INLINE` attachments — verify `useAnalyticsQuery` returns rows
- [ ] Deploy against a classic warehouse — verify `JSON_ARRAY` default works and `ARROW_STREAM` + `EXTERNAL_LINKS` still works
- [ ] Verify automatic fallback when a warehouse rejects `ARROW_STREAM` + `INLINE`

Fixes #242


MarioCadenas:
Seen that there's a case of Arrow + inline, let's refactor what we had instead of introducing a new format. Let's change the format "ARROW" to "ARROW_STREAM" and allow it to use both "EXTERNAL_LINKS" and "INLINE". Then for now let's keep JSON + inline as the default. This might require some UI hooks changes too

jamesbroadhead:
Hi Mario — thanks for the review, you were right on both counts. I've rebased and reworked the PR:

**What's happening:** Serverless warehouses return `format: ARROW_STREAM` with `disposition: INLINE`, putting base64 Arrow IPC data in `result.attachment` instead of `data_array`. The docs say ARROW_STREAM only works with EXTERNAL_LINKS, but serverless violates that — and the existing code silently returned empty results for these responses.

**What I changed (per your suggestion):**
- Collapsed the three formats (JSON, ARROW, ARROW_STREAM) into two that match the API enums: `JSON_ARRAY` and `ARROW_STREAM`
- `ARROW_STREAM` now supports both INLINE (with attachment decoding) and EXTERNAL_LINKS dispositions
- Default remains `JSON_ARRAY` as you requested
- Added format fallback: if ARROW_STREAM+INLINE is rejected by a classic warehouse, it falls back to JSON_ARRAY automatically
- Updated the UI hooks to use `json_array` / `arrow_stream` naming

Also added 147 new unit tests covering major coverage gaps (service-context 7%→100%, stream-registry 32%→100%, genie connector 61%→97%, files plugin 69%→89%). All 1711 tests pass.

Cleaned up the branch — it's now TS-only, no unrelated changes.

jamesbroadhead:
Hi — Claude here, on James's behalf.

Closing in favour of a 3-PR stack carved out of this branch. Same end-state, easier to review:

1. #327 — `test: backfill coverage for genie connector, service context, stream registry` (pure test additions, ~1825 lines)
2. #328 — `refactor: rename AnalyticsFormat to API enum names` (mechanical `JSON`/`ARROW` → `JSON_ARRAY`/`ARROW_STREAM`)
3. #329 — `feat: decode inline Arrow IPC + warehouse-compat fallback` (the actual fix — `arrow-schema.ts`, inline IPC decode, format fallback, zod-validated SSE wire protocol)

Merge order: #327 → #328 → #329. Verified the sum of the three branches matches the PR-touched files on this branch exactly (0 diffs across all 30 files).

PR #261, opened 2026-04-08
feat: new manifest format

atilafassina:
## Summary

Evolves the AppKit plugin manifest contract from v1.0 to v2.0 and swaps the canonical authoring surface from hand-written JSON Schema to **Zod via the Standard Schema interface**. AJV, the `json-schema-to-typescript` codegen pipeline, and the standalone semantic-validator delete; refinement and `.transform()` on Zod schemas replace them.

Mid-review (2026-05-18) the design picked up a **substitutability-gate amendment** that deleted `postScaffold` entirely in favor of a unified plugin-level `scaffolding.rules` block, and audited every prose rule against the gate. See "2026-05-18 amendment" below.

> [!IMPORTANT]
> **~91% of insertions are auto-regenerated artifacts** — two JSON Schema files (`docs/static/schemas/{plugin-manifest,template-plugins}.schema.json`) account for `+10,400` lines.

`.gitattributes` marks all generated artifacts `linguist-generated=true`. **GitHub collapses them in the file list** — substantive review surface is ~32 files / ~2,300 insertions.

## What changes vs `main`

**Contract**
- `discoveryDescriptor` is a discriminated union: `{ type: "kind", resourceKind, ... } | { type: "cli", cliCommand, ... }`. Six known kinds: `warehouse`, `genie_space`, `volume`, `postgres_project`, `postgres_branch`, `postgres_database`.
- `cli` variant rejects shell metacharacters (`;`, `|`, `&`, backtick, `$`, newlines) on `cliCommand` and `shortcut`. `.describe()` notes direct authors to `kind` for first-party resources and flag the `cli` shape as minimal-and-may-tighten.
- `RESOURCE_KIND_COMMANDS` map next to the schema; commands verified against Databricks CLI v0.299.0. **`volume` kind carries `parents: ["catalog", "schema"]`** — declares free-text prompts the runner must collect before invoking discovery (replaces the prior prose rule).
- `origin` on template fields is a `.transform()` output (drift-impossible).
- `rules.must` / `rules.should` / `rules.never` items capped at `maxLength: 120`. Per-bucket and cross-bucket dedup enforced via `.superRefine()`.
- 4 core plugin manifests (analytics, files, genie, lakebase) migrated to the kind variant; lakebase gains a `project` field.

**Authoring surface**
- New: `packages/shared/src/schemas/manifest.ts` (Zod schemas + `RESOURCE_KIND_COMMANDS` + `TEMPLATE_SCAFFOLDING`).
- New: `tools/generate-json-schema.ts` (Zod → draft-07 JSON Schema, emits to `docs/static/schemas/`).
- New runtime dep on `packages/shared`: `@standard-schema/spec`.

**Deletions**
- Deps removed: `ajv`, `ajv-formats`, `json-schema-to-typescript`.
- `packages/shared/src/schemas/{plugin-manifest.schema.json,template-plugins.schema.json,plugin-manifest.generated.ts}`.
- `tools/generate-schema-types.ts`, `docs/scripts/copy-schemas.ts`.
- `validate-manifest.ts` (498 → 177 lines): AJV plumbing + `runSemanticValidation` + helpers.
- `enrichFieldsWithOrigin`, `validateDiscoveryOrigin`.
- **`postScaffoldStepSchema` and `postScaffold` field** (post-amendment — see below).

**Migrated to Zod-direct**
- `schema-resources.ts` and `tools/generate-registry-types.ts` (no more runtime JSON-schema reads).
- `manifest-types.ts` is now a re-export shim (`z.infer` types + `StandardSchemaV1`).

## Plugin Rules (per core plugin, `scaffolding.rules`)

| Plugin | must | should |
|---|---|---|
| analytics | Before init, ensure the SQL Warehouse passed via `--set analytics.sql-warehouse.id` is running | After init, ensure `config/queries/` has at least one `.sql` file before running `npm run typegen` |
| files | Before init, verify your Unity Catalog volume exists and you have `WRITE_VOLUME` permission | — |
| genie | After init, configure the `'spaces'` map in plugin config with alias-to-Space-ID mappings | — |
| jobs | *(no rules)* | *(no rules)* |
| lakebase | After init, run database migrations via `'pnpm drizzle:migrate'` before first request | After init, verify Lakebase connectivity with `'psql $PGHOST -c "select 1"'` |

## Template-level scaffolding.rules

```json
{
  "must": [
    "Keep all secrets and credentials only in app.yaml, databricks.yml, and/or .env"
  ],
  "should": [
    "ask user when in doubt of resource to use for plugin"
  ],
  "never": [
    "guess resources when multiple or no options are available",
    "embed secrets in files that will go to the client-bundle"
  ]
}
```

Each entry passes the substitutability gate (describes a cross-cutting agent behavior at decision time, not derivable from structured manifest data). Each ≤120 chars. Conflict-detection-safe against the current plugin rule corpus (no `must` ↔ `never` overlaps).

## 2026-05-18 amendment — substitutability gate

The original v2.0 plan shipped `postScaffold` arrays (`{ instruction, required }`) at the plugin level. During v2.0 implementation we found the format couldn't participate in the `databricks-apps` agent skill's command-construction flow — prose doesn't compose across plugins, duplicates work the skill drives from structured resources, and `{ instruction, required }` adds zero parseable signal.

The amendment introduces one design principle and applies it uniformly:

> **A prose rule in the manifest is legitimate only if it cannot be expressed as structured data on a resource, plugin config, or template.**

Changes that landed under the amendment:

- **`postScaffold` deleted.** `postScaffoldStepSchema` and the `postScaffold` field removed from both `pluginManifestSchema` and the template plugin schema. All four core plugins migrated to plugin-level `scaffolding.rules`.
- **Plugin-level `scaffolding.rules`** with `must` / `should` / `never` arrays (≤120 chars each, no duplicates within a bucket, no overlap across buckets — all enforced via `.superRefine()`).
- **Template-level `scaffoldingRulesSchema` gains `should[]?`** for parity with plugin-level.
- **TEMPLATE_SCAFFOLDING audit** — original 4 `must` entries pruned (every one was substitutable: skill content, derivable from `requiredByTemplate`, derivable from `field.env`, or a discovery-descriptor concern). Original 3 `never` entries replaced with 2 sharper rules covering credentials hygiene and resource-selection decision points; one entry deleted as unreachable. See "Template-level scaffolding.rules" above.
- **A6 schema-extension decisions:**
  - (a) Volume catalog+schema prerequisite → **EXTEND**: `RESOURCE_KIND_COMMANDS.volume.parents = ["catalog", "schema"]`. Models transient query inputs as a kind-level property without conflating with `dependsOn` (which references sibling fields).
  - (b) Config-driven `--set` derivation for genie spaces → **ACCEPT AS PROSE** pending Track B confirmation of the skill's consumption model.
  - (c) Liveness preconditions (analytics warehouse) → **ACCEPT AS PROSE**. Runtime state checks aren't structural data.
- **Per-plugin rule audit:** 7 substitutability-gate violations introduced during experimentation were caught and reverted (permission duplications, resource-existence tautologies, inactionable-at-scaffold-time directives). Net effect: zero new plugin rules survive the gate that weren't already in the post-Phase-1 baseline.

## `.claude/commands/` skill extensions

Outside the `packages/` tree but relevant to reviewers — the in-repo plugin commands gained procedural v2.0 semantic checks:

- **`audit-core-plugin.md`** — Step 3.6 (3.6a substitutability-gate patterns, 3.6b discovery descriptor completeness, 3.6c `parents` vs `dependsOn` consistency). All findings feed Category 1 (Manifest Design).
- **`review-core-plugin.md`** — Step 5.6 with the same checks scoped to changed manifest files only.
- **`create-core-plugin.md`** — Step 4e.1 post-scaffold enrichment that walks each user-supplied field to add a `discovery` block and gates any `scaffolding.rules` against 5 explicit reject patterns.

No changes to `plugin-best-practices.md` or `plugin-review-guidance.md` — the additions are procedural workflow steps, not narrative rules.

## Track B — consumer skill (cross-repo)

The `databricks-apps` skill in [`databricks/databricks-agent-skills#79`](https://github.com/databricks/databricks-agent-skills/pull/79) ("docs(apps): update AppKit scaffolding setup for Manifest 2.0") adds a new "Scaffolding Rules Protocol" subsection to `skills/databricks-apps/SKILL.md` (skill version bump `0.1.1 → 0.1.2`). It consumes the rules emitted by this PR:

1. **Gather** rules from selected `--features` plugins + every plugin with `requiredByTemplate: true` + template-level
2. **Precedence** — manifest rules override skill-baked directives
3. **Phase ordering** — `Before init` / `After init` prefixes
4. **Conflict detection** — plugin `must` ↔ template `never` stops the flow
5. **Reporting** — surface merged guardrails to the user before init

**Sequencing:** Track A (this PR) ships first. Track B is **opened, not blocking merge** — plugin-level rules are advisory metadata in the transition state. Failure mode if Track B lands later: degraded (third-party plugin guardrails don't fire), not broken (resources still drive the build).

End-to-end integration runbook lives in `~/.xavier/research/manifest-2-skill-integration-test.md` (vault note, not in this repo).

## Deferred to follow-up PRs

| Change | Note |
|---|---|
| **TypeScript-authored manifests** (`defineManifest({ ... })`) | Future work; no stacked PR yet |
| Docs for the v2.0 format | [#370](https://github.com/databricks/appkit/pull/370) (open, branch `manifest-docs`) |
| **Branded field-name keys** for compile-time `dependsOn` checks | Not a must-have, larger implementation change |
| `cli` variant full hardening | Shell-metacharacter denylist + describe notes shipped here. Remaining: argv-array form (executor-side spawn-not-shell), output-shape contract (replace jq path with declared field names + unwrap). Tied to the executor PR. |
| Track B skill PR | [`databricks-agent-skills#79`](https://github.com/databricks/databricks-agent-skills/pull/79) — opens for skill author review; not required to merge before Track A |

> [!NOTE]
> Plugin authors continue to write `manifest.json`. See the docs PR ([#370](https://github.com/databricks/appkit/pull/370)) for the authoring guide.

## Note on the generated files diff

1. Zod emits keys in a different order than the hand-written JSON Schema did. For example, you'll see "description" flip from before "type" to after "type", and "required": [...] arrays appear in different positions. `z.toJSONSchema()` has its own key ordering, and that's what now writes the file.
2. Discovery is reshaped as a discriminated union. That's why `$ref: "#/$defs/resourceRequirement"` got expanded into giant `oneOf: [...]` blocks per resource type — the schema's actual shape changed, not just its layout.

Biome's JSON formatter only normalizes whitespace; it doesn't reorder keys. So the noise you're seeing is the cost of moving from hand-authored JSON Schema → Zod-generated JSON.


keugenek:
@atilafassina this one looks ok and evolution, however little heavy code-wise (let me post comments from isaac about these) and mixing concepts little bit. What do you think of this:

 Alternatives worth weighing                                                   
                  
  Zod as single source of truth (my default for a TS-first ecosystem):          
  - Types, runtime validation, and semantic checks in one file.
  - zod-to-json-schema emits the $schema JSON for VSCode intellisense.          
  - Cycle/dangling-ref checks become .refine() calls in the same module as the
  schema.                                                                       
  - Deletes plugin-manifest.generated.ts and the Ajv layer.                     
                                                                                
  TS-authored manifests compiled to JSON:                                       
  - manifest.ts → defineManifest({...}) with branded types for field names      
  (compile-time dependsOn checking, no DFS needed).                             
  - sync lowers to .json for non-TS consumers.                                  
  - Strongest author ergonomics; agent still reads JSON.                        
                                                                                
  Split concerns into two files:                                                
  - manifest.json — strict contract (resources, fields, env), stable schema.    
  - scaffold.yaml or .md — agent hints (discovery, prompts, postScaffold), free 
  to evolve.                                                                   
  - Today they're fused, which is why the schema keeps growing.

docs/static/schemas/plugin-manifest.schema.json
```diff
@@ -478,6 +489,60 @@
           "type": "boolean"
         }
       }
+    },
+    "discoveryDescriptor": {
```
keugenek, line 493: `cliCommand` is a string template with a `<PROFILE>` placeholder plus a jq path in `selectField`. The only validation is "contains `<PROFILE>`" — nothing checks the CLI output shape, so paginated or wrapped responses (`{warehouses:[...]}`) will fail at runtime. Consider `{resourceKind, select, display}` and letting the CLI own the command.

packages/shared/src/cli/commands/plugin/sync/sync.ts
```diff
@@ -36,6 +40,29 @@ function isWithinDirectory(filePath: string, boundary: string): boolean {
   );
 }
 
+/**
+ * Injects computed `origin` onto every resource field in all plugins.
+ * Mutates the plugins object in place for efficiency.
+ */
+function enrichFieldsWithOrigin(
```
keugenek, line 47: `origin` is derived from other fields then serialized back, which is why `validateDiscoveryOrigin` has to warn about inconsistent states. Computing at read time would remove both the enrichment step and that validator.

docs/static/schemas/template-plugins.schema.json
```diff
@@ -98,10 +120,330 @@
...
+        },
+        "must": {
+          "type": "array",
+          "items": { "type": "string" },
+          "description": "Actions the scaffolding agent must always perform."
+        }
+      },
+      "additionalProperties": false
+    },
+    "scaffoldingDescriptor": {
```
keugenek, line 425: `rules.never[]` / `rules.must[]` are LLM prompts inside the schema. No validator can check what actually matters (agent behavior), and the schema file becomes part prompt. Worth splitting into a sibling `scaffold.prompts.md`.

docs/static/schemas/template-plugins.schema.json
```diff
@@ -98,10 +120,330 @@
       "$ref": "plugin-manifest.schema.json#/$defs/resourceType"
     },
     "resourceFieldEntry": {
```
keugenek, line 122: Per commit `eb65776`, `resourceFieldEntry` / `resourceRequirement` were inlined instead of resolving `$ref`. The plugin and template schemas will drift. Better to fix the ref resolver (e.g. bundle via `@apidevtools/json-schema-ref-parser` at publish).

keugenek (commented):
Overall direction looks right — v2.0 versioning via `if/then` is correct and co-located `manifest.json` per plugin is the right locality. A few shape-level concerns worth addressing before or soon after this lands. Happy to chat on any of them.

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"f1aff2d6-c2fa-41a1-9596-0670b1e35bda","pid":6435,"acquiredAt":1775737052293}
```
keugenek, line 1: Accidental commit — add to `.gitignore`.

docs/static/schemas/template-plugins.schema.json
atilafassina: My reluctance to an `*.md` is that this will create contention between that and the plugin skill. I don't want freeform text here. These keys are short directives for things the plugin needs the LLM to **always** or **never** do in order to function efficiently.

So, while we can't validate the `string` fully, I'm capping it to a max-length. That and a clear validation error will nudge the plugin author in the right direction.

docs/static/schemas/plugin-manifest.schema.json
atilafassina: I'll change the 
    - `cliCommand` gets a `.refine()` denylist of shell operators
    - cli variant's `.describe()` gains a "use kind where possible; cli is an escape hatch and may tighten further" note. Surfaces in the published JSON. This mains we can tighten it up / deprecate going forward.


I'm a bit reluctant about making it more structured/strict right now because we are still a bit short in plugin quantity, so new use-cases and usage may surface here.

Wdyt?

atilafassina: Shipped both commitments from the May 8 reply:

- Shell-operator denylist landed in [`6f23cd8f`](https://github.com/databricks/appkit/commit/6f23cd8f) ("fix: deny shell metacharacters on cli discovery descriptor") — applies to both `cliCommand` and `shortcut`.
- The `.describe()` note on the `cli` variant flagging it as "intentionally minimal and may tighten further" ships in `packages/shared/src/schemas/manifest.ts` (around line 243).

Full output-shape contract (replace jq path with declared field names + unwrap) and the argv-array form (executor-side spawn-not-shell) are correctly deferred to the executor PR — captured under "Deferred — `cli` variant full hardening" in the updated PR description. Resolving here; reopen if you want output-shape concerns tracked separately.

docs/static/schemas/template-plugins.schema.json
atilafassina: Following up post-amendment (2026-05-18) — the rules now have a formal contract that addresses the "unverifiable prompt" concern:

- **Substitutability gate**: each rule must justify itself against a written principle ("a prose rule is legitimate only if it cannot be expressed as structured data on a resource, plugin config, or template"). See the new "2026-05-18 amendment" section in the PR description.
- **120-char cap + `.superRefine()` dedup**: enforces "short directive by contract" at the schema level — no duplicates within a bucket, no overlap across `must` / `should` / `never`.
- **Formal consumer**: [`databricks-agent-skills#79`](https://github.com/databricks/databricks-agent-skills/pull/79) adds a 5-step Scaffolding Rules Protocol (gather → precedence → phase ordering → conflict detection → reporting) to `skills/databricks-apps/SKILL.md`. The rules have a documented protocol now, not just freeform text.

The TEMPLATE_SCAFFOLDING audit (commits [`a8ba1c46`](https://github.com/databricks/appkit/commit/a8ba1c46) + [`9c2ee83c`](https://github.com/databricks/appkit/commit/9c2ee83c)) also pruned the original 4 unsubstantiated `must` entries and reshaped the `never` set to match Track B's conflict-detection step. Resolving here; happy to reopen if the gate/consumer story still feels under-specified.

packages/appkit/src/plugins/analytics/manifest.json
```diff
@@ -14,13 +14,27 @@
...
           }
         }
       }
     ],
     "optional": []
   },
+  "scaffolding": {
+    "rules": {
+      "must": [
+        "Before init, ensure the SQL Warehouse passed via --set analytics.sql-warehouse.id is running"
```
MarioCadenas, line 31: how does the agent know how to do this? should we add some command?

packages/appkit/src/plugins/files/manifest.json
```diff
@@ -14,13 +14,25 @@
...
           }
         }
       }
     ],
     "optional": []
   },
+  "scaffolding": {
+    "rules": {
+      "must": [
+        "Before init, verify your Unity Catalog volume exists and you have WRITE_VOLUME permission"
```
MarioCadenas, line 32: same as the comment before, how does it know how to do it?

packages/shared/src/cli/commands/plugin/sync/sync.ts
```diff
@@ -529,17 +539,40 @@ async function scanPluginsDir(
...
+        }
+      }
+    }
+  }
+
   const templateManifest: TemplatePluginsManifest = {
     $schema:
       "https://databricks.github.io/appkit/schemas/template-plugins.schema.json",
-    version: "1.1",
+    version: "2.0",
```
MarioCadenas, line 573: do we need to bump this to 2.0? is just incremental no? no breaking changes

tools/generate-json-schema.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Run from repo root: pnpm exec tsx tools/generate-json-schema.ts
+ */
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { z } from "zod";
+import {
+  pluginManifestSchema,
+  templatePluginsManifestSchema,
+} from "../packages/shared/src/schemas/manifest.ts";
```
MarioCadenas: can we remove the extensions here? (editor complaining)

tools/generate-registry-types.ts
```diff
@@ -1,109 +1,109 @@
...
+  postgresPermissionSchema,
+  resourceTypeSchema,
+  secretPermissionSchema,
+  servingEndpointPermissionSchema,
+  sqlWarehousePermissionSchema,
+  ucConnectionPermissionSchema,
+  ucFunctionPermissionSchema,
+  vectorSearchIndexPermissionSchema,
+  volumePermissionSchema,
+} from "../packages/shared/src/schemas/manifest.ts";
```
MarioCadenas: extension too

template/appkit.plugins.json
```diff
@@ -218,13 +302,43 @@
...
+  "scaffolding": {
+    "command": "databricks apps init",
+    "flags": {
+      "--template-dir": {
+        "description": "Path to the template directory containing the app scaffold",
+        "required": true
+      },
+      "--config-dir": {
+        "description": "Path to the output directory for the initialized app",
+        "required": true
```
MarioCadenas: these 2 flags are not required as far as I know no?

template/appkit.plugins.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://databricks.github.io/appkit/schemas/template-plugins.schema.json",
-  "version": "1.1",
+  "version": "2.0",
```
MarioCadenas, line 3: 1.2 maybe?

docs/docs/api/appkit/Interface.PluginManifest.md
```diff
@@ -213,14 +197,60 @@ Resources that must be available for the plugin to function
...
+
+***
+
 ### stability?
 
 ```ts
 optional stability: "beta" | "ga";
 ```
 
-Plugin stability level. Beta plugins may have breaking API changes between minor releases but are on a path to GA. GA (general availability) plugins follow semver strictly.
```
MarioCadenas, line 222: why is this one removed?

packages/appkit/src/plugins/lakebase/manifest.json
```diff
@@ -60,5 +81,15 @@
       }
     ],
     "optional": []
+  },
+  "scaffolding": {
+    "rules": {
+      "must": [
+        "After init, run database migrations via 'pnpm drizzle:migrate' before first request"
```
MarioCadenas: I don't think the template has this command in the scripts in package.json no? also it doesn't use pnpm

packages/appkit/src/plugins/analytics/manifest.json
atilafassina: It has just used the CLI (`databricks apps manifest`) so it's right in their context to find the command. In my xp it's capable of that.

I'm torn between adding a CLI command in the manifest to make it more effectivebecause such a contract is not enforced anywhere. I'm **assuming** that CLI commands are sufficiently self discoverable.

packages/appkit/src/plugins/files/manifest.json
atilafassina: It has just used the CLI (`databricks apps manifest`) so it's right in their context to find the command. In my xp it's capable of that.

I'm torn between adding a CLI command in the manifest to make it more effectivebecause such a contract is not enforced anywhere. I'm **assuming** that CLI commands are sufficiently self discoverable.

packages/shared/src/cli/commands/plugin/sync/sync.ts
atilafassina: you're right if we only look at the schema, but the validation treats the some of the new props as required, so version `1.1` will fail `2.0` validation.

If we bump to `1.2` the strict way would be to make the validation looser.

template/appkit.plugins.json
atilafassina: same as above [https://github.com/databricks/appkit/pull/261#discussion_r3275770459](https://github.com/databricks/appkit/pull/261#discussion_r3275770459)

we can, but validation would need to be less strict.

docs/docs/api/appkit/Interface.PluginManifest.md
atilafassina: I think this was an accident when I split docs PR from this one. Let me revert it.

packages/appkit/src/plugins/lakebase/manifest.json
atilafassina: replacing this wiht a package and ORM agnostic rule.

"run migrations". I'll also downgrade it to a `should`

packages/appkit/src/plugins/files/manifest.json
MarioCadenas: yep they are, but providing the instruction or direct command will make the agent execute faster and consume less tokens in the end, but totally optional

template/appkit.plugins.json
```diff
@@ -218,13 +300,43 @@
...
+        "required": false
+      },
+      "--config-dir": {
+        "description": "Path to the output directory for the initialized app",
+        "required": false
+      },
+      "--profile": {
+        "description": "Databricks CLI profile to use for authentication",
+        "required": false
+      }
```
MarioCadenas, line 357: if we are adding flags here we should probably add the `--name` flag which is the only flag required no?

MarioCadenas (approved):

template/appkit.plugins.json
atilafassina: Good catch — `--name` is now in and marked required (the scaffolder uses it to populate `{{.projectName}}` in `package.json`, `databricks.yml`, and `.env`, so every template file depends on it).

Took the opportunity to clean the full set:
- **Added** (`required: false`): `--template`, `--version`, `--features`, `--set`, `--output-dir`, `--description`, `--run`, `--auto-approve`
- **Removed**: `--template-dir` and `--config-dir` (not real `apps init` flags — they were stale), and `--profile` (global flag, not init-specific — agents can discover via `--help`)
- **Excluded intentionally**: `--branch` (niche GitHub-template-only, mutually exclusive with `--version` in a way the schema cannot express), `--deploy` (touches shared workspace — should require explicit user intent), `--warehouse-id` (deprecated → use `--set`), `--plugins` (hidden alias for `--features`)

One non-obvious call: `--auto-approve` is **not** required, even though omitting it can hang non-interactive agent runs. It conflicts with the existing rule *"ask user when in doubt of resource to use for plugin"* — marking it required would tell agents to skip the very disambiguation the rules ask for.

packages/appkit/src/plugins/files/manifest.json
atilafassina: I did a small assessment / check and the difference between the hardocoding the command and letting command be inferred is estimated around 100tokens

also, at this stage we don't have a way to validate the command so it's hard to flag if the hardcoded command and the released version drift. So I made a note and will consider addressing this as a follow-up depending on how evals behave

PR #264, opened 2026-04-10
feat(plugin): add stability system (beta/stable) with promote command

MarioCadenas:
Adds a `stability` field to plugin manifests with import-path enforcement and full CLI tooling. Two tiers (`beta ──→ stable`); promotion is one-way. Importing from `@databricks/appkit/beta` is explicit consent to breaking-change risk; `@databricks/appkit` follows semver strictly.

## Highlights

- **Schemas** — `stability` enum on `plugin-manifest.schema.json` and `template-plugins.schema.json`; template manifest bumped to `1.1`.
- **Subpath exports** — `@databricks/appkit/beta` and `@databricks/appkit-ui/{js,react}/beta` wired through `package.json` and `tsdown.config.ts`.
- **Manifest as single source of truth** — `tools/generate-plugin-entries.ts` reads each plugin's `manifest.json` and writes committed `stable-exports.generated.ts` / `beta-exports.generated.ts` barrels. `src/index.ts` and `src/beta.ts` delegate via `export *`. CI fails if any of the four files (the two barrels, the synced `template/appkit.plugins.json`, and the existing `plugin-manifest.generated.ts`) drift from the manifests.
- **`plugin sync` / `list` / `create`** — propagate `stability`, strip `requiredByTemplate` for non-stable plugins, surface a `STABILITY` column, prompt for tier on create.
- **`plugin promote`** *(new)* — flips manifest, rewrites import paths in user `.ts`/`.tsx`, regenerates barrels (monorepo only), runs sync. Hardened: schema-regex name validation, `isWithinDirectory` bounds, symlink skip, `--allow-installed` for `node_modules`, `--dry-run`, `--skip-imports`, `--skip-sync`.
- **Init template** — `template/server/server.ts` branches on `$p.Stability` and emits a single consolidated `from '@databricks/appkit/beta'` import for beta plugins. Pure `text/template`, no sprig.
- **Docs** — new [Plugin Stability Tiers](docs/docs/plugins/stability.md) page.

## Behavior contract

- Absent `stability` ⇒ `"stable"`. Unknown values rejected up front.
- Demotion rejected with a clear error.
- Only stable plugins can be `requiredByTemplate`.
- Failed post-promote regen or sync exits non-zero — no silent drift.

## Companion CLI

[databricks/cli#5090](https://github.com/databricks/cli/pull/5090) exposes `pluginVar.Stability` to the init template and labels beta plugins in the picker. End-to-end verified against real local tarballs.

## Test plan

- `pnpm build && pnpm -r typecheck && pnpm check:fix && pnpm test && pnpm docs:build` — all clean
- 38 promote integration tests cover invalid names, path traversal, unknown tiers, demotion, dry-run, import rewriting, symlink skipping, and `node_modules` protection.
- CI freshness checks fail the PR if the auto-generated barrels or the synced `template/appkit.plugins.json` drift from the source manifests.

## Demo

Demo shows analytics and serving as beta (just as an example) to test the functionality

https://github.com/user-attachments/assets/f6bff990-84f8-4d77-a2a8-e0d3db582e7a

image


packages/shared/src/cli/commands/plugin/sync/sync.ts
```diff
@@ -84,6 +84,10 @@ async function loadPluginEntry(
       ...(manifest.onSetupMessage && {
         onSetupMessage: manifest.onSetupMessage,
       }),
+      ...(manifest.stability &&
+        manifest.stability !== "stable" && {
```
atilafassina: do we need this 2nd check and the assertion?
this + the assertion below can cause some confusion if we add a 3rd value in the future

`manifest.stability = "alpha"`
but then types will say it's "beta"

packages/shared/src/cli/commands/plugin/manifest-types.ts
```diff
@@ -17,6 +17,8 @@ export interface TemplatePlugin extends Omit<PluginManifest, "config"> {
   package: string;
   /** When true, this plugin is required by the template and cannot be deselected during CLI init. */
   requiredByTemplate?: boolean;
+  /** Plugin stability level. Absent or undefined means "stable". */
+  stability?: "beta" | "stable";
```
atilafassina: i like that this is backwards compatible for now, but imho: moving forward we should consider moving this to a required key. The fallback from `undefined`  seems error-prone.

atilafassina (dismissed):

packages/shared/src/cli/commands/plugin/promote/promote.ts
```diff
@@ -0,0 +1,446 @@
...
+  }
+  return null;
+}
+
+function loadManifestFromFileSync(filePath: string): unknown {
+  const raw = fs.readFileSync(filePath, "utf-8");
+  return JSON.parse(raw);
+}
+
+function rewriteImportsInFile(
```
atilafassina, line 221: I think this function is rewriting **every** beta import to a stable one. Not only the ones from the plugin being promoted.

packages/shared/src/cli/commands/plugin/sync/sync.ts
MarioCadenas: Good catch — dropped the `as "beta"` cast in `ced37e60`. The narrowing on `!== "stable"` plus the truthy check already produces the non-stable variant; TypeScript infers `"beta"` today and would infer `"alpha" | "beta"` automatically the moment a third tier is added, so the code stays honest without revisiting this site.

packages/shared/src/cli/commands/plugin/promote/promote.ts
MarioCadenas: You're absolutely right — confirmed and fixed in `ced37e60`.

Pre-fix the rewriter did `content.split(oldPath).join(newPath)`, which rewrote the entire `from "..."` source for the whole import line. Promoting only `betaA` would move `betaB` along with it onto the stable subpath where `betaB` doesn't exist:

```ts
// before
import { betaA, betaB } from "@databricks/appkit/beta";
// promote betaA --to stable, pre-fix (BROKEN — betaB now undefined)
import { betaA, betaB } from "@databricks/appkit";
```

Now `rewriteImportsInFile` takes the plugin name, parses each import statement, finds the targeted specifier (handling `name`, `name as alias`, `type name`, and the kebab-to-camelCase manifest convention like `vector-search` → `vectorSearch`), and either rewrites the source if it's the only specifier or splits the import:

```ts
// after
import { betaB } from "@databricks/appkit/beta";
import { betaA } from "@databricks/appkit";
```

Eight new unit tests cover single-specifier rewrite, multi-specifier split, no-op when the plugin is not in the file, `import type`, aliased specifiers, multi-line specifier lists, and multiple packages in the same file. The existing `runPromote > rewrites imports across .ts and .tsx files` integration test was rewritten to lock in the contract that unrelated beta imports stay put.

packages/shared/src/cli/commands/plugin/manifest-types.ts
MarioCadenas: why do you see it being error prone?

docs/docs/plugins/stability.md
```diff
@@ -0,0 +1,155 @@
...
+```bash
+# Edit packages/appkit/src/plugins/<name>/manifest.json
+# Set "stability": "beta" (or remove the field for stable)
+pnpm run generate:plugin-entries  # regenerate the barrels
+pnpm sync:template                # regenerate appkit.plugins.json
+```
+
+## Current Plugins by Tier
+
+All built-in plugins are currently **stable**:
```
pkosiec: What about model serving and jobs? this list seems to be outdated; maybe we should get rid of this section?

pkosiec (commented):
To be honest I'm still not a fan of introducing the full changes right now - why:
- AppKit is 0.x, which doesn't mean it gained the full stability
- with the whole promotion system the plugin development will be slower: think about all the discussions about "what does it mean that a plugin is stable"; even right now it's hard to say what "stable" means. We're looking for user feedback and we might change plugins.
- I don't think we can confidently say that all plugins are stable - based on user feedback we might change some significantly; we have even internal discussions about 
- Usually "beta" means the API is already established and it won't change radically. So the "beta" level is already too high for e.g. Agent plugin, if we'll really consider it experimental.

I'm not saying it's not useful, it is, but I'd postpone the full mechanism as much as possible.

---

Instead, what do you think about something very lightweight: label some really early plugins as "experimental". No "stable" terminology at all. Later on we can add more stability levels.

Why?
1. It matches 0.x reality - we're not promising any plugin is "stable," you're just flagging which ones are especially likely to change.
2. It's the minimum viable distinction - one label, not a promotion system.
3. It reduces process overhead - no debates about what "stable" or "graduated" means. A plugin is either experimental or it's not.
4. Precedent: TypeScript itself uses --experimental* flags, Node.js marks APIs as "Experimental" with no formal "stable" label at the other end (just "not experimental").

WDYT?

pkosiec (approved):

PR #221, opened 2026-03-31
[DRAFT](do not merge) jobs plugin

keugenek:
## Summary

New AppKit plugin wrapping the Databricks Jobs API with typed methods, integrated telemetry, retry, and timeout.

### Plugin API

| Method | Description |
|--------|-------------|
| `submitRun(req)` | Submit a one-time run |
| `runNow(req)` | Trigger an existing job |
| `getRun(runId)` | Get run metadata |
| `getRunOutput(runId)` | Get task output |
| `cancelRun(runId)` | Cancel a run |
| `listRuns(req)` | List runs for a job |
| `getJob(jobId)` | Get job details |
| `createJob(req)` | Create a new job |
| `waitForRun(runId, opts?)` | Poll until terminal state (exponential backoff + jitter) |

### Usage

\`\`\`ts
import { createApp, jobs, server } from "@databricks/appkit";

const app = await createApp({
  plugins: [server(), jobs()],
});

const { run_id } = await app.jobs.submitRun({
  run_name: "my-analysis",
  tasks: [{
    task_key: "main",
    notebook_task: { notebook_path: "/Users/me/analysis" },
  }],
});

const run = await app.jobs.waitForRun(run_id);
console.log(run.state?.result_state); // "SUCCESS"
\`\`\`

### What's included

- Jobs connector: \`packages/appkit/src/connectors/jobs/\` (client with telemetry spans, exponential backoff polling)
- Jobs plugin: \`packages/appkit/src/plugins/jobs/\` (plugin, types, manifest)
- Unit tests: plugin factory + exports validation
- E2E scenario tests: Mock Jobs API server with Playwright (4 public, 7 private cases)
- Documentation: \`docs/docs/plugins/jobs.md\`

### Design decisions

- **Wraps \`@databricks/sdk-experimental\` Jobs API** — all SDK types flow through
- **Connector pattern** — matches sql-warehouse connector with per-call telemetry spans
- **\`waitForRun\` with exponential backoff + jitter** — prevents thundering herd on shared clusters
- **No resources required** — uses workspace auth from AppKit context
- **\`InitializationError\`** — matches files plugin error pattern
- **Terminal states as \`Set\`** — TERMINATED, SKIPPED, INTERNAL_ERROR checked via \`.has()\`

## Documentation safety checklist

- [x] Examples use least-privilege permissions
- [x] Elevated permissions explicitly called out where required
- [x] Sensitive values obfuscated (placeholder workspace IDs, URLs, no real tokens)
[...]

PR #265, opened 2026-04-10
feat: Lakeflow Jobs Plugin

atilafassina:
> [!IMPORTANT]
> to maintain contribution history this PR should be rebased when merged
> (don't merge it if it has more than 2 commits)

## Summary

Resource-scoped jobs plugin following the files plugin pattern. Jobs are configured as named resources discovered from environment variables at startup.

### Design

- **Resource-scoped**: Only configured jobs are accessible — this is not an open SDK wrapper
- **Env-var discovery**: Jobs are discovered from `DATABRICKS_JOB_<KEY>` env vars (e.g. `DATABRICKS_JOB_ETL=123`)
- **Single-job shorthand**: `DATABRICKS_JOB_ID` maps to the `"default"` key
- **Manifest declares job resources** with `CAN_MANAGE_RUN` permission
- **Works with `databricks apps init --features jobs`**

### API

```ts
// Trigger a configured job
const { run_id } = await appkit.jobs("etl").runNow();

// Trigger and wait for completion
const run = await appkit.jobs("etl").runNowAndWait();

// OBO access
await appkit.jobs("etl").asUser(req).runNow();

// List recent runs
const runs = await appkit.jobs("etl").listRuns({ limit: 10 });

// Single-job shorthand
await appkit.jobs("default").runNow();
```

### Files changed

- `plugins/jobs/manifest.json` — declares job resource with `CAN_MANAGE_RUN` permission
- `plugins/jobs/types.ts` — `JobAPI`, `JobHandle`, `JobsExport`, `IJobsConfig` types
- `plugins/jobs/plugin.ts` — `JobsPlugin` with `discoverJobs()`, `getResourceRequirements()`, resource-scoped `createJobAPI()`
- `plugins/jobs/index.ts` — barrel exports
- `connectors/jobs/client.ts` — `listRuns` now respects `limit` parameter
- `plugins/jobs/tests/plugin.test.ts` — 32 tests covering discovery, resource requirements, exports, OBO, multi-job, and auto-fill

## Documentation safety checklist
- [x] Examples use least-privilege permissions
- [x] Sensitive values are obfuscated
- [x] No insecure patterns introduced

> _Reopened from #223 — moved head branch from fork to upstream to fix CI secrets access._

## demo

https://github.com/user-attachments/assets/fa1b2bbd-bb09-4a84-b833-9d913819659b

### DIFF
--- a/docs/docs/plugins/index.md
+++ b/docs/docs/plugins/index.md
@@ -13,7 +13,7 @@ For complete API documentation, see the [`Plugin`](../api/appkit/Class.Plugin.md
 Configure plugins when creating your AppKit instance:
 
 ```typescript
-import { createApp, server, analytics, genie, files } from "@databricks/appkit";
+import { createApp, server, analytics, genie, files, jobs } from "@databricks/appkit";
 
 const AppKit = await createApp({
   plugins: [
@@ -21,6 +21,7 @@ const AppKit = await createApp({
     analytics(),
     genie(),
     files(),
+    jobs(),
   ],
 });
 ```
--- a/packages/appkit/package.json
+++ b/packages/appkit/package.json
@@ -77,7 +77,8 @@
     "semver": "7.7.3",
     "shared": "workspace:*",
     "vite": "npm:rolldown-vite@7.1.14",
-    "ws": "8.18.3"
+    "ws": "8.18.3",
+    "zod": "4.3.6"
   },
   "devDependencies": {
     "@types/express": "4.17.25",
--- a/packages/appkit/src/connectors/index.ts
+++ b/packages/appkit/src/connectors/index.ts
@@ -1,5 +1,6 @@
 export * from "./files";
 export * from "./genie";
+export * from "./jobs";
 export * from "./lakebase";
 export * from "./lakebase-v1";
 export * from "./sql-warehouse";
--- a/packages/appkit/src/plugins/jobs/index.ts
+++ b/packages/appkit/src/plugins/jobs/index.ts
@@ -1,0 +1,8 @@
+export { jobs } from "./plugin";
+export type {
+  IJobsConfig,
+  JobAPI,
+  JobConfig,
+  JobHandle,
+  JobsExport,
+} from "./types";
--- a/packages/appkit/src/plugins/jobs/plugin.ts
+++ b/packages/appkit/src/plugins/jobs/plugin.ts
@@ -1,26 +1,88 @@
+import { STATUS_CODES } from "node:http";
 import type { jobs as jobsTypes } from "@databricks/sdk-experimental";
-import type { IAppRequest } from "shared";
+import type express from "express";
+import type {
+  IAppRequest,
+  IAppRouter,
+  PluginExecutionSettings,
+  StreamExecutionSettings,
+} from "shared";
+import { toJSONSchema } from "zod";
 import { JobsConnector } from "../../connectors/jobs";
-import { getWorkspaceClient } from "../../context";
-import { InitializationError } from "../../errors";
+import { getCurrentUserId, getWorkspaceClient } from "../../context";
+import { ExecutionError, ValidationError } from "../../errors";
 import { createLogger } from "../../logging/logger";
+import type { ExecutionResult } from "../../plugin";
 import { Plugin, toPlugin } from "../../plugin";
 import type { PluginManifest, ResourceRequirement } from "../../registry";
 import { ResourceType } from "../../registry";
+import {
+  JOBS_READ_DEFAULTS,
+  JOBS_STREAM_DEFAULTS,
+  JOBS_WRITE_DEFAULTS,
+} from "./defaults";
 import manifest from "./manifest.json";
+import { mapParams } from "./params";
 import type {
   IJobsConfig,
   JobAPI,
   JobConfig,
   JobHandle,
+  JobRunStatus,
   JobsExport,
 } from "./types";
 
 const logger = createLogger("jobs");
 
-const DEFAULT_TIMEOUT = 60_000;
 const DEFAULT_WAIT_TIMEOUT = 600_000;
 const DEFAULT_POLL_INTERVAL = 5_000;
+/** Cap on param-key count when a job has no Zod schema. Jobs that need more keys must define a schema. */
+const MAX_UNVALIDATED_PARAM_KEYS = 50;
+
+/** Replace upstream error messages with generic descriptions keyed by HTTP status. */
+function errorResult(status: number): ExecutionResult<never> {
+  return {
+    ok: false,
+    status,
+    message: STATUS_CODES[status] ?? "Request failed",
+  };
+}
+
+function isTerminalRunState(state: string | undefined): boolean {
+  return (
+    state === "TERMINATED" || state === "SKIPPED" || state === "INTERNAL_ERROR"
+  );
+}
+
+/** Exponential backoff (1.5x) with +/- 20% jitter, capped at `max`. */
+function nextPollDelay(
+  current: number,
+  max: number,
+): { delay: number; next: number } {
+  const jitter = 1 + (Math.random() * 0.4 - 0.2);
+  return {
+    delay: Math.min(current * jitter, max),
+    next: Math.min(current * 1.5, max),
+  };
+}
+
+function abortableSleep(ms: number, signal?: AbortSignal): Promise<void> {
+  return new Promise<void>((resolve) => {
+    if (signal?.aborted) {
+      resolve();
+      return;
+    }
+    const timer = setTimeout(resolve, ms);
+    signal?.addEventListener(
+      "abort",
+      () => {
+        clearTimeout(timer);
+        resolve();
+      },
+      { once: true },
+    );
+  });
+}
 
 class JobsPlugin extends Plugin {
   static manifest = manifest as PluginManifest;
@@ -28,6 +90,7 @@ class JobsPlugin extends Plugin {
   protected declare config: IJobsConfig;
   private connector: JobsConnector;
   private jobIds: Record<string, number> = {};
+  private jobConfigs: Record<string, JobConfig> = {};
   private jobKeys: string[] = [];
 
   /**
@@ -56,7 +119,7 @@ class JobsPlugin extends Plugin {
       Object.keys(explicit).length === 0 &&
       Object.keys(discovered).length === 0
     ) {
-      discovered["default"] = {};
+      discovered.default = {};
     }
 
     return { ...discovered, ...explicit };
@@ -91,12 +154,12 @@ class JobsPlugin extends Plugin {
     super(config);
     this.config = config;
     this.connector = new JobsConnector({
-      timeout: config.timeout ?? DEFAULT_TIMEOUT,
       telemetry: config.telemetry,
     });
 
     const jobs = JobsPlugin.discoverJobs(config);
     this.jobKeys = Object.keys(jobs);
+    this.jobConfigs = jobs;
 
     for (const key of this.jobKeys) {
       const envVar =
@@ -105,8 +168,8 @@ class JobsPlugin extends Plugin {
           : `DATABRICKS_JOB_${key.toUpperCase()}`;
       const jobIdStr = process.env[envVar];
       if (jobIdStr) {
-        const parsed = parseInt(jobIdStr, 10);
-        if (!isNaN(parsed)) {
+        const parsed = Number.parseInt(jobIdStr, 10);
+        if (!Number.isNaN(parsed)) {
           this.jobIds[key] = parsed;
         }
       }
@@ -114,25 +177,6 @@ class JobsPlugin extends Plugin {
   }
 
   async setup() {
-    const client = getWorkspaceClient();
-    if (!client) {
-      throw new InitializationError(
-        "Jobs plugin requires a configured workspace client",
-      );
-    }
-
-    if (this.jobKeys.length === 0) {
-      logger.warn(
-        "No jobs configured. Set DATABRICKS_JOB_ID or DATABRICKS_JOB_<NAME> env vars.",
-      );
-    }
-
-    for (const key of this.jobKeys) {
-      if (!this.jobIds[key]) {
-        logger.warn(`Job "${key}" has no valid job ID configured.`);
-      }
-    }
-
     logger.info(
       `Jobs plugin initialized with ${this.jobKeys.length} job(s): ${this.jobKeys.join(", ")}`,
     );
@@ -145,11 +189,62 @@ class JobsPlugin extends Plugin {
   private getJobId(jobKey: string): number {
     const id = this.jobIds[jobKey];
     if (!id) {
+      const envVar =
+        jobKey === "default"
+          ? "DATABRICKS_JOB_ID"
+          : `DATABRICKS_JOB_${jobKey.toUpperCase()}`;
       throw new Error(
-        `Job "${jobKey}" has no configured job ID. Set DATABRICKS_JOB_${jobKey.toUpperCase()} env var.`,
+        `Job "${jobKey}" has no configured job ID. Set ${envVar} env var.`,
       );
     }
     return id;
+  }
+
+  private _readSettings(
+    cacheKey: (string | number | object)[],
+  ): PluginExecutionSettings {
+    return {
+      default: {
+        ...JOBS_READ_DEFAULTS,
+        ...(this.config.timeout != null && { timeout: this.config.timeout }),
+        cache: { ...JOBS_READ_DEFAULTS.cache, cacheKey },
+      },
+    };
+  }
+
+  private _writeSettings(): PluginExecutionSettings {
+    return {
+      default: {
+        ...JOBS_WRITE_DEFAULTS,
+        ...(this.config.timeout != null && { timeout: this.config.timeout }),
+      },
+    };
+  }
+
+  /**
+   * Validates params against the job's Zod schema (if any) and maps them
+   * to SDK request fields based on the task type. Shared by runNow and runAndWait.
+   */
+  private _validateAndMap(
+    jobKey: string,
+    params?: Record<string, unknown>,
+  ): Record<string, unknown> {
+    const jobConfig = this.jobConfigs[jobKey];
+    let validated = params;
+
+    if (jobConfig?.params) {
+      const result = jobConfig.params.safeParse(params ?? {});
+      if (!result.success) {
+        throw new ValidationError(
+          `Parameter validation failed for job "${jobKey}": ${result.error.message}`,
+        );
+      }
+      validated = result.data as Record<string, unknown>;
+    }
+
+    return jobConfig?.taskType && validated
+      ? mapParams(jobConfig.taskType, validated)
+      : (validated ?? {});
   }
 
   /**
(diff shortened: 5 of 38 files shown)


pkosiec:
How it is different from @keugenek PR here: https://github.com/databricks/appkit/pull/221?

If you based on top of Evgenii's work, maybe it's worth to recognize his contribution? What I'm thinking is either:
- have 2 commits on this PR: Evgenii's base one and yours (with all changes, PR review fixes etc.), and temporarily enable the "Rebase & Merge" option to have 2 commits on main
- or, merge Evgenii's PR as is, right before this PR (so we'd need to wait for this PR to be approved first)

WDYT?

atilafassina:
@pkosiec 
> How it is different from @keugenek PR here: #221?
> If you based on top of Evgenii's work, maybe it's worth to recognize his contribution? What I'm thinking is either:
> * have 2 commits on this PR: Evgenii's base one and yours (with all changes, PR review fixes etc.), and temporarily enable the "Rebase & Merge" option to have 2 commits on main
> * or, merge Evgenii's PR as is, right before this PR (so we'd need to wait for this PR to be approved first)
> 
> 
> WDYT?

I got stuck trying to update his PR because it was opened from a fork, it's noted in the PR description:

> Reopened from https://github.com/databricks/appkit/pull/223 — moved head branch from fork to upstream to fix CI secrets access.

So, I couldn't get a clean CI to merge it, and if I merged to `main`, it would trigger a release.
You'll notice that in the commit history the very first commit is his, that's his work that I iterated on top of.


It's also marked as a comment on the original PR

https://github.com/databricks/appkit/pull/223#issuecomment-4226607528

pkosiec:
> So, I couldn't get a clean CI to merge it, and if I merged to main, it would trigger a release.

It wouldn't trigger a release currently (finalizing releases is manual) so there shouldn't be a problem with that 👍  but I'd wait until your PR (on top of this one) is ready to merge, and I'd merge both one after another. What do you think?

packages/appkit/src/plugins/jobs/plugin.ts
```diff
@@ -0,0 +1,739 @@
...
+        params: config?.params ? toJSONSchema(config.params) : null,
+        taskType: config?.taskType ?? null,
+      };
+    }
+    return { jobs };
+  }
+}
+
+/**
+ * @internal
```
MarioCadenas, line 727: this shouldn't be internal no?

packages/appkit/src/plugins/jobs/plugin.ts
```diff
@@ -0,0 +1,739 @@
...
+ */
+export const jobs = toPlugin(JobsPlugin);
+
+export { JobsPlugin };
```
MarioCadenas, line 731: why do we need to export this? tests?

packages/appkit/src/plugins/jobs/plugin.ts
```diff
@@ -0,0 +1,739 @@
...
+  }
+
+  injectRoutes(router: IAppRouter) {
+    // POST /:jobKey/run
+    this.route(router, {
+      name: "run",
+      method: "post",
+      path: "/:jobKey/run",
+      handler: async (req: express.Request, res: express.Response) => {
+        const { jobKey } = this._resolveJob(req, res);
```
MarioCadenas: can we maybe simplify this endpoint a little bit? it seems too long no?

packages/appkit/src/plugins/jobs/plugin.ts
```diff
@@ -0,0 +1,739 @@
...
+              },
+              signal,
+            ),
+          self._writeSettings(),
+          userKey,
+        );
+        return result.ok ? result : errorResult(result.status);
+      },
+
+      async *runAndWait(
```
MarioCadenas, line 259: can `runAndWait` be simplified? maybe we can extract some parts?

packages/appkit/src/plugins/jobs/plugin.ts
```diff
@@ -0,0 +1,739 @@
...
+    }
+  }
+
+  async setup() {
+    const client = getWorkspaceClient();
+    if (!client) {
+      throw new InitializationError(
+        "Jobs plugin requires a configured workspace client",
+      );
+    }
```
MarioCadenas: do we need this check? I don't think its needed, at least no plugin is doing this check?

packages/appkit/package.json
```diff
@@ -77,7 +77,8 @@
     "semver": "7.7.3",
     "shared": "workspace:*",
     "vite": "npm:rolldown-vite@7.1.14",
-    "ws": "8.18.3"
+    "ws": "8.18.3",
+    "zod": "^4.3.6"
```
MarioCadenas: pin the version to `4.3.6` please 😄

.claude/skills/install-appkit-tarball.md
pkosiec, line 1: Could you rebase the PR? The updated skill is on main: https://github.com/databricks/appkit/blob/main/.claude/skills/install-appkit-artifact.md

pkosiec, line 1: Here's agentic review:


## Code Review: Jobs Plugin for Databricks Lakeflow Jobs

**Scope:** `ekniazev/jobs-plugin-core` branch, 2 commits, ~39 files changed
**Intent:** Add a new resource-scoped Jobs plugin for Databricks Lakeflow Jobs with connector, plugin, HTTP routes, SSE streaming, param validation, OBO support, frontend page, docs, and tests. Also fixes StreamManager to abort generators on client disconnect.

### P1 -- High

| # | File | Issue | Confidence |
|---|------|-------|------------|
| 1 | `docs/docs/plugins/jobs.md:41` | **Docs/code mismatch: per-job field name**. Docs table says `timeout` for per-job config, but the TypeScript field in `types.ts:69` is `waitTimeout`. Users following the docs will set the wrong field. | 0.95 |
| 2 | `jobs.route.tsx:95-100` | **SSE stream parsing doesn't handle split chunks**. `decoder.decode(value, { stream: true })` can split a `data:` line across two chunks. The code splits on `\n` and checks `line.startsWith("data: ")`, so a line split mid-chunk would be silently dropped or corrupted. Should buffer incomplete lines between reads. | 0.85 |

### P2 -- Moderate

| # | File | Issue | Confidence |
|---|------|-------|------------|
| 3 | `jobs.route.tsx:255-256` | **React key collision on stream log**. `key={line}` uses the line content as key. Duplicate SSE messages (heartbeats, repeated status like `"RUNNING"`) will produce duplicate keys, causing React reconciliation issues and potential UI glitches. Use index-based key or prepend a counter. | 0.90 |
| 4 | `plugin.ts:229-278` | **Duplicated param validation block**. `runNow` (lines 229-245) and `runAndWait` (lines 268-283) contain identical validation + mapping logic. If validation rules change, both must be updated. Extract a shared `_validateAndMapParams(jobKey, params)` method. | 0.85 |
| 5 | `connectors/jobs/client.ts:176` | **Semantic mismatch: `ExecutionError.statementFailed`**. Used for generic Jobs API errors, but the name implies SQL statement failures. Other connectors (sql-warehouse) use it for actual SQL errors. While the error chain works technically, it's misleading for debugging/logging. | 0.75 |

### P3 -- Low

| # | File | Issue | Confidence |
|---|------|-------|------------|
| 6 | `plugin.ts:717` | **Unnecessary wrapper function**. `((jobKey: string) => resolveJob(jobKey)) as JobsExport` can be simplified to `resolveJob as JobsExport`. | 0.90 |
| 7 | `jobs.route.tsx:23-24` | **TERMINATED mapped to green unconditionally**. `stateColor("TERMINATED")` returns green, but TERMINATED only means the run ended -- it could have failed. The result_state column handles this separately, but the green lifecycle indicator is misleading when paired with a red result. | 0.70 |
| 8 | `package.json` (appkit) | **Zod caret range `^4.3.6`**. For a published SDK, pinned versions are safer to avoid unexpected breaks from minor Zod releases. The lockfile shows two Zod versions (4.1.13 + 4.3.6) coexisting. | 0.65 |

### Coverage

- **Tests:** Comprehensive test suite (1712 lines) covering discovery, resource requirements, exports, parameter validation, interceptors, polling, error handling, OBO, route handlers, XSS sanitization.
- **StreamManager fix:** Includes a targeted test for the new abort-on-disconnect behavior.
- **Testing gaps:** No integration test for the full SSE streaming path (triggerRun -> poll -> client disconnect -> abort). Frontend component has no tests (acceptable for dev-playground).

---

**Verdict: Ready with fixes.** The docs/code mismatch (#1) and SSE parsing (#2) should be addressed before merge. Items #3-5 are recommended. Items #6-8 are at your discretion.

MarioCadenas (commented):
I added some comments, but I'm actually wondering if this should just be a "jobs" plugin or if we should extend this and make it support both jobs and pipelines tbh 😅

MarioCadenas:
a few more things I noticed going through the core package:

---

**1. `listRuns` cache key probably never hits cache**

in `plugin.ts`, `listRuns` uses `options ?? {}` in the cache key:

```ts
self._readSettings(["jobs:listRuns", jobKey, options ?? {}])
```

when `options` is `undefined` (which is common), this creates a **new `{}` every time** — so if the cache compares by reference, every call is a miss and the 60s TTL does nothing. the other read methods get this right with primitives (`["jobs:getRun", jobKey, runId]`). can we normalize this to something like `["jobs:listRuns", jobKey, options?.limit ?? "default"]`?

---

**2. `mapParams` silently coerces non-primitives to garbage**

in `params.ts`, the notebook/python_wheel/sql cases do `String(v)` on every value:

```ts
notebook_params: Object.fromEntries(
  Object.entries(params).map(([k, v]) => [k, String(v)]),
),
```

if someone passes an object or array through (e.g. with a loose zod schema like `z.record(z.unknown())`), `String(v)` produces `"[object Object]"` or `"1,2,3"` — the job would run with garbage params and no error. can we at least add a guard that rejects non-primitive values here, or coerce more defensively?

---

**3. polling in `runAndWait` — no backoff or jitter**

I know I already asked if we can simplify `runAndWait`, but specifically the poll interval is fixed at 5s forever. with the default 10min timeout that's 120 API calls per waiter, and if you have multiple SSE clients polling the same run they all hit the Jobs API at the same cadence with no jitter. should we add some basic exponential backoff (e.g. 2s → 5s → 10s → cap at 15s) and a small random jitter so concurrent pollers don't align?

packages/appkit/src/plugins/jobs/index.ts
pkosiec, line 1: The reason I asked for an e2e demo is to see how the whole flow looks like. Checked and IMO we should adjust the description - shorten it and do not mention envs:

Image

What do you think?

BTW, please try to deploy such app to see if the DAB config is correct etc. 👍

pkosiec (commented):
Just a few comments 👍 Nice work!

packages/appkit/src/plugins/jobs/plugin.ts
atilafassina: correct. I should probably add an `@internal` here.
it makes the tests a bit easier to write by reaching the statics directly

I can remove it though, just felt the tradeoff was worth it

packages/appkit/src/plugins/jobs/plugin.ts
atilafassina: the `toPlugin` export is internal in every plugin.
We talked on slack and we'll keep it like this for consistency with others while we reevaluate how to export the plugins.

marking as resolved for now.

packages/appkit/src/plugins/jobs/plugin.ts
atilafassina: refactored the client usage to match other plugins - the check is gone now

packages/appkit/src/plugins/jobs/index.ts
atilafassina: i did a few deploys and fixed all the quirks

atilafassina:
> I added some comments, but I'm actually wondering if this should just be a "jobs" plugin or if we should extend this and make it support both jobs and pipelines tbh 😅

Adding this here just for historical reference.
We synced on this and decided to ship them as separate plugins initially and observe usage first.

packages/appkit/src/connectors/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./files";
 export * from "./genie";
+export * from "./jobs";
```
pkosiec, line 3: PTAL if that makes sense 🙏 

## Code Review Results

**Scope:** merge-base `422afb3`..working tree (42 files, ~5200 lines changed, ~1900 non-test/non-generated)
**Intent:** Add a Lakeflow Jobs plugin following the resource-scoped pattern with env-var discovery, HTTP endpoints, SSE streaming, Zod parameter validation, OBO support, and a StreamManager abort-on-disconnect fix.
**Mode:** interactive

**Reviewers:** correctness, testing, maintainability, project-standards, agent-native-reviewer, learnings-researcher, security, api-contract, reliability, adversarial, kieran-typescript, previous-comments
- security -- new HTTP endpoints accept user-provided jobKey, params, runId; OBO handling
- api-contract -- 5 new REST endpoints, new exported TypeScript types
- reliability -- polling loops, abort signals, timeouts, StreamManager abort fix
- adversarial -- 1900+ non-test changed lines, external API integration, data mutations
- kieran-typescript -- substantial TypeScript with async generators, closures, generics
- previous-comments -- PR has 20 comments and 7 reviews

### P1 -- High

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 1 | `plugins/jobs/tests/plugin.test.ts` | No test coverage for `POST /:jobKey/run?stream=true` — entire SSE streaming code path untested | testing | 0.95 | `safe_auto -> review-fixer` |
| 2 | `plugins/jobs/plugin.ts:653` | GET `/runs/:runId` returns raw `result.data` while POST `/run`, GET `/runs`, GET `/status` all use envelope wrappers — inconsistent API surface | api-contract | 0.85 | `gated_auto -> human` |

### P2 -- Moderate

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 3 | `plugins/jobs/plugin.ts:191` | Job ID 0 treated as missing: `if (!id)` falsy check rejects `id=0` — should use `id === undefined` | correctness | 0.95 | `safe_auto -> review-fixer` |
| 4 | `plugins/jobs/plugin.ts:520` | Double Zod validation: `_parseRunParams` validates then returns raw params, `_validateAndMap` re-validates — Zod transforms run twice, non-idempotent transforms produce wrong output | adversarial, kieran-typescript | 0.88 | `manual -> downstream-resolver` |
| 5 | `docs/docs/plugins/jobs.md:170` | GET `/runs/:runId` response shape undocumented — other endpoints document returns explicitly | api-contract | 0.90 | `safe_auto -> review-fixer` |
| 6 | `plugins/jobs/tests/plugin.test.ts` | Missing tests for `SKIPPED` and `INTERNAL_ERROR` terminal states in `runAndWait` | testing | 0.85 | `safe_auto -> review-fixer` |
| 7 | `plugins/jobs/tests/plugin.test.ts` | Missing test for `runAndWait` polling timeout branch (line 337–341) | testing | 0.80 | `safe_auto -> review-fixer` |
| 8 | `plugins/jobs/plugin.ts:360` | Undefined `life_cycle_state` causes polling to continue silently until timeout — no log/warning for unrecognized states | adversarial, kieran-typescript | 0.82 | `advisory -> human` |
| 9 | `template/appkit.plugins.json:93` | Template resource description too verbose per reviewer feedback — pkosiec requested shortening and removing env var details | previous-comments | 0.75 | `safe_auto -> review-fixer` |

### P3 -- Low

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 10 | `plugins/jobs/plugin.ts:110` | Case sensitivity mismatch: env discovery lowercases keys but HTTP routing is case-sensitive — `/ETL/run` returns 404 for job discovered as `etl` | adversarial | 0.75 | `advisory -> human` |
| 11 | `plugins/jobs/plugin.ts:754` | `return resolveJob as JobsExport` uses type cast instead of typed assignment — weakens type checking | kieran-typescript | 0.73 | `advisory -> human` |
| 12 | `plugins/jobs/plugin.ts:639` | `Number.parseInt` silently truncates `"42abc"` to `42` — no strict numeric validation on runId | adversarial | 0.68 | `advisory -> human` |
| 13 | `plugins/jobs/plugin.ts:332` | Dual timeout: polling loop timeout (waitTimeout) and execute interceptor timeout (JOBS_STREAM_DEFAULTS) are independent, could confuse debugging | reliability | 0.68 | `advisory -> human` |
| 14 | `jobs.route.tsx` / `JobsPage.tsx` | Duplicated frontend code (~287 lines each) between dev-playground route and template page | maintainability | 0.82 | `advisory -> human` |

### Pre-existing Issues

| # | File | Issue | Reviewer |
|---|------|-------|----------|
| 1 | `plugin/interceptors/retry.ts:68` | `sleep()` uses plain setTimeout without abort signal awareness — retry backoff blocks abort propagation | reliability |

### Agent-Native Gaps

- **No `getRunOutput` HTTP route**: `JobAPI.getRunOutput()` exists programmatically but has no HTTP endpoint — agents using HTTP can't fetch run output
- **Generic 400 error messages**: Parameter validation failures return `"Invalid job parameters"` without Zod error details — agents can't self-correct based on the error
- **No job config discovery endpoint**: Available job keys, parameter schemas, and task types are only exposed to the frontend via `clientConfig()` — no HTTP route for agents to discover what jobs are available

### Coverage

- Suppressed: 3 findings below 0.60 confidence
- Residual risks: No rate limiting on job endpoints; TOCTOU window between `verifyRunScope` and actual operation; prototype pollution not explicitly guarded in unvalidated params
- Testing gaps: No integration test for SSE streaming with client disconnect; no test for Zod transforms with side effects; no multi-client SSE reconnection test
- Failed reviewers: 0 of 12

---

> **Verdict:** Ready with fixes
>
> **Reasoning:** Well-designed plugin that follows established patterns. The security surface is clean (input sanitization, resource scoping, error message redaction). The main concerns are: (1) missing streaming test coverage for a critical code path, (2) a minor API consistency issue with unwrapped GET response, and (3) the double Zod validation pattern that could break with transform schemas. No critical bugs or security vulnerabilities found.
>
> **Fix order:** #1 streaming test gap → #3 job ID falsy check → #9 template description → #5/#6/#7 remaining test gaps → #2 response envelope (if desired, this is a design decision)

docs/docs/plugins/index.md
```diff
@@ -13,14 +13,15 @@ For complete API documentation, see the [`Plugin`](../api/appkit/Class.Plugin.md
...
-import { createApp, server, analytics, genie, files } from "@databricks/appkit";
+import { createApp, server, analytics, genie, files, jobs } from "@databricks/appkit";
 
 const AppKit = await createApp({
   plugins: [
     server({ port: 8000 }),
     analytics(),
     genie(),
     files(),
+    jobs(),
```
pkosiec, line 24: I did `apps init --template (from your PR` and did immediate deploy.
Firstly, it failed because it couldn't install npm packages 🤔 not sure why? I deleted the package-lock and the deploy succeeded, but I cannot use my app:

```

[appkit:plugin] Plugin execution failed {   error: ApiError: Provided OAuth token does not have required scopes: jobs [ReqId: 716c9e46-6921-4438-b856-2366ffa51ae2]       at parseErrorFromResponse (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apierr.js:100:12)       at fn (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/api-client.js:193:69)       at process.processTicksAndRejections (node:internal/process/task_queues:105:5)       at async retry (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/retries/retries.js:95:22)       at async ApiClient.request (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/api-client.js:168:26)       at async JobsService._listRuns (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apis/jobs/api.js:606:17)       at async JobsService.listRuns (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apis/jobs/api.js:620:30)       at async file:///app/python/source_code/node_modules/@databricks/appkit/dist/connectors/jobs/client.js:60:21       at async telemetry.startActiveSpan.name (file:///app/python/source_code/node_modules/@databricks/appkit/dist/connectors/jobs/client.js:83:20)       at async file:///app/python/source_code/node_modules/@databricks/appkit/dist/plugin/interceptors/telemetry.js:29:20 {     errorCode: 403,     statusCode: 403,     response: {       error_code: 403,       message: 'Provided OAuth token does not have required scopes: jobs [ReqId: 716c9e46-6921-4438-b856-2366ffa51ae2]'     },     details: [],     errorInfoType: 'type.googleapis.com/google.rpc.ErrorInfo'   },   plugin: 'jobs' }
--
Apr 23, 2026, 03:28:09 PM GMT+2 | APP | ...77d18cf | [appkit:plugin] Plugin execution failed {   error: ApiError: Provided OAuth token does not have required scopes: jobs [ReqId: 05fc24d4-686e-462f-8469-adc1c2b0da93]       at parseErrorFromResponse (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apierr.js:100:12)       at fn (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/api-client.js:193:69)       at process.processTicksAndRejections (node:internal/process/task_queues:105:5)       at async retry (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/retries/retries.js:95:22)       at async ApiClient.request (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/api-client.js:168:26)       at async JobsService._runNow (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apis/jobs/api.js:822:17)       at async JobsService.runNow (/app/python/source_code/node_modules/@databricks/appkit/node_modules/@databricks/sdk-experimental/dist/apis/jobs/api.js:836:32)       at async telemetry.startActiveSpan.name (file:///app/python/source_code/node_modules/@databricks/appkit/dist/connectors/jobs/client.js:83:20)       at async file:///app/python/source_code/node_modules/@databricks/appkit/dist/plugin/interceptors/telemetry.js:29:20       at async TimeoutInterceptor.intercept (file:///app/python/source_code/node_modules/@databricks/appkit/dist/plugin/interceptors/timeout.js:17:11) {     errorCode: 403,     statusCode: 403,     response: {       error_code: 403,       message: 'Provided OAuth token does not have required scopes: jobs [ReqId: 05fc24d4-686e-462f-8469-adc1c2b0da93]'     },     details: [],     errorInfoType: 'type.googleapis.com/google.rpc.ErrorInfo'   },   plugin: 'jobs' }


```

Image

Do you know why? Can you reproduce it? Here's my job: https://dogfood.staging.databricks.com/apps-v2/app/pkjobs2/overview?o=6051921418418893

Image

MarioCadenas: I think the databricks.yaml file in the template is missing the user scopes no?

atilafassina: > I think the databricks.yaml file in the template is missing the user scopes no?

this plugin uses SP, resources are auto-discovered based on plugin manifest.
Right now my culprit is that this is an older version of the code we have at the moment, I may have fixed this before seeing Pawel's feedback, hence the 🐰 🕳️

pkosiec (approved):
Image

Works now! Didn't notice CI jobs weren't triggered for the latest run and probably tested an outdated version.

LGTM, as the OBO approach wouldn't be possible currently anyway because of lack of the jobs scope in Apps (unless I missed it)

PR #267, opened 2026-04-13
refactor: move generated types into appkit-types/ directory

pkosiec:
Generated type declaration files (analytics query types, serving endpoint types) used to live as separate top-level files in `client/src/`. This PR moves them into a dedicated `src/appkit-types/` directory — one file per plugin. TypeScript auto-discovers them via `"include": ["src"]` in tsconfig, so no index or barrel file is needed.

## Screenshots

image (3)
image (4)

## Changes

- Analytics types: `src/appKitTypes.d.ts` → `src/appkit-types/analytics.d.ts`
- Serving types: `src/appKitServingTypes.d.ts` → `src/appkit-types/serving.d.ts`
- Both generators now `mkdir -p` before writing, so the directory is created on first run
- Directory name and filenames extracted as shared constants (`TYPES_DIR`, `ANALYTICS_TYPES_FILE`, `SERVING_TYPES_FILE`)
- Updated Vite plugin defaults, CLI defaults, docs, tests, and `.gitignore`


packages/appkit/src/type-generator/vite-plugin.ts
```diff
@@ -71,14 +75,17 @@ export function appKitTypesPlugin(options?: AppKitTypesPluginOptions): Plugin {
...
       root = config.root;
-      outFile = path.resolve(root, options?.outFile ?? "src/appKitTypes.d.ts");
+      outFile = path.resolve(
+        root,
+        options?.outFile ?? `src/${TYPES_DIR}/${ANALYTICS_TYPES_FILE}`,
+      );
       watchFolders = options?.watchFolders ?? [
         path.join(process.cwd(), "config", "queries"),
       ];
     },
```
MarioCadenas, line 85: Nit: this previously fired `generate()` without `await`, so Vite could continue the build pipeline while type generation ran in the background. Now that it's awaited, type generation is serialized on the `buildStart` critical path — build startup will block until it completes.

Was this intentional? If so, totally fine (deterministic types before compilation is a good trade-off). Just flagging in case the fire-and-forget was deliberate for faster dev startup.

MarioCadenas (approved):

packages/appkit/src/type-generator/vite-plugin.ts
pkosiec: you're right - the original idea was to generate the index file so we needed to await for it.. but as the file is no longer generated, we can avoid blocking 👍 I'll fix it

PR #268, opened 2026-04-13
fix: use end-user ID in OBO analytics cache key

MarioCadenas:
## Summary

- **Fix cross-user data leak in OBO analytics caching.** `getCurrentUserId()` was called outside `runInUserContext`, so the cache key for `.obo.sql` queries used the service principal's ID instead of the requesting user's. Two different users making the same OBO query with the same parameters shared a single cache entry.
- **Read user ID from request header directly** via `resolveUserId(req)` instead of relying on `AsyncLocalStorage`, which hadn't entered user context yet at the point where the cache key was constructed.
- **Add regression test** that verifies two users making identical OBO queries get separate cache entries and their own query results.

## Test plan

- [x] New test: "OBO cache key must use the end user's ID, not the service principal's" — confirms `executeStatement` is called once per user, not shared via cache
- [x] All existing analytics tests pass (unit + integration)
- [x] Full test suite passes (82 files, 1535 tests)
- [x] TypeScript type check passes
- [x] Biome lint/format passes

### DIFF
--- a/packages/appkit/src/plugins/analytics/analytics.ts
+++ b/packages/appkit/src/plugins/analytics/analytics.ts
@@ -7,11 +7,7 @@ import type {
   StreamExecutionSettings,
 } from "shared";
 import { SQLWarehouseConnector } from "../../connectors";
-import {
-  getCurrentUserId,
-  getWarehouseId,
-  getWorkspaceClient,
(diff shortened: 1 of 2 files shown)


ditadi (approved):
nice

PR #269, opened 2026-04-13
fix(appkit): add jitter to RetryInterceptor exponential backoff

MarioCadenas:
## Summary
The `RetryInterceptor` used pure exponential backoff (`initialDelay * 2^attempt`), causing all concurrent retries after a simultaneous failure to fire at identical intervals — a thundering herd problem. This PR adds full jitter by default: each retry delay is randomized between 50% and 100% of the computed backoff value, spreading retries over time.

## Changes
- Added `jitter?: boolean` field to `RetryConfig` (defaults to `true`)
- Updated `RetryInterceptor.calculateDelay()` to apply random jitter: `capped * (0.5 + Math.random() * 0.5)`
- Existing tests updated to set `jitter: false` where exact timing is asserted
- New tests: jitter enabled by default, boundary behavior at `Math.random() = 0` (50%) and `Math.random() = 1` (100%)

## Test plan
- [x] Existing tests pass (backoff + maxDelay tests now use `jitter: false`)
- [x] New tests cover jitter default-on, boundary values, and opt-out
- [x] Manual verification: `jitter: false` produces deterministic delays identical to prior behavior

### DIFF
--- a/packages/appkit/src/plugin/interceptors/retry.ts
+++ b/packages/appkit/src/plugin/interceptors/retry.ts
@@ -9,11 +9,13 @@ export class RetryInterceptor implements ExecutionInterceptor {
   private attempts: number;
   private initialDelay: number;
   private maxDelay: number;
+  private jitter: boolean;
 
   constructor(config: RetryConfig) {
     this.attempts = config.attempts ?? 3;
     this.initialDelay = config.initialDelay ?? 1000;
     this.maxDelay = config.maxDelay ?? 30000;
+    this.jitter = config.jitter ?? true;
   }
 
   async intercept<T>(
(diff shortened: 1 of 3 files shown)


packages/appkit/src/plugin/interceptors/retry.ts
```diff
@@ -59,11 +61,13 @@ export class RetryInterceptor implements ExecutionInterceptor {
...
-    // exponential backoff
     const delay = this.initialDelay * 2 ** (attempt - 1);
+    const capped = Math.min(delay, this.maxDelay);
 
-    // max delay cap
-    return Math.min(delay, this.maxDelay);
+    if (!this.jitter) return capped;
+
+    // Full jitter: random value between 50% and 100% of the calculated delay
+    return capped * (0.5 + Math.random() * 0.5);
```
ditadi: For full jitter should be `capped * Math.random()` only

With the 0.5, you bound to half of the window.

packages/appkit/src/plugin/interceptors/retry.ts
```diff
@@ -9,11 +9,13 @@ export class RetryInterceptor implements ExecutionInterceptor {
   private attempts: number;
   private initialDelay: number;
   private maxDelay: number;
+  private jitter: boolean;
```
ditadi: Do we need this flag? jitter is better in our cases for any production scenario

ditadi (approved):
Sounds good, I just added two points.

MarioCadenas:
> Sounds good, I just added two points.

addressed, thank you!

PR #270, opened 2026-04-15
fix: convert BigInt to strings in Arrow IPC row.toJSON()

jamesbroadhead:
## Summary

- `row.toJSON()` from apache-arrow preserves `BigInt` for Int64 columns. When this data is later serialized with `JSON.stringify()` for SSE streaming, it throws `TypeError: Do not know how to serialize a BigInt`.
- Convert BigInt values to strings in `_transformArrowAttachment` during row transformation, before the data reaches JSON serialization.
- This ensures Arrow IPC deserialization works correctly for serverless warehouses that return inline results as Arrow IPC in `result.attachment`.

## Test plan

- [ ] Verify that queries returning Int64 columns via Arrow IPC attachment no longer throw `TypeError: Do not know how to serialize a BigInt` during SSE streaming
- [ ] Verify that non-BigInt columns are unaffected by the change
- [ ] Verify that the string representation of BigInt values is correct and usable downstream

### DIFF
--- a/packages/appkit/src/connectors/sql-warehouse/client.ts
+++ b/packages/appkit/src/connectors/sql-warehouse/client.ts
@@ -3,6 +3,7 @@ import {
   type sql,
   type WorkspaceClient,
 } from "@databricks/sdk-experimental";
+import { tableFromIPC } from "apache-arrow";
 import type { TelemetryOptions } from "shared";
 import {
   AppKitError,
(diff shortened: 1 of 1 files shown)


jamesbroadhead:
Superseded by #315 — recreated against the upstream repo so CI workflows can run (fork PRs don't get secrets, hence the broken Lint/Unit/Playground/Docs checks here).

PR #271, opened 2026-04-15
fix: handle ARROW_STREAM attachment in type generator

jamesbroadhead:
## Summary
- When serverless warehouses return ARROW_STREAM format, DESCRIBE QUERY results come as an inline base64 Arrow IPC `attachment` rather than `data_array`. This caused `convertToQueryType` to generate empty types `{}`.
- Added a fallback in `convertToQueryType` that decodes the Arrow IPC attachment schema to extract column names and types when `data_array` is empty.
- Extended `DatabricksStatementExecutionResponse` type with the optional `attachment` field.

## Test plan
- [ ] Verify type generation against a serverless warehouse that returns ARROW_STREAM format
- [ ] Verify existing JSON_ARRAY warehouses still produce correct types (no regression)
- [ ] Run `pnpm typecheck` and `pnpm test` to confirm no build or test regressions

### DIFF
--- a/packages/appkit/src/type-generator/query-registry.ts
+++ b/packages/appkit/src/type-generator/query-registry.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 import { WorkspaceClient } from "@databricks/sdk-experimental";
+import { tableFromIPC } from "apache-arrow";
 import pc from "picocolors";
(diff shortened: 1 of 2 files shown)


jamesbroadhead:
Superseded by #316 — recreated against the upstream repo so CI workflows can run (fork PRs don't get secrets, hence the broken Lint/Unit/Playground/Docs checks here).

PR #276, opened 2026-04-15
fix(appkit): check isRetryable before retrying in interceptor

MarioCadenas:
## Summary

The `RetryInterceptor` retried all errors indiscriminately — including `AuthenticationError` (401), `ValidationError` (400), and Databricks SDK 4xx errors — wasting warehouse compute and delaying error responses. This PR adds an `isRetryableError` check that respects the `AppKitError.isRetryable` property and duck-types the Databricks SDK `ApiError.isRetryable()` method, with a status-code fallback (5xx/429 retry, 4xx don't).

## Changes

- Added `isRetryableError()` function to `retry.ts` that checks:
  1. `AppKitError.isRetryable` boolean property (covers all AppKit error subclasses)
  2. SDK `ApiError.isRetryable()` method via duck typing (no SDK import needed)
  3. Status-code heuristic fallback: 5xx and 429 → retryable, other 4xx → not retryable
  4. Unknown errors default to retryable (backward compatibility)
- Inserted the check in the retry catch block, immediately after the abort-signal check
- Added 9 new test cases covering all branches of `isRetryableError`

## Test plan

- [x] All 19 retry tests pass (10 existing + 9 new)
- [x] All existing tests pass (5 pre-existing vector-search failures unrelated to this change)
- [x] Build succeeds
- [x] Biome lint/format passes
- [ ] Manual verification: throw an `AuthenticationError` inside a plugin's `execute()` — it should propagate immediately without retries

### DIFF
--- a/packages/appkit/src/plugin/interceptors/retry.ts
+++ b/packages/appkit/src/plugin/interceptors/retry.ts
@@ -1,10 +1,38 @@
 import type { RetryConfig } from "shared";
+import { AppKitError } from "../../errors/base";
 import { createLogger } from "../../logging/logger";
 import type { ExecutionInterceptor, InterceptorContext } from "./types";
 
 const logger = createLogger("interceptors:retry");
 
-// interceptor to handle retry logic
+/**
+ * Determines whether an error is safe to retry.
+ *
+ * Priority:
+ *  1. AppKitError — reads the `isRetryable` boolean property.
+ *  2. Databricks SDK ApiError (duck-typed) — calls `isRetryable()` method,
+ *     or falls back to status-code heuristic (5xx / 429 → retryable).
+ *  3. Unknown errors — treated as retryable to preserve backward compatibility.
+ */
+function isRetryableError(error: unknown): boolean {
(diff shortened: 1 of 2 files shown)


packages/appkit/src/plugin/interceptors/retry.ts
```diff
@@ -1,10 +1,38 @@
...
+ *
+ * Priority:
+ *  1. AppKitError — reads the `isRetryable` boolean property.
+ *  2. Databricks SDK ApiError (duck-typed) — calls `isRetryable()` method,
+ *     or falls back to status-code heuristic (5xx / 429 → retryable).
+ *  3. Unknown errors — treated as retryable to preserve backward compatibility.
+ */
+function isRetryableError(error: unknown): boolean {
+  if (error instanceof AppKitError) {
+    return error.isRetryable;
```
pkosiec, line 19: Should we make some errors retryable already?

EDIT: Ah okay, some errors are already retryable, nvm 👍

pkosiec (approved):

PR #279, opened 2026-04-16
fix: add AST extraction to serving type generator and move types to shared/

pkosiec:
## Summary

Move generated `appkit-types/` from `client/src/` to `shared/` so both client and server can consume the generated types. Add AST-based endpoint discovery, server-side type-checking, and fix the serving plugin to work correctly with generated types.

### What changed

- **Generated types → `shared/`**: Output moved from `client/src/appkit-types/` to `shared/appkit-types/`. Tsconfigs updated. Old files cleaned up automatically on first regeneration.
- **AST endpoint discovery**: The serving type generator now parses `server/index.ts` to extract `serving({ endpoints: {...} })` config — no manual config needed in vite.config.ts.
- **Server type-checking in template**: `build:server` previously only ran tsdown (transpile-only, no type-checking), while the client already ran `tsc -b` before vite build. Added `noEmit: true` to `tsconfig.server.json` so `tsc -b` can type-check without emitting files — tsdown handles all JS output and only reads the tsconfig for path aliases and module resolution, ignoring emit options. Both sides are now symmetric: type-check (`tsc -b`) → bundle (tsdown/vite).
- **`ServingFactory` adapts to endpoint count**: Single endpoint → alias optional. Multiple endpoints → alias required for disambiguation.
- **Unnamed serving mode registers both URL patterns**: `/api/serving/invoke` (legacy) and `/api/serving/default/invoke` (type-generated clients) both work.
- **Auto-migration for existing apps**: When type generation runs on an existing app with old-style configs, it automatically patches project files (see migration details below).

### Auto-migration for existing apps

Existing apps scaffolded from older templates have config files that still reference the old `client/src/appkit-types/` location. The type generator now automatically patches these on the next run:

| File | What changes | Detection |
|------|-------------|-----------|
| `tsconfig.client.json` | Adds `"shared/appkit-types"` to `include` | Checks if already present |
| `tsconfig.server.json` | Removes `outDir`/`declaration`/`declarationMap`/`sourceMap`, adds `noEmit: true` | Checks if `outDir` exists |
| `package.json` | Updates `build:server` and `typecheck` scripts to use `tsc -b` | Exact-match on old script values |

**Characteristics:**
- **Idempotent** — each sub-migration checks current file state, skips if already migrated
- **Runs once per session** — dedup flag prevents double execution when both analytics and serving generators fire
- **Best-effort** — missing files or parse failures are silently skipped
- **Opt-out** — set `"appkit": { "autoMigrate": false }` in `package.json` to disable

**Note:** JSONC comments in tsconfig files (`/* Emit */`, `// comment`) are stripped during migration since `JSON.parse` requires strict JSON. This only affects the one-time migration write-back.

## Demo

Rough, but shows that everything works as it should. I also deployed those apps - no issues.

https://github.com/user-attachments/assets/e5e9f951-bc1e-4eb0-8a02-7667f639da47

(btw the second schema should be unresolved as that endpoint doesn't expose OpenAPI)

Type generation migration:

https://github.com/user-attachments/assets/de423a9f-6387-4a8d-8a7e-8011d435801e


packages/appkit/src/type-generator/index.ts
```diff
@@ -87,6 +93,10 @@ export async function generateFromEntryPoint(options: {
   await fs.mkdir(path.dirname(outFile), { recursive: true });
   await fs.writeFile(outFile, typeDeclarations, "utf-8");
 
+  // One-time migration: remove old generated file and patch project configs
+  await removeOldGeneratedTypes(projectRoot, "analytics.d.ts");
```
MarioCadenas: `analytics.d.ts` is the old file? I thought the old one was appkitTypes.d.ts no?

packages/appkit/src/type-generator/vite-plugin.ts
```diff
@@ -74,10 +73,10 @@ export function appKitTypesPlugin(options?: AppKitTypesPluginOptions): Plugin {
     },
 
     configResolved(config) {
-      root = config.root;
+      const projectRoot = path.resolve(config.root, "..");
```
MarioCadenas, line 76: this looks a bit fishy, why do we have to go up 1 dir? in theory it already runs from the root no?

pkosiec: `config.root` in Vite points to where index.html lives, which is the `client/` directory. The actual project root is one level up. This matters during pnpm build where `process.cwd()` can also be client/, so we can't rely on `process.cwd()`. Going up one level from `config.root` gives us the real project root where `tsconfig.client.json` and `tsconfig.server.json`, `shared/` etc. all live.

packages/appkit/src/type-generator/index.ts
pkosiec: right, the analytics.d.ts in client was an intermediary type but that was never released. I'll fix it 👍

MarioCadenas (approved):

PR #288, opened 2026-04-20
feat: add execution context attributes to telemetry spans

pkosiec:
## Summary

- Adds `execution.context`, `caller.id`, and `execution.obo_dev_fallback` span attributes to the telemetry interceptor for OBO observability
- Fixes orphaned `plugin.execute` spans by preserving OTel context across the async generator boundary in `executeStream()`
- Adds `db.user` attribute to lakebase query spans
- Documents telemetry span attributes in execution-context.md and custom-plugins.md

### Telemetry span attributes (`plugin.execute` span)

| Attribute | Type | Description |
|---|---|---|
| `execution.context` | `"user"` \| `"service"` | Whether the operation runs as OBO or service principal |
| `caller.id` | `string` | The user ID (OBO) or service principal ID |
| `execution.obo_dev_fallback` | `boolean` | `true` when an OBO call falls back to SP in dev mode (no `x-forwarded-access-token`) |

### OTel context propagation fix

`executeStream()` runs the interceptor chain inside an async generator — OTel lost the parent HTTP span context at this boundary. Spans were created but orphaned in separate traces. Fix: capture `otelContext.active()` before the generator, restore with `otelContext.with()` inside.

### OBO dev mode fallback

In dev mode without `x-forwarded-access-token`, `asUser()` returns a thin proxy that sets an OTel context key (`DEV_OBO_FALLBACK_KEY`). The telemetry interceptor reads this to set `execution.obo_dev_fallback: true`. No mutable state — scoped automatically per execution via OTel context.

### `caller.id` fix

Previously used `context.userKey` (a cache key — analytics passes `"global"` for SP queries). Now uses `getCurrentUserId()` which always returns the real identity.

### Lakebase `db.user`

Adds `db.user` attribute to `lakebase.query` spans showing the PostgreSQL role used for the connection.

### Arrow-result clarification

Added comment clarifying the arrow-result route intentionally bypasses the interceptor chain (data download, not query execution).

## Test plan

- [x] Unit tests for all new attributes (user/service/dev fallback paths)
- [x] All 1578 tests pass
- [x] Full build pipeline passes
- [x] Manual: Grafana traces verified via Chrome DevTools MCP + Tempo API

### Service Principal

image

### User (for deployed app - when we have the forwarded token)

image

### Local for OBO

obo-trace-screenshot

### Lakebase - DB user

image

## Plugin OBO coverage verification

All built-in plugins that use `asUser()` go through `execute()` / `executeStream()`:

| Plugin | Method | File |
|---|---|---|
| analytics | `executeStream()` | [analytics.ts:187](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/analytics/analytics.ts#L187) |
| genie | `executeStream()` | [genie.ts:124](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/genie/genie.ts#L124), [:176](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/genie/genie.ts#L176), [:228](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/genie/genie.ts#L228) |
| files | `execute()` | [plugin.ts:445](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L445), [:481](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L481), [:551](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L551), [:622](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L622), [:658](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L658), [:694](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L694), [:782](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L782), [:842](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L842), [:887](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/files/plugin.ts#L887) |
| serving | `execute()` | [serving.ts:297](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/serving/serving.ts#L297) |
| vector-search | `execute()` | [vector-search.ts:98](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/vector-search/vector-search.ts#L98), [:173](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/vector-search/vector-search.ts#L173), [:240](https://github.com/databricks/appkit/blob/main/packages/appkit/src/plugins/vector-search/vector-search.ts#L240) |
| lakebase | N/A (OBO not supported yet) | `db.user` attribute added to `lakebase.query` spans showing the PostgreSQL role. `execution.context`/`caller.id` will be added when OBO (per-user pools) is implemented. |

**Not covered (by design):**
- **analytics arrow-result** (`GET /arrow-result/:jobId`) — data download endpoint, always runs as SP. Clarified with comment.

### DIFF
--- a/packages/appkit/src/plugin/interceptors/telemetry.ts
+++ b/packages/appkit/src/plugin/interceptors/telemetry.ts
@@ -1,4 +1,5 @@
 import type { TelemetryConfig } from "shared";
+import { isInUserContext } from "../../context/execution-context";
 import type { ITelemetry, Span } from "../../telemetry";
 import { SpanStatusCode } from "../../telemetry";
 import type { ExecutionInterceptor, InterceptorContext } from "./types";
@@ -24,6 +25,12 @@ export class TelemetryInterceptor implements ExecutionInterceptor {
       spanName,
       { attributes: this.config?.attributes },
       async (span: Span) => {
+        span.setAttribute(
+          "execution.context",
+          isInUserContext() ? "user" : "service",
+        );
+        span.setAttribute("caller.id", context.userKey);
+
         let abortHandler: (() => void) | undefined;
         let isAborted = false;
 
--- a/packages/appkit/src/plugin/interceptors/telemetry.ts
+++ b/packages/appkit/src/plugin/interceptors/telemetry.ts
@@ -1,5 +1,8 @@
 import type { TelemetryConfig } from "shared";
-import { isInUserContext } from "../../context/execution-context";
+import {
+  getCurrentUserId,
+  isInUserContext,
+} from "../../context/execution-context";
 import type { ITelemetry, Span } from "../../telemetry";
 import { SpanStatusCode } from "../../telemetry";
 import { isDevOboFallback } from "../plugin";
@@ -30,7 +33,7 @@ export class TelemetryInterceptor implements ExecutionInterceptor {
           "execution.context",
           isInUserContext() ? "user" : "service",
         );
-        span.setAttribute("caller.id", context.userKey);
+        span.setAttribute("caller.id", getCurrentUserId());
         if (isDevOboFallback()) {
           span.setAttribute("execution.obo_dev_fallback", true);
         }
--- a/packages/appkit/src/plugin/interceptors/telemetry.ts
+++ b/packages/appkit/src/plugin/interceptors/telemetry.ts
@@ -2,6 +2,7 @@ import type { TelemetryConfig } from "shared";
 import { isInUserContext } from "../../context/execution-context";
 import type { ITelemetry, Span } from "../../telemetry";
 import { SpanStatusCode } from "../../telemetry";
+import { isDevOboFallback } from "../plugin";
 import type { ExecutionInterceptor, InterceptorContext } from "./types";
 
 export class TelemetryInterceptor implements ExecutionInterceptor {
@@ -30,6 +31,9 @@ export class TelemetryInterceptor implements ExecutionInterceptor {
           isInUserContext() ? "user" : "service",
         );
         span.setAttribute("caller.id", context.userKey);
+        if (isDevOboFallback()) {
+          span.setAttribute("execution.obo_dev_fallback", true);
+        }
 
         let abortHandler: (() => void) | undefined;
         let isAborted = false;
--- a/packages/appkit/src/plugin/interceptors/telemetry.ts
+++ b/packages/appkit/src/plugin/interceptors/telemetry.ts
@@ -29,15 +29,6 @@ export class TelemetryInterceptor implements ExecutionInterceptor {
       spanName,
       { attributes: this.config?.attributes },
       async (span: Span) => {
-        span.setAttribute(
-          "execution.context",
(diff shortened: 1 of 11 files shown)


packages/appkit/src/plugin/interceptors/telemetry.ts
```diff
@@ -45,6 +50,15 @@ export class TelemetryInterceptor implements ExecutionInterceptor {
         }
 
         try {
+          span.setAttribute(
+            "execution.context",
+            isInUserContext() ? "user" : "service",
```
MarioCadenas, line 55: I'm not sure this will work, have you tried it works? the user context will only be set if asUser was called but it doesn't propagate down all the chain, did you test it?

pkosiec: Have you seen my screenshots? Yes, I tested it and those screenshots are from my tests 👍

Here's another fresh screenshot for a simulated call with OBO  😄 

To reproduce it you can do this:
```
TOKEN=$(databricks auth token --profile DEFAULT | python3 -c "import json,sys; print(json.load(sys.stdin)['access_token'])")
                                                                                                                                                                                                  
  curl -X POST \
    -H "Content-Type: application/json" \                                                                                                                                                         
    -H "x-forwarded-access-token: $TOKEN" \                                                                                                                                                     
    -H "x-forwarded-user: pawel.kosiec@databricks.com" \
    -d '{}' \                                                                                                                                                                                     
    "http://localhost:8001/api/analytics/query/untagged_apps"
```

---

**Context:**
- `asUser(req)` → `_createUserContextProxy(userContext)` wraps every method call in `runInUserContext(userContext, fn)`, which uses `AsyncLocalStorage.run()` — this propagates through all async operations including the interceptor chain
- When `TelemetryInterceptor` calls `isInUserContext()` / `getCurrentUserId()`, the `AsyncLocalStorage` store is still active

**Manually tested all three paths against the local OTEL stack (Grafana/Tempo):**

- **Service principal** (no `asUser`): `execution.context = "service"`, `caller.id = <SP ID>` — no `obo_dev_fallback`
- **OBO dev fallback** (`asUser(req)` without token in dev mode): `execution.context = "service"`, `caller.id = <SP ID>`, `execution.obo_dev_fallback = true`
- **Real OBO** (`asUser(req)` with `x-forwarded-access-token`): `execution.context = "user"`, `caller.id = pawel.kosiec@databricks.com` — no `obo_dev_fallback`

Covered by unit tests in `telemetry-interceptor.test.ts:128-200` and `plugin.test.ts:938-1108`.

MarioCadenas (approved):

PR #290, opened 2026-04-21
feat: add appkit codemod customize-callback CLI

MarioCadenas:
## Summary

Adds `appkit codemod customize-callback` CLI command to auto-migrate apps from the old `autoStart`/`extend`/`start` pattern to the new `onPluginsReady` callback.

Depends on #280.

## What it does

- **Pattern A** (.then chain): Extracts callback body, removes `.start()`, injects as `onPluginsReady` property, preserves `.catch()` handlers (including arrow functions with parens)
- **Pattern B** (await + imperative): Gathers extend/start calls, injects into `onPluginsReady`, removes standalone statements. Bails with warning for complex cases (non-server usage of appkit handle, multiple extend calls)
- **autoStart stripping**: Removes `autoStart: true/false` from all `server()` configs, preserves other properties
- **File discovery**: Auto-detects files importing `createApp` from `@databricks/appkit` in `server/`, `src/`, root
- **Idempotent**: No-ops with friendly message if already migrated
- **Dry-run by default**: `--write` flag to apply changes

## Usage

```bash
npx appkit codemod customize-callback              # dry-run, auto-detect
npx appkit codemod customize-callback --write      # apply changes
npx appkit codemod customize-callback --path file  # specific file
```

## Bugs fixed from review

- H1: Fixed `bodyOpenBrace` offset calculation after `.trim()` — was using trimmed string index as offset into original content
- H2: Fixed `.catch()` handler parsing — now uses brace-aware `findMatchingClose` instead of regex, correctly handles `.catch((err) => console.error(err))`
- M1: Pattern B now bails with warning when multiple `.extend()` calls are detected

## Test plan

- [x] 9 tests covering both patterns, bail-outs, idempotency, autoStart stripping, arrow function catch handlers, multi-extend bail
- [x] Full suite passes (1571/1571)
- [x] Typecheck passes
- [x] Build passes


MarioCadenas:
Superseded by new PR with renamed codemod (customize-callback -> on-plugins-ready).

PR #296, opened 2026-04-21
refactor(appkit): relocate shared agent utilities into plugins/agents

MarioCadenas:
## Summary

Mechanical move: with the new `agents()` plugin as the canonical home for all agent-related runtime code, the pre-existing shared utilities that lived in the old `plugins/agent/` (singular, deprecated) folder move into `plugins/agents/` (plural).

Files moved:
- `defaults.ts`, `event-translator.ts`, `schemas.ts`, `system-prompt.ts`, `thread-store.ts`
- Entire `tools/` subfolder (`tool.ts`, `define-tool.ts`, `function-tool.ts`, `hosted-tools.ts`, `mcp-client.ts`)
- `build-toolkit.ts`

Plus the associated test files.

Import paths updated across all call sites. Public exports in `packages/appkit/src/index.ts` and `packages/appkit/src/plugins/agents/index.ts` re-point at the new locations. No behavior change.

The old `plugins/agent/` folder still exists in this branch — it's deleted in PR #299 once all consumers have migrated off `createAgentApp`.

## PR Stack

1. Shared types + Adapters — #282
2. Tool types + MCP client — #283
3. Agent plugin core + ToolProvider implementations — #284
4. PluginContext mediator — #285
5. createAgent + agent-app + docs (v1) — #286
6. Plugin context-binding separation — #293
7. `agents()` plugin + `createAgent(def)` + `.toolkit()` — #294
8. agent-app + docs migrated to `agents()` — #295
9. **Relocate shared agent utilities** (this PR)
10. `preparePlugins` forwards eager instance — #297
11. `fromPlugin()` API — #298
12. Retire deprecated `agent()` + `createAgentApp` — #299
13. Retire `toPluginWithInstance` + bug fixes — #300

## Test plan

- [x] All 1300+ tests pass with updated import paths
- [x] Typecheck clean
- [x] Build clean

PR #295, opened 2026-04-21
feat(appkit): migrate agent-app and docs to the new agents() plugin

MarioCadenas:
## Summary

Cuts over the reference application and documentation to the `agents()` plugin API introduced in PR #294.

### Apps

- **`apps/agent-app/server.ts`** — rewritten from the deprecated `createAgent({ plugins, tools, port })` app-shortcut to `createApp({ plugins: [server(), analytics(), files(), agents({ agents: { support } })] })`. The `support` agent is a code-defined `createAgent(def)` that spreads tools from the analytics and files plugins via `.toolkit()` and mixes in inline `tool()` + `mcpServer()` helpers.
- **`apps/agent-app/config/agents/assistant.md`** — rewritten to the new YAML frontmatter schema (proper `---` delimiters, `endpoint:` / `default:` keys).

### Docs

- **`docs/docs/plugins/agents.md`** — full guide for the new plugin. Covers the five progressive levels (markdown-only → frontmatter scoping → code-defined agents → sub-agents → standalone `runAgent`). Includes the configuration reference and the full frontmatter schema table.
- **`docs/docs/guides/migrating-to-agents-plugin.md`** — migration guide from the old `createAgent` shortcut to `createApp + agents()`. Covers name changes, frontmatter format differences, code migration examples.
- Docs sidebar + regenerated typedoc API reference.

### CLI template

- **`template/appkit.plugins.json`** — plugin registry synced to include `agents` for `databricks apps init --features agents` scaffolding.

## PR Stack

1. Shared types + Adapters — #282
2. Tool types + MCP client — #283
3. Agent plugin core + ToolProvider implementations — #284
4. PluginContext mediator — #285
5. createAgent + agent-app + docs (v1) — #286
6. Plugin context-binding separation — #293
7. `agents()` plugin + `createAgent(def)` + `.toolkit()` — #294
8. **agent-app + docs migrated to `agents()`** (this PR)
9. Relocate shared agent utilities — #296
10. `preparePlugins` forwards eager instance — #297
11. `fromPlugin()` API — #298
12. Retire deprecated `agent()` + `createAgentApp` — #299
13. Retire `toPluginWithInstance` + bug fixes — #300

## Test plan

- [x] agent-app boots, registers both the code-defined support agent and the markdown-loaded assistant agent, streams responses
- [x] Docs build clean (`pnpm docs:build`)
- [x] Typecheck clean

PR #294, opened 2026-04-21
feat(appkit): add agents() plugin, createAgent() factory, and .toolkit()

MarioCadenas:
## Summary

Introduces the new agents system — pure-data `AgentDefinition`s, markdown-driven agent discovery, and the `agents()` plugin that registers them with `createApp`. Replaces the old `agent()` plugin + `createAgent` app-shortcut (PR #286) with three focused primitives:

- **`createAgent(def)`** — pure factory at `packages/appkit/src/core/create-agent-def.ts`. Returns the passed-in definition after cycle-detecting the sub-agent graph. No adapter construction, no side effects — safe at module top-level.
- **`agents(config)`** plugin — registers markdown agents from `config/agents/*.md` (via `js-yaml`) and merges code-defined agents. Builds per-agent `toolIndex` maps. Mounts `/invocations` + `/chat` routes. Exposes `appkit.agent.*` runtime helpers (`register`, `list`, `get`, `reload`, `getThreads`).
- **`.toolkit(opts?)`** method on core `ToolProvider` plugins (analytics/files/genie/lakebase) — returns a keyed record of `ToolkitEntry` markers, suitable for spreading into `AgentDefinition.tools`.

### Key behaviors

- **Markdown frontmatter** (real YAML, not regex-based): `endpoint`, `default`, `toolkits`, `tools`, `maxSteps`, `maxTokens`, `baseSystemPrompt`. Unknown keys logged and ignored.
- **Asymmetric auto-inherit**: markdown agents default to inheriting every registered `ToolProvider`'s tools (`file: true`); code-defined agents default to strict (`code: false`). Rationale: prompt authors want zero ceremony; engineers want explicit dependencies. Configurable via `autoInheritTools: { file: boolean, code: boolean }`.
- **Sub-agents**: `agents: { child }` inside an AgentDefinition exposes `agent-<key>` tools on the parent. Cycle-detected at build time.
- **`runAgent(def, input)`** standalone executor for tests / CLI scripts — drives the adapter outside of `createApp` with inline tools.

### New public exports

`agents`, `createAgent` (new pure-data meaning), `runAgent`, `isToolkitEntry`, `loadAgentFromFile`, `loadAgentsFromDir`, plus types `AgentDefinition`, `AgentsPluginConfig`, `AgentTool`, `ToolkitEntry`, `ToolkitOptions`, `PromptContext`, `BaseSystemPromptOption`.

### Dependencies

- Adds `js-yaml` runtime dep for frontmatter parsing.

## PR Stack

[...]

PR #293, opened 2026-04-21
refactor(appkit): separate plugin context binding from plugin construction

MarioCadenas:
## Summary

Splits the `Plugin` base-class lifecycle into two phases:

- **Construction** (now pure): `new Plugin(config)` does nothing side-effectful. Factories like `toPlugin` can return a `PluginData` tuple at module scope without pulling in `CacheManager`, `TelemetryManager`, or `PluginContext`.
- **`attachContext({ context, telemetryConfig })`**: runs inside `createApp` after all plugins have been constructed. Binds `telemetry`, `cache`, and the `PluginContext` mediator onto each instance, then fires `setup()`.

This unblocks the new `agents()` plugin (PR #294) which needs to construct agent definitions at module scope and resolve their plugin references lazily at `setup()` time, and it's a prerequisite for any factory that wants to hand back a typed, reusable plugin handle before `createApp` runs.

## Notes

- `ServerPlugin.setup()` now defers its HTTP start to the `setup:complete` lifecycle hook so that routes registered by `deferred`-phase plugins (like `agents()`) are visible when the server binds. Without this, `extendRoutes()` would run before `AgentsPlugin.injectRoutes()` had a chance to add its handlers.
- Adds `attachContext` to `BasePlugin` in `packages/shared/src/plugin.ts`.

## PR Stack

1. Shared types + Adapters — #282
2. Tool types + MCP client — #283
3. Agent plugin core + ToolProvider implementations — #284
4. PluginContext mediator — #285
5. createAgent + agent-app + docs (v1) — #286
6. **Plugin context-binding separation** (this PR)
7. `agents()` plugin + `createAgent(def)` + `.toolkit()` — #294
8. agent-app + docs migrated to `agents()` — #295
9. Relocate shared agent utilities — #296
10. `preparePlugins` forwards eager instance — #297
11. `fromPlugin()` API — #298
12. Retire deprecated `agent()` + `createAgentApp` — #299
13. Retire `toPluginWithInstance` + bug fixes — #300

## Test plan

- [x] Full vitest suite (unchanged test count, all passing)
- [x] Typecheck clean across all 8 workspace packages

PR #297, opened 2026-04-21
refactor(appkit): forward eager plugin instance through preparePlugins

MarioCadenas:
## Summary

Stepping-stone PR unblocking `fromPlugin` (PR #298).

At this point in the stack, `toPluginWithInstance` eagerly constructs a plugin instance (Instance A) at factory-call time and attaches it to the returned `PluginData` as an `.instance` field. Previously, `createApp`'s `preparePlugins` stripped that field before handing control to `createAndRegisterPlugin`, which then fell through to `new Plugin(baseConfig)` and constructed a second instance (Instance B). Dispatch used B; the handle the user held at module scope was A.

This PR:

- Widens `OptionalConfigPluginDef` in `packages/shared/src/plugin.ts` with an optional `instance?: BasePlugin` field.
- Forwards `pluginData.instance` through `preparePlugins` in `packages/appkit/src/core/appkit.ts`. The existing `preBuilt ?? new Plugin(baseConfig)` branch handles reuse — if `instance` is present, reuse it; otherwise construct fresh.

Effect: for plugins built via `toPluginWithInstance` (analytics, files, genie, lakebase), the handle a user holds at module scope is now the same object that answers runtime requests. Plugins built via plain `toPlugin` have no `instance` field and continue to construct fresh — no behavior change.

## Why this is its own PR

The instance-forwarding fix is small, mechanical, and reviewable in isolation. PR #298's `fromPlugin` builds on top of it.

**Note for reviewers**: this plumbing is *reverted* later in the stack (PR #300) once `toPluginWithInstance` itself is retired in favor of `fromPlugin`. At that point no factory produces an `instance` field, so the forwarding logic becomes dead code and gets removed. We kept this PR in the stack rather than squashing because it was genuinely useful while it existed — it let #298 land incrementally on top of a working baseline, and it makes the `fromPlugin` design rationale legible in review.

## PR Stack

1. Shared types + Adapters — #282
2. Tool types + MCP client — #283
3. Agent plugin core + ToolProvider implementations — #284
4. PluginContext mediator — #285
5. createAgent + agent-app + docs (v1) — #286
6. Plugin context-binding separation — #293
7. `agents()` plugin + `createAgent(def)` + `.toolkit()` — #294
8. agent-app + docs migrated to `agents()` — #295
9. Relocate shared agent utilities — #296
10. **`preparePlugins` forwards eager instance** (this PR)
11. `fromPlugin()` API — #298
12. Retire deprecated `agent()` + `createAgentApp` — #299
13. Retire `toPluginWithInstance` + bug fixes — #300

## Test plan

- [x] New test `packages/appkit/src/core/tests/appkit.test.ts` asserting instance identity for `toPluginWithInstance` factories and unchanged behavior for plain `toPlugin`
- [x] Full vitest suite passing
- [x] Typecheck clean

### DIFF
--- a/packages/appkit/src/core/appkit.ts
+++ b/packages/appkit/src/core/appkit.ts
@@ -78,8 +78,7 @@ export class AppKit<TPlugins extends InputPluginMap> {
     };
     // If the factory eagerly constructed an instance (via
     // `toPluginWithInstance`), reuse it; otherwise construct now.
-    const preBuilt = (pluginData as { instance?: BasePlugin }).instance;
-    const pluginInstance = preBuilt ?? new Plugin(baseConfig);
+    const pluginInstance = pluginData.instance ?? new Plugin(baseConfig);
 
     if (typeof pluginInstance.attachContext === "function") {
       pluginInstance.attachContext({
@@ -232,9 +231,11 @@ export class AppKit<TPlugins extends InputPluginMap> {
   ) {
     const result: InputPluginMap = {};
     for (const currentPlugin of plugins) {
+      const instance = (currentPlugin as { instance?: BasePlugin }).instance;
       result[currentPlugin.name] = {
         plugin: currentPlugin.plugin,
         config: currentPlugin.config as Record<string, unknown>,
+        instance,
       };
     }
     return result;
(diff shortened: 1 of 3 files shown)


MarioCadenas:
Superseded by the v2 6-PR stack:

The v2 stack reorganizes the same work so no PR ships API that a later PR deletes. Start at #301 for the new entry point. Branches from this older stack are preserved unchanged.

PR #301, opened 2026-04-21
feat(appkit): shared agent types and LLM adapter implementations

MarioCadenas:
Foundation layer for the agents feature. Adds the portable type surface
that every downstream layer builds on, plus three LLM adapter
implementations so the agents plugin (later PR) can target whatever the
user has.

### Shared agent types

`packages/shared/src/agent.ts` — no behavior, just the type vocabulary:
`AgentAdapter`, `AgentEvent`, `AgentInput`, `AgentRunContext`,
`AgentToolDefinition`, `Message`, `Thread`, `ThreadStore`, `ToolAnnotations`,
`ToolCall`, `ToolProvider`, `ResponseStreamEvent`. Exported from the
shared barrel.

### Adapters

- `packages/appkit/src/agents/databricks.ts` — `DatabricksAdapter`:
  streams OpenAI-compatible completions against a Databricks Model
  Serving endpoint. The `fromServingEndpoint` / `fromModelServing`
  factories route through the shared `connectors/serving/stream`
  helper, which delegates to the SDK's `apiClient.request({ raw: true })`.
  That gives the adapter centralised URL encoding + authentication
  with the rest of the serving surface — no bespoke `fetch()` +
  `authenticate()` plumbing. The raw `new DatabricksAdapter({ endpointUrl,
  authenticate })` constructor is preserved as an escape hatch for
  tests and for pointing the adapter at non-workspace endpoints.

Each adapter is self-contained and independently testable.

### Test plan

- 17 adapter tests (Databricks) covering raw-fetch escape hatch,
  serving-connector routing, URL encoding of endpoint names with
  special characters, streaming + tool-call dispatch.
- Full appkit vitest suite: 1552 tests passing at stack tip.

## PR Stack

1. **Shared agent types + LLM adapters** (this PR)

## Demo

### DIFF
--- a/packages/appkit/src/agents/databricks.ts
+++ b/packages/appkit/src/agents/databricks.ts
@@ -199,7 +199,7 @@ export class DatabricksAdapter implements AgentAdapter {
   ): Promise<DatabricksAdapter> {
     const { workspaceClient, endpointName, maxSteps, maxTokens } = options;
     return new DatabricksAdapter({
-      streamBody: (body) =>
+      streamBody: (body, signal) =>
         // Cast through the structural shape: the connector types
         // `workspaceClient` as the SDK's concrete `WorkspaceClient`, but we
         // only need `apiClient.request`.
@@ -207,6 +207,7 @@ export class DatabricksAdapter implements AgentAdapter {
           workspaceClient as unknown as Parameters<typeof servingStream>[0],
           endpointName,
           body,
+          signal,
         ),
       maxSteps,
       maxTokens,
@@ -434,7 +435,16 @@ export class DatabricksAdapter implements AgentAdapter {
         }
       }
     } finally {
-      reader.releaseLock();
+      try {
+        await reader.cancel();
+      } catch {
+        // Best-effort: reader may already be closed or the stream errored.
+      }
+      try {
+        reader.releaseLock();
+      } catch {
+        // Lock may already be released after cancel.
+      }
     }
 
     const toolCalls: OpenAIToolCall[] = Array.from(
--- a/packages/appkit/src/agents/databricks.ts
+++ b/packages/appkit/src/agents/databricks.ts
@@ -7,6 +7,28 @@ import type {
 } from "shared";
 import { stream as servingStream } from "../connectors/serving/client";
 
+/** Default cap for a single incomplete SSE line tail (DoS guard). */
+const DEFAULT_MAX_SSE_LINE_CHARS = 1024 * 1024;
+
+/** Default cap for accumulated assistant text from `delta.content`. */
+const DEFAULT_MAX_STREAM_TEXT_CHARS = 4 * 1024 * 1024;
+
+/** Default cap for accumulated JSON arguments per streamed tool call index. */
+const DEFAULT_MAX_TOOL_ARGUMENT_CHARS = 2 * 1024 * 1024;
+
+function throwIfExceedsStreamLimit(
+  label: string,
+  currentLength: number,
+  chunk: string,
+  max: number,
+): void {
+  if (currentLength + chunk.length > max) {
+    throw new Error(
+      `DatabricksAdapter: ${label} exceeds configured limit (${max} UTF-16 code units)`,
+    );
+  }
+}
+
 /**
  * Transport shim: given an OpenAI-compatible request body, returns the raw
  * SSE byte stream from the serving endpoint. Injected at construction time so
@@ -28,6 +50,12 @@ interface RawFetchAdapterOptions {
   authenticate: () => Promise<Record<string, string>>;
   maxSteps?: number;
   maxTokens?: number;
+  /** Max length of one SSE line (including an incomplete tail in the buffer). */
+  maxSseLineChars?: number;
+  /** Max total length of assistant `delta.content` across the stream. */
+  maxStreamTextChars?: number;
+  /** Max length of streamed `function.arguments` per tool call index. */
+  maxToolArgumentsChars?: number;
 }
 
 /**
@@ -40,6 +68,9 @@ interface StreamBodyAdapterOptions {
   streamBody: StreamBody;
   maxSteps?: number;
   maxTokens?: number;
+  maxSseLineChars?: number;
+  maxStreamTextChars?: number;
+  maxToolArgumentsChars?: number;
 }
 
 type DatabricksAdapterOptions =
@@ -70,12 +101,18 @@ interface ServingEndpointOptions {
   endpointName: string;
   maxSteps?: number;
   maxTokens?: number;
+  maxSseLineChars?: number;
+  maxStreamTextChars?: number;
+  maxToolArgumentsChars?: number;
 }
 
 interface ModelServingOptions {
   maxSteps?: number;
   maxTokens?: number;
   workspaceClient?: WorkspaceClientLike;
+  maxSseLineChars?: number;
+  maxStreamTextChars?: number;
+  maxToolArgumentsChars?: number;
 }
 
 interface OpenAIMessage {
@@ -154,10 +191,19 @@ export class DatabricksAdapter implements AgentAdapter {
   private streamBody: StreamBody;
   private maxSteps: number;
   private maxTokens: number;
+  private maxSseLineChars: number;
+  private maxStreamTextChars: number;
+  private maxToolArgumentsChars: number;
 
   constructor(options: DatabricksAdapterOptions) {
     this.maxSteps = options.maxSteps ?? 10;
     this.maxTokens = options.maxTokens ?? 4096;
+    this.maxSseLineChars =
+      options.maxSseLineChars ?? DEFAULT_MAX_SSE_LINE_CHARS;
+    this.maxStreamTextChars =
+      options.maxStreamTextChars ?? DEFAULT_MAX_STREAM_TEXT_CHARS;
+    this.maxToolArgumentsChars =
+      options.maxToolArgumentsChars ?? DEFAULT_MAX_TOOL_ARGUMENT_CHARS;
 
     if (isStreamBodyOptions(options)) {
       this.streamBody = options.streamBody;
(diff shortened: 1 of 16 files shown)


packages/appkit/src/agents/databricks.ts
```diff
@@ -0,0 +1,775 @@
...
+): o is StreamBodyAdapterOptions {
+  return "streamBody" in o;
+}
+
+/**
+ * Minimal structural shape consumed by `connectors/serving/stream`. We avoid
+ * importing the concrete `WorkspaceClient` type to keep the adapter free of a
+ * compile-time dependency on the SDK.
+ */
+interface WorkspaceClientLike {
```
calvarjorge, line 126: Is there any purpose to this interface? It's not being used for anything that I see, could we just use `undefined` when this is used, or sth like `type WorkspaceClientLike = undefined;`

packages/appkit/src/agents/langchain.ts
```diff
@@ -0,0 +1,292 @@
...
+          }
+          if (chunk?.tool_call_chunks) {
+            for (const tc of chunk.tool_call_chunks) {
+              const idx = tc.index ?? 0;
+              const existing = toolCallAccumulator.get(idx);
+              if (existing) {
+                if (tc.args) existing.arguments += tc.args;
+                // Later chunks for the same tool call may carry the id/name
+                // that the first chunk lacked.
+                if (tc.id && !existing.id.startsWith("lc_"))
```
calvarjorge:
```suggestion
                if (tc.id && existing.id.startsWith("lc_"))
```
the ID that starts with `lc_` is the fallback one that should be replaced right?

packages/appkit/src/agents/databricks.ts
```diff
@@ -0,0 +1,775 @@
...
+  } catch {
+    return body;
+  }
+}
+
+/**
+ * Creates a TransformStream that rewrites tool names in SSE response chunks
+ * (double-underscores -> dots).
+ */
+function createToolNameRewriteStream(): TransformStream<
```
calvarjorge: We want to use this for the tool names, but we're applying it to the whole response body. Could the regex be too generic and also be applied unexpectedly in the content? Could we just apply this to the tool name?

packages/appkit/src/agents/langchain.ts
ditadi: **[F2] HIGH — Inverted `lc_` fallback condition (calvarjorge's prior comment is correct)**

```ts
if (tc.id && !existing.id.startsWith("lc_"))
  existing.id = tc.id;
```

The block comment two lines above states the intent: "Later chunks for the same tool call may carry the id/name that the first chunk lacked." The synthesized fallback at L125 is `lc_${name}_${idx}_${++fallbackIdx}` — always starts with `"lc_"`. So the condition should be `existing.id.startsWith("lc_")` (drop the `!`).

As written, the logic is the **opposite** of intent:
- First chunk has no id → synthesize `lc_search_0_1`. Second chunk arrives with real `call_abc` → `!"lc_search_0_1".startsWith("lc_")` is `false` → real id silently dropped. The synthesized id leaks to clients as `tool_call.callId` and never matches the model's actual id.

Pair stays internally consistent because `runIdToCallId` uses the same synthesized id, but Responses-API correlation downstream is broken.

**Suggested fix:**
```suggestion
        if (tc.id && existing.id.startsWith("lc_"))
```

**Test gap:** `tests/langchain.test.ts:248` covers "no model id ever" and L189 always provides id on first chunk. The exact "first chunk lacks id, later chunk brings it" sequence is untested — that's why the bug ships. Recommend adding a streaming test that splits id arrival across two chunks.

Note: 5 reviewers initially read this as "correct as written" — orchestrator verified by direct read of the file. The condition is inverted.

packages/appkit/src/agents/databricks.ts
MarioCadenas: there's a description on top to explain why we need it now, thanks for a raising it 😄

packages/appkit/src/agents/databricks.ts
```diff
@@ -0,0 +1,757 @@
...
+ *
+ * @example Using the raw constructor
+ * ```ts
+ * const adapter = new DatabricksAdapter({
+ *   endpointUrl: "https://host/serving-endpoints/my-endpoint/invocations",
+ *   authenticate: async () => ({ Authorization: `Bearer ${token}` }),
+ * });
+ * ```
+ */
+export class DatabricksAdapter implements AgentAdapter {
```
pkosiec, line 223: Why `DatabricksAdapter`? maybe it's better to rename it to `ModelServingAdapter`?

packages/appkit/src/agents/databricks.ts
```diff
@@ -0,0 +1,757 @@
...
+ *     }),
+ *   ],
+ * });
+ * ```
+ *
+ * @example Using the raw constructor
+ * ```ts
+ * const adapter = new DatabricksAdapter({
+ *   endpointUrl: "https://host/serving-endpoints/my-endpoint/invocations",
+ *   authenticate: async () => ({ Authorization: `Bearer ${token}` }),
```
pkosiec, line 219: I assume it's low-level so user doesn't need to provide the bearer token right?

packages/appkit/src/agents/databricks.ts
pkosiec, line 1: agentic review:


# Fix Plan: `agent/v2/1-types-adapters` Review Findings

## Context

Code review of the `agent/v2/1-types-adapters` branch (7 commits, ~1700 lines) identified 12 findings across 9 parallel reviewers. This plan addresses the actionable items. The branch adds the foundation layer for the agents feature: shared agent types, `DatabricksAdapter` for Model Serving SSE streaming, and AbortSignal support in the serving connector.

## Must Fix (P1)

### 1. Wire name collision guard
**File:** `packages/appkit/src/agents/databricks.ts:335-341`

The dot-to-`__` replacement (`analytics.query` → `analytics__query`) assumes no tool is already named with double underscores. If tools `foo.bar` and `foo__bar` both exist, `wireToName` silently overwrites the first mapping, dispatching calls to the wrong tool.

**Fix:** After building the maps, check for collisions:
```ts
for (const tool of input.tools) {
  const wire = tool.name.replace(/\./g, "__");
  if (wireToName.has(wire) && wireToName.get(wire) !== tool.name) {
    throw new Error(
      `Tool name collision: '${tool.name}' and '${wireToName.get(wire)}' both map to wire name '${wire}'`,
    );
  }
  nameToWire.set(tool.name, wire);
  wireToName.set(wire, tool.name);
}
```

### 2. Missing DCO signoff on commit `ee3a677e`
All other 6 commits have proper `Signed-off-by:` trailers. This one is missing it.

**Fix:** `git rebase --signoff` the branch, or squash into the next commit.

### 3. Extract shared tool execution helper
**File:** `packages/appkit/src/agents/databricks.ts:372-413, 575-616`

The structured tool-call loop in `run()` and `executeToolCalls()` contain near-identical try/catch, yield, and message-push sequences (~40 duplicated lines).

**Fix:** Extract a private async generator method:
```ts
private async *executeSingleTool(
  tc: OpenAIToolCall,
  originalName: string,
  messages: OpenAIMessage[],
  context: AgentRunContext,
): AsyncGenerator<AgentEvent, void, unknown> { ... }
```
Call it from both the structured path (line 372 loop body) and `executeToolCalls()` (line 575 loop body).

## Should Fix (P2)

### 4. Wrap streamBody in try-catch
**File:** `packages/appkit/src/agents/databricks.ts:436`

If `authenticate()` throws or the fetch rejects, the error propagates uncaught from the generator with no status event.

**Fix:** Wrap the `this.streamBody(body, context.signal)` call in try-catch. Yield `{ type: "status", status: "error", error: err.message }` before rethrowing.

### 5. ReDoS mitigation in text tool-call parsers
**File:** `packages/appkit/src/agents/databricks.ts:712, 738`

- `tryParseLlamaJsonToolCalls`: Replace `[\s\S]*` regex with manual `indexOf('[{')` + `lastIndexOf('}]')` substring extraction, then `JSON.parse` the slice.
- `tryParsePythonStyleToolCalls`: Add an input length cap (e.g., 64KB) before running the regex fallback.

### 6. Narrow `any` types in SSE parsing
**File:** `packages/appkit/src/agents/databricks.ts:476, 498`

- Change `let parsed: any` to `let parsed: unknown` and add a minimal type guard for `choices[0].delta`.
- Replace `as DeltaToolCall[]` cast with runtime validation of `index`/`function` fields.

### 7. No timeout on raw fetch path
**File:** `packages/appkit/src/agents/databricks.ts:214`

If no `signal` is provided to the raw-fetch constructor path, create a default `AbortSignal.timeout(120_000)` as fallback.

### 8. Log malformed SSE lines
**File:** `packages/appkit/src/agents/databricks.ts:477-480`

Add debug-level logging when `JSON.parse` fails on an SSE `data:` line, so silent data loss from truncated deltas is at least observable.

## Nice to Have (P3)

### 9. Log reader.cancel() errors at debug level
**File:** `packages/appkit/src/agents/databricks.ts:528-537`

Instead of empty catch blocks, log at debug level for production observability.

### 10. Document tool error message responsibility
**File:** `packages/appkit/src/agents/databricks.ts:398-409`

Add a JSDoc note on `AgentRunContext.executeTool` that tool implementations should sanitize error messages since they are forwarded to the LLM.

### 11. Knip ignores for future files — no action
Preemptive ignores for stacked PR series (`agent/v2/4`, `agent/v2/5`). Intentional.

## Verification

1. `pnpm build && pnpm check:fix && pnpm -r typecheck`
2. `pnpm test` — all existing tests pass
3. Add a test for wire name collision (expects thrown error)
4. Add a test for `executeSingleTool` error path (executeTool rejects)
5. Verify DCO signoff: `git log --format='%h %s %b' | grep -c 'Signed-off-by'` matches commit count

packages/appkit/src/agents/databricks.ts
MarioCadenas: no because it will have model serving, supervisor api, etc or at least that's the idea I think

packages/appkit/src/agents/databricks.ts
MarioCadenas: yep

pkosiec (approved):

PR #307, opened 2026-04-22
feat(appkit): zero-trust MCP host policy with URL allowlist and scoped auth

MarioCadenas:
Closes the critical MCP-URL token-exfiltration surface identified in the agents stack review ([S1 in the security plan](https://github.com/databricks/appkit/pull/306)).

## The problem

Before this change, `AppKitMcpClient` accepted any absolute `http(s)://` URL as a hosted tool endpoint. It:

- Forwarded the **service-principal bearer** on every `initialize` / `notifications/initialized` / `tools/list` during `connect()`, to **any** URL — including a developer-supplied `mcpServer("x", "https://attacker.example.com/mcp")`.
- Forwarded the end-user's **OBO token** on every `tools/call` to the same unchecked URL.
- Accepted plaintext `http://` for any host.
- Had no defense against URLs whose hostname resolves to `169.254.169.254` (EC2/Databricks metadata), RFC1918, or CGNAT IPs.

Result: one mistyped or compromised MCP URL leaked SP + OBO workspace credentials without any user interaction.

## What ships

A new `mcp` block on `AgentsPluginConfig`:

\`\`\`ts
agents({
  mcp: {
    trustedHosts: ["mcp.corp.internal"],
    allowLocalhost: true, // default: NODE_ENV !== "production"
  },
});
\`\`\`

Rules enforced **before the first byte is sent** on every MCP URL (`connect`, `tools/call`, `notifications/initialized`):

1. Only `http:` and `https:` schemes.
2. Plaintext `http://` refused except for `localhost` / `127.0.0.1` / `::1` when `allowLocalhost` is on.
3. Hostname must match the resolved workspace host, equal localhost (if permitted), or appear in `trustedHosts`.
4. The resolved DNS address of the host must not land in loopback, RFC1918 (`10/8`, `172.16/12`, `192.168/16`), CGNAT (`100.64/10`), link-local (`169.254/16` — covers cloud metadata), `0.0.0.0/8`, multicast (`>=224.0.0.0`), ULA (`fc00::/7`), IPv6 link-local (`fe80::/10`), or IPv4-mapped IPv6 equivalents. IP-literal URLs in these ranges are rejected without a DNS lookup; malformed IPs fail-closed.
5. Workspace credentials (SP or OBO) are **never** forwarded to non-workspace hosts. Trusted external MCP servers must authenticate themselves (e.g., a token baked into the URL or a custom MCP header — follow-up).

`AppKitMcpClient.callTool` drops caller-supplied `Authorization` headers (typically the OBO bearer) whenever the destination was admitted with `forwardWorkspaceAuth: false`. `sendRpc` / `sendNotification` never invoke the workspace `authenticate()` closure on those routes.

### New — \`mcp-host-policy.test.ts\` (42 tests)

Covers the pure policy unit:

- \`buildMcpHostPolicy\`: trustedHosts normalization, \`NODE_ENV\` default for \`allowLocalhost\`, invalid workspace URL throws.
- \`checkMcpUrl\`: same-origin admit with auth, trusted-host admit without auth, plaintext http rejection (including "same workspace host wrong scheme"), non-http(s) scheme rejection, invalid URL rejection, case-insensitive hostname match, helpful error messages that name the rejected host and point at \`trustedHosts\`.
- \`isBlockedIp\`: RFC1918, link-local (169.254/16), CGNAT, 0.0.0.0/8, multicast, loopback (gated by \`allowLocalhost\`), ULA, IPv6 link-local, IPv4-mapped IPv6, malformed IP fail-closed.
- \`assertResolvedHostSafe\`: public-IP pass, metadata / RFC1918 rejection, DNS failure surfaced, empty result rejection, mixed-result "split DNS" defense.

### New — \`mcp-client.test.ts\` (8 tests)

Covers the integrated client with a recording \`fetchImpl\` and injected \`dnsLookup\`:

- \`connect\` rejects a non-allowlisted host **without making any fetch** and **without calling \`authenticate()\`**.
- \`connect\` rejects plaintext \`http://\` without making any fetch.
- \`connect\` rejects a URL whose DNS resolves to a blocked IP — SP token never touches the socket.
- Same-origin workspace \`connect\` sends SP on all three setup RPCs.
- Trusted external \`connect\` attaches no \`Authorization\` on any RPC.
- \`callTool\` drops a caller-supplied OBO token when the destination is external.
- \`callTool\` forwards the OBO token when the destination is same-origin workspace.
- \`callTool\` falls back to SP auth when no OBO override is provided and the destination is workspace.

### Gates

- \`pnpm typecheck\`: clean across all 8 workspace projects.
- \`pnpm biome check\`: clean.
- \`pnpm knip\`: clean.
- \`pnpm exec vitest run\` (full appkit project): **1361 tests passing** (up from 1311).

## Non-goals / follow-ups (not in this PR)

- Socket-level IP pinning to close DNS-rebinding (policy admits a hostname at \`connect\`, we rely on DNS caching + the short-lived connections; this is out of scope for a first cut).
- MCP session hardening (S9 in the security plan — separate PR).
- Tool-output redaction on thread persistence (S8).

## Drive-bys

- \`json-schema.ts\` reformatted — biome violation that pre-existed on the stack tip and would fail CI regardless of this PR.
- \`AppKitMcpClient\` constructor now accepts an optional \`{ dnsLookup, fetchImpl }\` options bag used by the new test suites.

## Base

This PR targets \`agent/v2/6-apps-docs\` so it lands on top of the agent stack (#301 → #306). Review + merge order is #301 → #306 first, then this.

## Test plan

- [x] New \`mcp-host-policy.test.ts\` — 42 tests
- [x] New \`mcp-client.test.ts\` — 8 tests
- [x] Existing appkit suite unchanged — 1361 total tests passing
- [x] \`pnpm typecheck\` clean
- [x] \`pnpm biome check\` clean
- [x] \`pnpm knip\` clean
- [x] \`pnpm generate:types\` produces no drift


MarioCadenas:
Closing — this fix has been split and folded into the stack where each piece belongs: policy module + client + tests into #302, `AgentsPluginConfig.mcp` + `connectHostedTools` wiring into #304, docs into #306. This makes every stack PR secure in isolation, so mid-stack merges don't leave a credential-exfil window on main.

PR #310, opened 2026-04-23
fix: unblock SP calls for UC in production

atilafassina:
Add on-behalf-of-user (OBO) support to the files plugin so SDK calls run as the end user, unblocking Unity Catalog volume access for OBO apps deployed on Databricks Apps.

## What changed

Per-volume opt-in to OBO via a new `auth` field on each volume config:

```ts
files({
  volumes: {
    user_uploads: { auth: "on-behalf-of-user", policy: usersOnly() },
    shared:       { auth: "service-principal" }, // default; existing behavior
  },
})
```

When `auth: "on-behalf-of-user"`:

- HTTP routes require `x-forwarded-user` + `x-forwarded-access-token` (401 in
  production if either is missing; dev mode falls back to SP identity with a
  warning).
- The identity passed to volume policies is the end user (not the SP).
- SDK calls execute via `runInUserContext` so UC sees the user, not the SP.
- `files.auth_mode` span attribute distinguishes OBO vs SP traffic in
  telemetry.
- Programmatic `appKit.files("vol").asUser(req).read(...)` works on any volume
  (forces OBO regardless of the volume's default auth mode).

> [!NOTE]
> OBO read tier intentionally disables caching (the cache layer keys by user; a write by user A wouldn't invalidate user B's view of the same path). Documented in the code; future work tracked for per-(volume, path) generation counters.

## What's NOT changed

- Default volume behavior is unchanged — existing apps keep `auth: "service-principal"` semantics with no migration.
- The static manifest still grants `WRITE_VOLUME` to the SP for every volume. OBO volumes need the *user* to also have UC permission, which has to be communicated out-of-band until the manifest schema gains a per-volume auth scope. (TODO noted in `getResourceRequirements`.)

## Hardening

- Upload `Content-Length` parsing (strict regex, rejects negative / partial /
  array values; stream-level byte counter as the real gate).
- Error-message hardening: `_handleApiError` no longer reflects raw SDK `error.message` on 4xx (CWE-209); generic body, raw to server log.
- `/read` route is atomic: drains the body into a memory buffer up to `maxReadSize`; on overflow responds 413 atomically (no partial 200 body leak).
- List-cache invalidation is symmetric: root-level writes invalidate both  `"__root__"` and `connector.resolvePath("/")` so `?path=/` listings stay fresh.
- `_extractObiUser` is strict in production (both headers required), warning +  SP fallback only in dev.
- `_invalidateListCache` is awaited before the HTTP success response so a  follow-up `GET /list` in the same tick can't observe stale data.

## Tests

- 113 plugin tests + 28 integration tests + 33 helpers/policy
- New tests cover: OBO read/write happy path + 401 paths, OBO write cross-user freshness (cache disabled), `asUser()` programmatic path, `files.auth_mode` span attribute on every code path, root-level cache invalidation key parity.

## Test plan

- [x] Unit tests in `packages/appkit/src/plugins/files/tests/`
- [x] Integration tests against a real Express app (`plugin.integration.test.ts`)
- [x] `pnpm -r typecheck`


packages/appkit/src/plugins/files/plugin.ts
```diff
@@ -100,21 +131,72 @@ export class FilesPlugin extends Plugin {
...
+   * Extraction for OBO (on-behalf-of-user) volumes on the HTTP path. Both the
+   * `x-forwarded-access-token` and `x-forwarded-user` headers must be present
+   * for a valid end-user identity. When the token is missing:
+   * - In production we throw `AuthenticationError.missingToken` so the route
+   *   responds with 401 (no SDK call is made).
+   * - In development (`NODE_ENV === "development"`) we emit a single warning
+   *   and fall back to the service principal identity so local testing
+   *   without a reverse proxy continues to work.
+   */
+  private _extractObiUser(req: express.Request): FilePolicyUser {
```
MarioCadenas:
```suggestion
  private _extractOboUser(req: express.Request): FilePolicyUser {
```

MarioCadenas:
## Agentic review — performance & placement notes (Xavier panel)

This comment summarizes **three medium-severity performance / cost-of-operations** findings from an automated review pass (Xavier remoras on `origin/sp-files` vs `origin/main`). They are **not correctness or security blockers**; they describe tradeoffs and tuning opportunities.

**Where to look:** [PR “Files changed”](https://github.com/databricks/appkit/pull/310/files) → `packages/appkit/src/plugins/files/plugin.ts` (search `_readSettings`, `_handleRead`).

---

### 1) HTTP `/read` no longer uses read-tier cache; uses `download` path

**Summary:** `/read` loads bytes via `connector.download` + `_downloadSettings`. Inline comments note `/read` **no longer participates in the read-tier cache**. For **service-principal** volumes, hot repeated `/read` of the same small object may hit the SDK/network every time instead of benefiting from the read cache → higher **latency**, **API usage**, and **load**.

**Anchor:** `_handleRead` — block calling `connector.download(getWorkspaceClient(), path)` with `_downloadSettings(mode)` (~L821–L833 on `sp-files`).

---

### 2) `/read` accumulates stream chunks then `Buffer.concat`

**Summary:** The handler reads via `getReader()`, **`chunks.push(value)`** per read, then **`Buffer.concat(chunks, bytesRead)`**. Many **tiny chunks** → extra **allocations**, a **large `chunks` array**, and **GC churn**, while staying within `maxReadSize`. Peak memory can spike briefly.

**Anchor:** `_handleRead` reader loop + `Buffer.concat` (~L843–L873 on `sp-files`).

---

### 3) OBO volumes: read-tier cache disabled by design

**Summary:** `_readSettings` sets **`cache.enabled: false`** when `authMode === "on-behalf-of-user"`. Correctness motivation: cache keys tie to `getCurrentUserId()`, so **cross-user staleness** would be wrong for OBO without richer keying. Tradeoff: **higher latency and backend cost** on OBO-heavy workloads until e.g. per-(volume, path) generation counters (as noted in comments).

**Anchor:** `_readSettings` (`isObo` branch) (~L542–L555 on `sp-files`).

---

### Bonus (docs — flagged by correctness + security remoras)

**Stale NOTE on `_enforcePolicy`:** The doc block still implies SDK execution stays on the service principal and OBO is a “later phase,” but handlers now use `_runWithAuth` / `runInUserContext` for OBO. Updating that NOTE avoids maintainer confusion (search for “still uses the service principal until a later phase”).

---

**Panel verdict (context):** correctness **approve**, security **approve**, performance **approve** — no `request changes` / `rethink` from the automated panel; items above are follow-ups / polish.

packages/appkit/src/plugins/files/plugin.ts
atilafassina: Done in cec17db4 — renamed at the declaration, the call site in `_enforcePolicy`, and the one test comment.

MarioCadenas (approved):

PR #315, opened 2026-04-27
fix: convert BigInt to strings in Arrow IPC row.toJSON()

jamesbroadhead:
## Summary

- `row.toJSON()` from apache-arrow preserves `BigInt` for Int64 columns. When this data is later serialized with `JSON.stringify()` for SSE streaming, it throws `TypeError: Do not know how to serialize a BigInt`.
- Convert BigInt values to strings in `_transformArrowAttachment` during row transformation, before the data reaches JSON serialization.
- This ensures Arrow IPC deserialization works correctly for serverless warehouses that return inline results as Arrow IPC in `result.attachment`.

## Test plan

- [ ] Verify that queries returning Int64 columns via Arrow IPC attachment no longer throw `TypeError: Do not know how to serialize a BigInt` during SSE streaming
- [ ] Verify that non-BigInt columns are unaffected by the change
- [ ] Verify that the string representation of BigInt values is correct and usable downstream

Replaces #270 (fork-based PR where CI couldn't run).

### DIFF
--- a/packages/appkit/package.json
+++ b/packages/appkit/package.json
@@ -69,6 +69,7 @@
     "@opentelemetry/sdk-trace-base": "2.6.0",
     "@opentelemetry/semantic-conventions": "1.38.0",
     "@types/semver": "7.7.1",
+    "apache-arrow": "21.1.0",
     "dotenv": "16.6.1",
     "express": "4.22.0",
     "obug": "2.1.1",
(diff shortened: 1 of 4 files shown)


jamesbroadhead:
Closing — superseded by #256.

The architectural change in #256 moved Arrow IPC decoding from server-side to client-side: the SQL warehouse connector now passes the base64 attachment through unchanged (with a 64MiB size cap), and the React hook decodes via the existing `ArrowClient.processArrowBuffer` pipeline. Apache Arrow JS in the browser handles BigInt natively, and there is no `JSON.stringify` in the data path on either side anymore — so the BigInt-vs-JSON.stringify failure mode this PR addressed no longer exists in #256.

Closing without merging.

PR #316, opened 2026-04-27
fix: handle ARROW_STREAM attachment in type generator

jamesbroadhead:
## Summary
- When serverless warehouses return ARROW_STREAM format, DESCRIBE QUERY results come as an inline base64 Arrow IPC `attachment` rather than `data_array`. This caused `convertToQueryType` to generate empty types `{}`.
- Added a fallback in `convertToQueryType` that decodes the Arrow IPC attachment schema to extract column names and types when `data_array` is empty.
- Extended `DatabricksStatementExecutionResponse` type with the optional `attachment` field.

## Test plan
- [ ] Verify type generation against a serverless warehouse that returns ARROW_STREAM format
- [ ] Verify existing JSON_ARRAY warehouses still produce correct types (no regression)
- [ ] Run `pnpm typecheck` and `pnpm test` to confirm no build or test regressions

Replaces #271 (fork-based PR where CI couldn't run).

### DIFF
--- a/packages/appkit/package.json
+++ b/packages/appkit/package.json
@@ -69,6 +69,7 @@
     "@opentelemetry/sdk-trace-base": "2.6.0",
     "@opentelemetry/semantic-conventions": "1.38.0",
     "@types/semver": "7.7.1",
+    "apache-arrow": "21.1.0",
     "dotenv": "16.6.1",
     "express": "4.22.0",
     "obug": "2.1.1",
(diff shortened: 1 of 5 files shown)


jamesbroadhead:
Closing — superseded by #256.

The ARROW_STREAM attachment decode fallback in `convertToQueryType` (and the type extension on `DatabricksStatementExecutionResponse.result.attachment`) was ported into #256 in commit e1e9017, along with the four supporting tests. #256 also restores the JSON_ARRAY-rejection retry loop so the new decode path actually engages on warehouses that don't support JSON_ARRAY + INLINE.

Closing without merging.

PR #318, opened 2026-04-27
chore(tools): add sse-timeout-probe for UI-01 / UI-02 empirical trace

jamesbroadhead:
## Summary

Adds `tools/sse-timeout-probe/` — a deterministic SSE / streaming-response probe for Databricks Apps. Built to characterize the cap reported in [ES-1742245] / UI-01 / UI-02; landing as a reusable diagnostic and the regression test for the platform fix.

[ES-1742245]: https://databricks.atlassian.net/browse/ES-1742245

## What we found

SSE streams through Databricks Apps Container were terminated at exactly **301.55s ± 18ms** regardless of heartbeat traffic — variance below 100ms across four hold values (300s, 400s, 500s, 600s) confirms an absolute, not idle, timer.

Source: `apps/oauth2-proxy/proxy.cfg` `upstream_timeout = "5m"`, mapped to Go's `transport.ResponseHeaderTimeout` at `apps/oauth2-proxy/pkg/upstream/http.go:143-144`. The stdlib documents that timer as header-receipt-only, but on HTTP/1.1 chunked-encoding upstream connections it also fires mid-body — a known practical issue acknowledged in `apps/runtime/pkg/proxy/local_proxy.go:217-221`.

Spaces-product-type apps front oauth2-proxy with TLS (HTTP/2) so the body-affecting variant of the quirk doesn't trigger; verified clean to 600s.

Other suspects ruled out:
- AWS NLB `connection-idle-timeout=300s` — TCP idle is reset by traffic; 5s heartbeats would have prevented any idle cut.
- Apps Envoy route `databricks-apps-cluster timeout: 60s` — observed lifetimes were 300s, so this 60s wasn't the binding cap.
- Lakegate (Apps Gateway) `request_timeout=60s` — verified empirically not body-cutting on Spaces apps; the hyper API contract holds (timeout wraps the headers-receipt future, not the body).

Concurrency check: 20 parallel 120s SSE streams completed cleanly; avg TTFB matched single-stream baseline. No HTTP/2 multiplexing penalty observed at this concurrency. The original UI-02 multiplexing-pressure hypothesis is not reproducing at this N — higher-N tests are listed as a follow-up.

## Universe fix

<https://github.com/databricks-eng/universe/pull/1867246> bumps `upstream_timeout` to `30m` on Container's oauth2-proxy. Container is the dominant product type and remains so; the fix needs to land there.

## Why this PR is still worth landing

The probe code is the regression test for the universe fix and a reusable diagnostic for any future SSE / streaming issue on Apps. Once the universe PR is broadly deployed, re-running the probe with `--durations 1500000 --heartbeat 5000` against any Container app should complete cleanly at ~25 minutes; a regression manifests as `outcome: server-close` at some lower value.

## What's in this PR

- **`probe.ts`** — opens one SSE connection per duration in a configurable ladder; records lifetime, bytes, end-cause (`completed` / `server-close` / `network-error` / `client-hard-timeout` / `auth-redirect` / `wrong-content-type`).
- **`server.ts`** — companion server speaking the `/sse-probe?hold-ms=…&heartbeat-ms=…` contract, for app images that include `npx` / `tsx`.
- **`server.py`** — Python stdlib equivalent of `server.ts` for app images that don't (e.g. Spaces).
- **`README.md`** — usage, the result-pattern interpretation table, and the recipe for verifying the universe fix.

## Test plan

- [x] Empirically reproduce the cap on dogfood Container app — done; 301.5s ± 18ms confirmed.
- [x] Verify Spaces apps don't experience the cap — done; clean to 600s.
- [x] Locate the source of the cap in universe — done; `apps/oauth2-proxy/proxy.cfg:34`.
- [ ] Once <https://github.com/databricks-eng/universe/pull/1867246> is broadly deployed, re-run `--durations 1500000 --heartbeat 5000` against a Container app and confirm clean completion at ~25 minutes.

## Follow-ups (out of scope for this PR)

In appkit's lane:
- WebSocket variant of the probe + wire both probes into `apps/dev-playground` (single piece of work).
- Promote AppKit's existing `Last-Event-ID` reconnection pattern (`StreamManager` + ring buffer + abort signals) as the recommended shape for long-running SSE apps. Defense in depth.
- Higher-N concurrency probe (N=100, N=500) to actually stress the HTTP/2 multiplexing hypothesis from the original UI-02 doc.
- CI regression test once the universe fix is broadly deployed.

Out of appkit, tracked elsewhere:
- Architectural fix to bring Container's oauth2-proxy hop to TLS / HTTP/2 (Apps platform / networking work).
- Durable docs of the diagnosis under `apps/tech-docs/` (universe doc PR after #1867246 lands).

This pull request and its description were written by Claude (claude.ai).

### DIFF
--- a/tools/sse-timeout-probe/README.md
+++ b/tools/sse-timeout-probe/README.md
@@ -1,0 +1,71 @@
+# sse-timeout-probe
+
+Empirical reproducer for the Databricks Apps SSE timeout gap
+([ES-1742245](https://databricks.atlassian.net/browse/ES-1742245), UI-01 / UI-02 in
+the EMEA Apps "gaps that matter" doc).
+
+## What it does
+
+Opens one SSE connection per duration in a configurable ladder, holds each one
+idle (or paced by a server-side heartbeat), and reports how long each connection
+actually survived and how it was terminated. Runs against a Databricks App, an
+EKS control, or `localhost`. Running it against a known-good origin and the
+Databricks-hosted one in sequence gives you the per-layer ceiling without
+having to triangulate from noisy LLM traces.
+
+## Why this exists
+
+The source doc (`Apps Gaps That Matter to EMEA Apps`) and ES-1742245 disagree
+about what kills SSE connections. The doc says the drop is "distinct from the
+120s idle timeout" and blames buffering / HTTP/2 multiplexing. The ticket's own
+diagnosis (Naïm Achahboun) says the drop *is* caused by the effective request
+timeout — multi-agent LLM calls take varying durations, so some finish under
+the ceiling and some don't.
+
+This probe answers the question deterministically: hold an idle SSE for X
+seconds, see if it survives, record where the timeout lives. Comparing results
+across durations and with/without heartbeat tells you whether the ceiling is
+idle-based (heartbeats save it) or absolute (heartbeats don't).
+
+## Usage
+
+```bash
+# Inside a workspace: deploy `server.ts` as the app entrypoint.
+# Then from a machine with network access to the app URL:
+tsx tools/sse-timeout-probe/probe.ts \
+  --base-url https://my-app.<workspace>.cloud.databricks.com \
+  --header "Cookie=<oauth2-proxy session cookie>" \
+  --durations 30000,60000,90000,120000,150000,180000,240000,300000 \
+  --json | tee apps-results.jsonl
+
+# Control run against the same codebase on EKS/localhost:
+tsx tools/sse-timeout-probe/probe.ts --base-url http://localhost:8000 \
+  --json | tee local-results.jsonl
+
+# Compare: if `outcome: server-close` clusters sharply at a duration in the
+# Apps run but not locally, that duration is your ceiling.
+```
+
+Flags:
+- `--base-url <URL>` (required) — base URL of the SSE-serving app.
+- `--path <PATH>` — SSE endpoint path. Default: `/sse-probe`.
+- `--durations <LIST>` — comma-separated milliseconds, one connection per entry. Default: `30000,60000,90000,120000,150000,180000,240000,300000`.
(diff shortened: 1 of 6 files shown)


pkosiec:
Hey James, thanks for putting this together!

A few thoughts:

1. **This seems like a one-time diagnostic tool** — once the probe runs and identifies the root cause, the code becomes obsolete. I'm not sure merging a temporary tool into AppKit's main branch is the right path, as it adds 660 lines + a new vitest test project for something that won't be needed afterward.
2. **This isn't really AppKit-scoped** — the probe targets Databricks Apps proxy/gateway timeout behavior, not the AppKit SDK itself.
3. **What if we skip the merge and go straight to the investigation?** The probe could be run from a branch or standalone repo just as well. Once the root cause is clear, the real value would be contributing the fix directly to the relevant Apps components (`apps/gateway`, `oauth2-proxy`, etc.). If you'd like to drive this, that would be great — otherwise I can pick it up in a week or so once the bigger roadmap items are wrapped up.

Thanks!

jamesbroadhead:
Closing in favor of https://github.com/databricks-eng/universe/pull/1869944, which adds the same probe at `apps/tools/sse-probe/` in universe.

The probe characterizes the Apps platform proxy chain (NLB, Envoy, Lakegate, oauth2-proxy), not appkit itself, so it belongs alongside the platform it tests and the fix landing alongside it (https://github.com/databricks-eng/universe/pull/1867246). The universe PR carries the empirical findings, the result-interpretation table, the migration note, and the recipe for verifying the upstream_timeout fix once it's broadly deployed.

No code changes are lost — `probe.ts`, `probe.test.ts`, `server.ts`, plus the `server.py` added in 9b74f3e, all moved over verbatim with path/name references updated.

Follow-ups (WS variant, dev-playground integration, AppKit-side reconnection-pattern docs) tracked in the universe PR's README.

PR #320, opened 2026-04-28
RFC: stash inline Arrow IPC server-side, serve via /arrow-result

jamesbroadhead:
**Status: DRAFT — for design discussion only, not for immediate merge.**

This PR proposes an alternative architecture for the inline-Arrow path landed in #256.

## Background

#256 lands `ARROW_STREAM + INLINE` support by emitting a new `{ type: "arrow_inline", attachment: <base64> }` SSE message and decoding the Arrow IPC bytes on the client. To make that work for non-trivial result sizes, #256 also bumps both the server and client SSE event-size caps from 1 MiB to 8 MiB.

That works, but has two architectural concerns:
1. **SSE was designed for streams of small control messages.** Pushing multi-MiB Arrow IPC bytes through SSE inflates the per-event memory ceiling on both server and client.
2. **Proxies and load balancers** sometimes have their own SSE event-size limits, independent of our defaults. The 8 MiB bump makes us depend on those tolerating larger events.

## What this PR does

Delivers ARROW_STREAM + INLINE responses out-of-band via the existing `/arrow-result/:jobId` endpoint:

1. The connector remains unchanged — still passes `result.attachment` through, still synthesizes empty schemas as base64 IPC.
2. The analytics route base64-decodes the attachment once, stashes the resulting buffer in `InlineArrowStash` (bounded LRU + TTL), and emits the **same** `{ type: "arrow", statement_id }` SSE message that EXTERNAL_LINKS already uses — but with a synthetic id prefixed `inline-`.
3. The client makes no distinction. It calls `/arrow-result/:jobId` for any `type: "arrow"` message (unchanged from `main`).
4. The route handler checks the stash first; if the id has the `inline-` prefix and is still present, it serves bytes from memory. Otherwise it falls through to the existing warehouse-fetch path for real EXTERNAL_LINKS statement_ids.

The `arrow_inline` SSE message type, the client-side base64 decode helper, and the SSE buffer-size bumps from #256 all go away. SSE returns to the 1 MiB default and only carries control messages.

## Why this is cleaner

| Aspect | #256 (`arrow_inline` over SSE) | This PR (stash + `/arrow-result`) |
|---|---|---|
| SSE event size pressure | 8 MiB per event | <1 KiB per event (just `statement_id`) |
| Proxy compatibility | Depends on proxy SSE event limits | Standard HTTP — works everywhere |
| Inline result size cap | 8 MiB (matches SSE) | 25 MiB (matches Databricks API hard limit) |
| Client wire-protocol branches | 2 (`arrow`, `arrow_inline`) | 1 (`arrow` for both INLINE and EXTERNAL_LINKS) |
| HTTP/2 streaming, gzip, background-fetch | Not applicable | Works naturally |

## Tradeoffs

- **Server-side memory.** Each stash entry holds the decoded IPC buffer until the client fetches it. `take()` is one-shot and removes the entry on read; TTL evicts unread entries. Worst case with default bounds (100 entries × 25 MiB) is ~2.5 GiB peak. Steady state is much smaller because the client fetches within a few seconds.
- **Single-process only.** A multi-server deployment would need a shared store (Redis or equivalent) keyed on the synthetic id. The stash interface is intentionally small (`put` / `take` / `size`) to make this swap straightforward.
- **One-shot reads.** Clients cannot retry a failed `/arrow-result` fetch — by the time they retry, the stash entry is gone. Acceptable for our case (the warehouse query returned data; the hook either succeeds or surfaces a generic error).

## Implementation surface vs #256

```
+1 file:  packages/appkit/src/plugins/analytics/inline-arrow-stash.ts
+1 file:  packages/appkit/src/plugins/analytics/tests/inline-arrow-stash.test.ts (10 tests)
~1 file:  packages/appkit/src/plugins/analytics/analytics.ts (route handler + stash wiring)
~1 file:  packages/appkit-ui/src/react/hooks/use-analytics-query.ts (drop arrow_inline branch)
~3 files: SSE/connector caps reverted to pre-bump values
```

10 net files, +296/−125 vs #256.

## Test status

- 1862/1862 tests pass
- Typecheck clean
- 10 new stash unit tests (LRU eviction, TTL, one-shot reads, prefix routing, UUID uniqueness)

## What I want to know

1. Is the server-side memory profile acceptable for the SDK's expected scale? (How many concurrent SSE clients does an AppKit deployment realistically have?)
2. Is single-process state a hard blocker, or is "single replica is fine, multi-replica needs Redis" an acceptable starting position?
3. Should the stash bounds (`maxEntries`, `ttlMs`) be configurable through the analytics plugin config?

If we like the direction, I'll cherry-pick this on top of #256 once it lands; otherwise we keep #256's 5b approach.


jamesbroadhead:
Closing in favour of #329.

Going with the SSE inline-arrow path with the 1 MiB → 8 MiB event-cap bump. The proxy/LB-tolerance and per-event memory ceiling concerns this RFC raised are real but speculative — we'll revisit if the 8 MiB cap actually bites in production. Until then, the simpler design wins.

(Comment drafted by Claude on James's behalf.)

PR #322, opened 2026-04-28
fix(appkit): allow undefined values in IGenieConfig.spaces

jamesbroadhead:
## Why

`IGenieConfig.spaces` is currently typed as `Record<string, string>`, but the plugin's runtime contract is already permissive of `undefined` values. `defaultSpaces()` constructs the same map and freely returns `{}` when the env var is unset:

```ts
private defaultSpaces(): Record<string, string> {
  const spaceId = process.env.DATABRICKS_GENIE_SPACE_ID;
  return spaceId ? { default: spaceId } : {};
}

private resolveSpaceId(alias: string): string | null {
  return this.config.spaces?.[alias] ?? null;
}
```

When env is unset, `spaces.default` is `undefined` at runtime; `?? null` coalesces it; routes 404 with "Unknown space alias". That's the intended, working behaviour. The type just doesn't admit it.

The mismatch surfaces the moment a caller wants more than one named alias and reaches for the env var directly:

```ts
genie({ spaces: { wanderbricks: process.env.DATABRICKS_GENIE_SPACE_ID } })
```

```
error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

The workarounds (`?? ''`, `as string`, hoist-and-narrow with a runtime throw) all paper over a type that's stricter than the code it describes.

## Change

`spaces?: Record<string, string>` → `spaces?: Record<string, string | undefined>`. `defaultSpaces()` return type updated to match. `resolveSpaceId` is unchanged — `?? null` already handles undefined values identically to missing keys.

## Alternatives considered

- **Require callers to hoist + assert.** Three lines per usage to fight a type that doesn't reflect the runtime. Rejected — perpetuates the lie.
- **Document `as string` in the README.** Type assertions silently mask config bugs at the callsite without changing runtime behaviour. Rejected — same hazard, less honest.
- **Make `spaces` keys `string` and require non-empty values via Zod.** Real validation, but a strictly bigger change and arguably the wrong layer — the plugin already returns 404 on miss, which *is* the validation step.

## Compatibility

Non-breaking. Function parameter types of `Record<K, T>` are contravariant for callers: existing callers passing `Record<string, string>` (a stricter map) still satisfy `Record<string, string | undefined>` (a looser map). `resolveSpaceId`'s `string | null` return is unchanged.

## Test plan

- [x] `pnpm -r typecheck` passes
- [x] `npx vitest run packages/appkit/src/plugins/genie/tests/genie.test.ts` — 30 passed, includes a new test asserting an explicitly-undefined alias 404s the same as a missing key
- [x] `npx biome check` on touched files — clean
- [x] `pnpm build` — appkit + appkit-ui rebuild green


calvarjorge (commented):
Should we actually fail when a user passes an undefined ID? Instead of waiting until the user queries the space. This seems more like a code error, and we can leave 404 for defined but inexistent spaces.

We could improve the error so that it's easier to understand the cause (instead of just type error)

jamesbroadhead:
Claude: good call — pushed 13ffc9f6.

The plugin now validates `config.spaces` in the constructor: any alias with an `undefined` value throws with a clear, named message (e.g. `GeniePlugin: space alias "wanderbricks" was configured with an undefined Genie Space ID. This usually means an environment variable used to populate the config is unset. Set the env var, or remove the alias from the config.`). Multiple offending aliases are all listed in one error.

404 "Unknown space alias" is now reserved for the case it was meant for: an alias the caller asks for at request time that simply isn't in the config.

The type stays `Record<string, string | undefined>` so callers can still pass `process.env.X` directly without a hoist-and-narrow — they just get a runtime fail-fast at construction instead of a misleading 404 later.

The old "should 404 when an alias maps to undefined" test is replaced with two new ones asserting the throw and that all offending aliases appear in the message.

packages/appkit/src/plugins/genie/genie.ts
```diff
@@ -51,7 +51,24 @@ export class GeniePlugin extends Plugin implements ToolProvider {
       maxMessages: 200,
     });
 
-    for (const alias of Object.keys(this.config.spaces ?? {})) {
+    const spaces = this.config.spaces ?? {};
+    const undefinedAliases = Object.entries(spaces)
+      .filter(([, id]) => id === undefined)
```
calvarjorge:
```suggestion
      .filter(([, id]) => !id)
```
We could also use this to check for empty strings & `null`

calvarjorge (approved):

packages/appkit/src/plugins/genie/genie.ts
jamesbroadhead: (I'm Claude, replying on James's behalf.)

Good call — taken in 22a9d5bd. The type is `string | undefined` so `!id` only catches `undefined` and `""`; renamed `undefinedAliases` → `missingAliases` and the error wording to "missing" to cover both. Added a test for the empty-string case.

calvarjorge (approved):

PR #326, opened 2026-04-29
fix(appkit): honor TTL in CacheManager.getOrExecute

### LINKED ISSUE #325: Cache TTL not honored in getOrExecute with in-memory storage (results never expire)
## Summary

`CacheManager.getOrExecute()` does not honor TTL when using `InMemoryStorage`. Once an entry is cached, the function passed to `getOrExecute` is never re-executed past the TTL — the stale value is served until the entry is evicted by LRU or the process restarts.

This silently breaks data freshness for any AppKit app using the analytics plugin without a Lakebase persistent cache, since the analytics plugin caches every query result via `getOrExecute` (`packages/appkit/src/plugins/analytics/analytics.ts`, defaults `enabled: true, ttl: 3600`).

In our deployed Databricks App we observed query results staying cached for **>19 hours** despite the default 1-hour TTL, until we set `cache: { enabled: false }` in `createApp`.

## Root cause

Two pieces of code interact:

1. `InMemoryStorage.get()` ([`packages/appkit/src/cache/storage/memory.ts`](https://github.com/databricks/appkit/blob/main/packages/appkit/src/cache/storage/memory.ts)) returns entries unconditionally, without checking `expiry`. This is documented as intentional — the test [`should return expired entry on get (expiry check is done by CacheManager)`](https://github.com/databricks/appkit/blob/main/packages/appkit/src/cache/tests/memory.test.ts) makes the contract explicit: expiry handling is the CacheManager's responsibility.

2. `CacheManager.getOrExecute()` ([`packages/appkit/src/cache/index.ts`](https://github.com/databricks/appkit/blob/main/packages/appkit/src/cache/index.ts)) calls `this.storage.get(cacheKey)` directly and returns `cached.value` if non-null, **without an expiry check**:

   ```ts
   const cached = await this.storage.get<T>(cacheKey);
   if (cached !== null) {
     // hit — returned immediately, expiry never checked
     return cached.value as T;
   }
   ```

   By contrast, the standalone `CacheManager.get()` method *does* check `Date.now() > entry.expiry`. So the expiry data is being written correctly; only the `getOrExecute` path skips the check.

The probabilistic `cleanupExpired` path doesn't help either — it's gated by `if (!this.storage.isPersistent()) return;`, so it never runs against in-memory storage.

Net result: with the in-memory backend, cached entries returned by `getOrExecute` live until LRU eviction at `maxSize` (default 1000) or process restart.

## Reproduction

Standalone Node script using the published `@databricks/appkit@0.11.0`:

```js
import { InMemoryStorage } from "@databricks/appkit/dist/cache/storage/memory.js";
import { CacheManager } from "@databricks/appkit/dist/cache/index.js";

const cm = await CacheManager.getInstance({
  enabled: true,
  ttl: 1,
  storage: new InMemoryStorage({ maxSize: 100 }),
});

let executions = 0;
const fn = async () => { executions++; return `result-${executions}`; };

const r1 = await cm.getOrExecute(["k"], fn, "u", { ttl: 1 });
await new Promise(r => setTimeout(r, 2500));   // wait past 1s TTL
const r2 = await cm.getOrExecute(["k"], fn, "u", { ttl: 1 });

console.log({ r1, r2, executions });
// Actual:   { r1: 'result-1', r2: 'result-1', executions: 1 }
// Expected: { r1: 'result-1', r2: 'result-2', executions: 2 }
```

`fn` is provably not re-executed after the TTL has elapsed.

For completeness, a parallel call to the standalone `cm.get()` for the same expired key correctly returns `null`, confirming the issue is isolated to the `getOrExecute` code path.

## Suggested fix

Either:

- In `CacheManager.getOrExecute`, replace the raw `storage.get` lookup with the expiry-aware logic from `CacheManager.get` (check `Date.now() > entry.expiry`, delete the entry, treat as miss). Inlining is straightforward; alternatively factor out a `getValid()` helper used by both.

- Or move the expiry check into the storage implementations and have all consumers go through it.

Either fix should be paired with a regression test in `cache-manager.test.ts`'s `describe("getOrExecute", ...)` block — the existing TTL test (`"should respect TTL expiry"`) only covers the `set` / `get` pair, not `getOrExecute`, which is why this slipped through.

I'm happy to put up a PR if the maintainers have a preferred direction.

## Environment

- `@databricks/appkit@0.11.0`
- Node 20 (issue is logic-level, not runtime-specific)
- Bug reproduces on `main` as well — same code paths.

## Workaround

Disable the cache globally:

```ts
await createApp({
  plugins: [server(...), analytics({})],
  cache: { enabled: false },
});
```

This short-circuits `getOrExecute` (`if (!this.config.enabled) return fn();`) and bypasses the bug entirely.

Tim-Hoare:
Fixes #325.

## Problem

`CacheManager.getOrExecute` returns cached entries without checking expiry, so when `InMemoryStorage` is the backend (the default when Lakebase isn't configured) cached values are served indefinitely — until LRU eviction at `maxSize` or process restart. The configured `ttl` is silently ignored on this code path.

This affects every AppKit app using the analytics plugin's default cache (`enabled: true, ttl: 3600`) without Lakebase. In a deployed Databricks App we observed query results staying cached for >19 hours despite a 1-hour TTL.

## Why it slipped through

`InMemoryStorage.get()` is intentionally a raw lookup — the existing test [`should return expired entry on get (expiry check is done by CacheManager)`](../blob/main/packages/appkit/src/cache/tests/memory.test.ts) makes that contract explicit. The standalone `CacheManager.get()` honors it correctly. But `CacheManager.getOrExecute` didn't, and the existing TTL test (`"should respect TTL expiry"`) only covered the `set` / `get` pair, not `getOrExecute`.

## Fix

In `getOrExecute`, after `storage.get(cacheKey)` returns an entry, check `Date.now() > cached.expiry`. If expired: delete it and fall through to the normal cache-miss path so `fn()` re-executes. Otherwise return the cached value as before.

Net change is small and stays within the existing "expiry handling lives at the CacheManager layer" contract — no signature changes, no storage changes.

## Test

Added a regression test in the `describe("getOrExecute", …)` block: caches a value with a 1ms TTL, waits, calls `getOrExecute` again, and asserts that `fn` was invoked twice and returns the new value.

Confirmed:
- Without the fix: new test fails (`expected 'result-1' to be 'result-2'`).
- With the fix: new test passes; all 1287 existing `appkit` package tests continue to pass.

Lint (`biome check`) and typecheck (`tsc --noEmit`) clean.

## Notes

- I picked the CacheManager-level fix to preserve the existing contract documented in the storage tests. Happy to refactor (e.g. push expiry into storage implementations, or factor a `getValid()` helper used by both `get` and `getOrExecute`) if a different shape is preferred.
- DCO sign-off included.


MarioCadenas (approved):
hey @Tim-Hoare, first of all thanks so much for reporting the bug and fixing it, sorry it took a bit to review 😅 

LGTM on the fix — the bug is real and the change correctly aligns `getOrExecute` with the same expiry contract that `get()` and `has()` already use. 


### Suggestion: extract a `getValid()` helper

You called this out in the PR description, and I think it's worth doing — the `Date.now() > entry.expiry` + `storage.delete` + treat-as-miss pattern now lives in three places (`get`, `has`, and the new branch in `getOrExecute`). If a fourth caller appears, this becomes a bug magnet (exactly how this PR happened).

A small private helper collapses it:

```ts
private async getValid<T>(
  key: string,
): Promise<{ value: T; expiry: number } | null> {
  const entry = await this.storage.get<T>(key);
  if (!entry) return null;
  if (Date.now() > entry.expiry) {
    await this.storage.delete(key);
    return null;
  }
  return entry;
}
```

As I said, up to you if you want to refactor it and implement the `getValid` method 😄 , happy to merge as is

Tim-Hoare:
Thanks @MarioCadenas! I have added the getValid helper if you would like to re-review

MarioCadenas (approved):

MarioCadenas:
Thanks a lot @Tim-Hoare ! approving and merging 😄

PR #330, opened 2026-04-29
docs: track aitools rename to top-level command

jamesbroadhead:
## Summary

Mirrors databricks/cli#4917 which promotes the aitools skills-management surface from `databricks experimental aitools` to top-level `databricks aitools`. Updates 3 install-command references in appkit so that user-facing copy points at the new canonical path:

- `packages/shared/src/cli/commands/setup.ts` (×2) — the CLAUDE.md snippet emitted by the appkit setup flow
- `docs/docs/index.md` (×1)

## Why now

Purely cosmetic — there is no deprecation warning to avoid. The CLI keeps `databricks experimental aitools install` working as a **silent backward-compat alias**, so the old path doesn't break and doesn't print anything. This PR is just keeping new docs aligned with the canonical command path so users learning aitools for the first time meet the right name.

## Note on the `tools` subtree

`databricks experimental aitools tools …` (query / discover-schema / get-default-warehouse / statement) is **not** being promoted in #4917 — it stays under `experimental`. appkit doesn't reference any of those commands, so nothing else here needed updating.

## Test plan

- [ ] `databricks-appkit setup` emits a CLAUDE.md that points at `databricks aitools install`
- [ ] Docs site renders the new command

### DIFF
--- a/docs/docs/development/ai-assisted-development.mdx
+++ b/docs/docs/development/ai-assisted-development.mdx
@@ -15,7 +15,7 @@ AppKit integrates with AI coding assistants through the Agent Skills.
 To install the Databricks Agent Skills for your preferred AI assistant, run:
 
 ```bash
-databricks experimental aitools install
+databricks aitools install
 ```
 
 ## Skills capabilities
--- a/docs/docs/_prerequisites.mdx
+++ b/docs/docs/_prerequisites.mdx
@@ -1,4 +1,4 @@
 ## Prerequisites
 
 - [Node.js](https://nodejs.org) v22+ environment with `npm`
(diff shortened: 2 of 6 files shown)


pkosiec:
Databricks CLI 1.0 has been released: https://github.com/databricks/cli/releases/tag/v1.0.0

I've just verified that the `aitools` command group is no longer under experimental, so we can merge this 👍

pkosiec (approved):

docs/docs/development/ai-assisted-development.mdx
```diff
@@ -15,7 +15,7 @@ AppKit integrates with AI coding assistants through the Agent Skills.
 To install the Databricks Agent Skills for your preferred AI assistant, run:
 
 ```bash
-databricks experimental aitools install
+databricks aitools install
```
pkosiec, line 18: Pushed a fix for the last missing occurrence 👍

PR #331, opened 2026-04-30
fix(appkit): harden SSE response headers in setupHeaders

MarioCadenas:
## What

Improves the SSE response headers set by `SSEWriter.setupHeaders()` to be more correct and proxy-friendly.

## Why

The previous header set had two issues that could cause problems in real deployments:

- **`Connection: keep-alive`** is forbidden by the HTTP/2 spec and was being stripped or rejected by HTTP/2 intermediaries anyway. Node.js manages keep-alive internally.
- **`Content-Encoding: none`** is not a valid header value and could trigger `502 Bad Gateway` or `RST_STREAM` errors in strict proxies (nginx, AWS ALB, GCP LB, Cloudflare).

There were also two missing proxy-buffering mitigations:

- **`Cache-Control: no-transform`** tells proxies not to modify the response body (important for streaming).
- **`X-Accel-Buffering: no`** is the de-facto header for disabling response buffering in nginx and nginx-backed CDNs/proxies (Cloudflare, AWS CloudFront, GCP, and most corporate reverse proxies honour it).

## Changes

| File | Change |
|------|--------|
| `packages/appkit/src/stream/sse-writer.ts` | Updated headers; `Connection` and `Content-Encoding` removed; `X-Accel-Buffering: no` added |

### DIFF
--- a/packages/appkit/src/plugins/analytics/tests/analytics.integration.test.ts
+++ b/packages/appkit/src/plugins/analytics/tests/analytics.integration.test.ts
@@ -105,7 +105,9 @@ describe("Analytics Plugin Integration", () => {
       );
 
       expect(response.status).toBe(200);
-      expect(response.headers.get("Content-Type")).toBe("text/event-stream");
+      expect(response.headers.get("Content-Type")).toBe(
+        "text/event-stream; charset=utf-8",
+      );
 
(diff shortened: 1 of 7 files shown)


atilafassina (approved):
Nice detective work! 🚀

PR #332, opened 2026-04-30
feat(appkit): send internal telemetry via AppkitLog schema

calvarjorge:
## Summary
- Introduce the `AppkitLog` events (`APP_STARTUP`, `HEARTBEAT`, `REQUEST_METRICS`) and a `TelemetryReporter` singleton that owns shared dispatch state, the periodic heartbeat, and per-endpoint request metrics aggregation.
- Server plugin records each matched route via `res.on('finish')` middleware; the reporter flushes one event per `(method, route)` on a periodic timer (configurable via `APPKIT_TELEMETRY_HEARTBEAT_INTERVAL_MS` / `APPKIT_TELEMETRY_METRICS_FLUSH_INTERVAL_MS`).
- POSTs to `/telemetry-ext?o=<workspaceId>` with the workspace SP bearer token. Errors propagate from the inner senders so consumers can see exactly what was sent and how the endpoint responded; the only swallows live at the SDK's outermost boundaries (fire-and-forget startup + interval timers).
- Adds an **Internal Telemetry** tab in dev-playground that lets you trigger each event on demand and renders the request, response, and an equivalent `curl` command. Disable globally with `disableInternalTelemetry: true` on `createApp` or `APPKIT_TELEMETRY_DISABLED=true`.
- Fixes: `app.yaml` was missing the `DATABRICKS_JOB_ID` binding the jobs plugin requires (deploys to a fresh app failed startup validation); knip false-positive on the `pnpm exec cdxgen`-invoked `@cyclonedx/cdxgen` dep blocked all pre-commit hooks.


packages/appkit/src/index.ts
```diff
@@ -47,6 +47,12 @@ export {
   TunnelError,
   ValidationError,
 } from "./errors";
+export {
+  TelemetryReporter,
+  type TelemetrySendRequest,
+  type TelemetrySendResponse,
+  type TelemetrySendResult,
+} from "./internal-telemetry";
```
pkosiec: Do we need to export those types? If so, why?

docs/docs/internal-telemetry.mdx
pkosiec, line 1: Maybe we should rework this page as Privacy policy or sth like that? Internal telemetry might be confused with plugin's telemetry 🤔 
Also, IMO it should be the last item on the sidebar

packages/appkit/src/core/appkit.ts
```diff
@@ -171,6 +176,7 @@ export class AppKit<TPlugins extends InputPluginMap> {
       cache?: CacheConfig;
       client?: WorkspaceClient;
       onPluginsReady?: (appkit: PluginMap<T>) => void | Promise<void>;
+      disableInternalTelemetry?: boolean;
```
pkosiec, line 199: I'd vote for removing that. IMO the environmental variable is enough 👍

docs/docs/internal-telemetry.mdx
```diff
@@ -0,0 +1,76 @@
...
+```
+
+**Environment (workspace-wide):**
+
+```sh
+DISABLE_APPKIT_INTERNAL_TELEMETRY=true
+```
+
+Either one fully disables the reporter — no events are emitted and no
+network calls are made.
```
pkosiec: We could also support `DO_NOT_TRACK=1` - https://donottrack.sh/

docs/docs/internal-telemetry.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Inspecting events locally
+
+The dev-playground ships an **Internal Telemetry** tab that lets you
+trigger each event type on demand and inspect the exact request, the
+response, and a `curl` command you can replay. Use it to verify what your
+deployed app would send before enabling telemetry in production.
```
pkosiec: Dev playground is only for us, AppKit maintainers and contributors. IMO we shouldn't describe it here.

apps/dev-playground/server/internal-telemetry-debug-plugin.ts
```diff
@@ -0,0 +1,120 @@
+import {
+  Plugin,
+  type PluginManifest,
+  TelemetryReporter,
+  type TelemetrySendRequest,
+  type TelemetrySendResult,
```
pkosiec: As I posted in other comments, we shouldn't export those. If so, does it still make sense to have such dev playground plugin? 
I tested it and not sure if it brings enough value - yeah, there are some requests sent but some integration tests could verify that already, right? Because I cannot judge if those requests are valid or not 🤔 


If that's still useful to have, maybe we should run a simple separate server near the dev playground setup? WDYT?

apps/dev-playground/client/src/routes/__root.tsx
pkosiec, line 1: Agentic review - please see if that makes sense:

# Code Review: Internal Telemetry (`jorge.calvar/send_telemetry`)

## Scope & Intent

**Branch:** `jorge.calvar/send_telemetry` (11 commits, ~31 files, base: `origin/main`)

**Intent:** Add internal telemetry reporting to AppKit that sends anonymized `APP_STARTUP`, `HEARTBEAT`, and `REQUEST_METRICS` events to the workspace's `/telemetry-ext` endpoint. Includes opt-out, a debug playground page, public documentation, and comprehensive tests.

**Mode:** Report-only (no fixes applied)

**Review team:**
- correctness (always) -- logic errors, edge cases, race conditions
- testing (always) -- coverage gaps, weak assertions
- maintainability (always) -- coupling, complexity, naming
- project-standards (always) -- CLAUDE.md compliance
- security -- auth tokens, header redaction, debug endpoints
- reliability -- fire-and-forget, timers, shutdown, circuit breaking
- api-contract -- new public exports, createApp signature change
- performance -- per-request middleware, bucket growth

---

## Findings

### P1 -- Critical / High

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 1 | `reporter.ts:150-153` | **Metrics data lost on flush failure.** `flushRequestMetrics()` swaps the bucket map *before* the send completes. If `postTelemetry()` throws, the drained metrics are lost with no recovery. Should swap only on success, or restore on failure. | reliability, correctness | 0.85 | `manual -> downstream-resolver` |
| 2 | `server/index.ts:339-382` | **Graceful shutdown doesn't stop TelemetryReporter.** `_gracefulShutdown()` never calls `TelemetryReporter.getInstance()?.stop()`. While timers are `.unref()`'d (won't block exit), in-flight HTTP requests race against `process.exit()` and the final metrics window is lost. | reliability | 0.85 | `safe_auto -> review-fixer` |
| 3 | `reporter.ts:91-99` | **No circuit breaker for persistent failures.** If `/telemetry-ext` is down, heartbeat and flush timers retry every 5min/1min forever. Each retry spawns a 10s fetch. No backoff, no max-attempt limit, no way to detect the retry storm. | reliability | 0.80 | `manual -> downstream-resolver` |
| 4 | `server/index.ts:419-440` | **requestMetricsMiddleware has zero test coverage.** The middleware is the critical integration point (Express -> reporter) but has no tests at all. Route template extraction, latency calculation, and edge cases (undefined `req.route`) are untested. | testing | 0.95 | `manual -> downstream-resolver` |
| 5 | `client.ts:62-84` | **AbortController timeout path is untested.** The 10s timeout is the main safeguard against hanging telemetry. No test simulates a slow/never-resolving fetch to verify the abort fires. | testing | 0.92 | `manual -> downstream-resolver` |

### P2 -- Moderate

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 6 | `client.ts:71` | **`redirect: "manual"` silently drops auth redirects.** If the workspace issues a 302 (e.g., to login), the response is returned as-is and the caller treats it as success. Telemetry data is silently lost. Consider checking `response.status` in the 3xx range and logging a warning. | correctness, security | 0.75 | `gated_auto -> downstream-resolver` |
| 7 | `reporter.ts:93-99` | **Error swallowing with no observability.** Timer callbacks use `.catch(() => {})`. If the endpoint is misconfigured, auth fails, or the network is down, there is zero diagnostic trail -- no log, no metric. Add `logger.debug('Telemetry send failed:', err.message)` to preserve fire-and-forget semantics while enabling diagnostics. | reliability | 0.78 | `safe_auto -> review-fixer` |
| 8 | `reporter.ts:49, 111-130` | **Unbounded bucket growth.** Each unique `method + routeTemplate` creates a new bucket entry. If flush fails repeatedly while `recordRequest()` continues, the map grows indefinitely. Add a max bucket count (e.g., 1000) or reset on failed flushes. | reliability | 0.78 | `gated_auto -> downstream-resolver` |
| 9 | `client.ts:60` | **`authenticate()` call not covered by timeout.** The AbortController timer starts *after* `authenticate()`. If the OAuth provider hangs, the 10s timeout doesn't apply. Wrap `authenticate()` in `Promise.race()` with the timeout. | reliability | 0.70 | `gated_auto -> downstream-resolver` |
| 10 | `reporter.ts`, `client.ts` | **TelemetryReporter & types exported without JSDoc.** API docs show `\| [TelemetryReporter](...) \| - \|` with missing descriptions. `TelemetrySendRequest/Response/Result` similarly undocumented. Add class/interface-level JSDoc. | api-contract | 0.82 | `safe_auto -> review-fixer` |
| 11 | `internal-telemetry-debug-plugin.ts` | **Debug plugin has zero test coverage (120 LOC).** `redactHeaders`, `toCurl`, error handling when reporter is null, and the record endpoint are all untested. | testing | 0.85 | `manual -> downstream-resolver` |
| 12 | `internal-telemetry-debug-plugin.ts` | **Debug plugin endpoints have no auth/env guard.** Anyone with network access to the app can trigger telemetry sends and inspect request/response details. Add a `NODE_ENV` check or prominent warning that this plugin is dev-only. | security | 0.85 | `gated_auto -> downstream-resolver` |
| 13 | `reporter.ts:126` | **Negative latency clamping (`Math.max(0, latencyMs)`) is untested.** Defensive code without test coverage risks silent removal in refactors. | testing | 0.88 | `safe_auto -> review-fixer` |

### P3 -- Low

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 14 | `.gitignore:51` | **Unrelated change:** `.superset/config.json` added to gitignore. Not part of the telemetry feature; should be a separate commit. | maintainability | 0.90 | `advisory -> human` |
| 15 | `app.yaml:64-65` | **Unrelated change:** `DATABRICKS_JOB_ID` binding added. Not part of internal telemetry. | maintainability | 0.85 | `advisory -> human` |
| 16 | `appkit.plugins.json` | **New 92-line generated file** committed alongside the telemetry feature. Appears unrelated to the telemetry work. | maintainability | 0.80 | `advisory -> human` |
| 17 | `reporter.ts`, `client.ts` tests | **Test helper duplication.** Both test files define identical `createMockClient()` functions. Extract to shared test helper. | maintainability | 0.75 | `advisory -> human` |
| 18 | `appkit-log.ts:44` | **`items: never[]` type is confusing.** Typed as `never[]` but assigned `[]`. A comment explaining the legacy payload format would help future readers. | maintainability | 0.65 | `advisory -> human` |
| 19 | `reporter.ts:136, 145` | **`placeholder: true` in startup/heartbeat events is undocumented.** No comment explaining why empty objects use a placeholder field (proto compat?). | maintainability | 0.65 | `advisory -> human` |

---

## Residual Risks

- **Opt-out by default:** This is a behavior change -- all apps will now send telemetry unless explicitly disabled. Users relying on network isolation may be surprised. The documentation and dual opt-out (code + env var) mitigate this, but it should be called out in the changelog.
- **Singleton conflicts:** Multiple `createApp()` calls in the same process overwrite the reporter singleton. The second app's metrics mix with the first's via the shared middleware.
- **`workspaceId` promise rejection:** If workspace ID resolution fails persistently, every send silently fails. No diagnostic trail exists until finding #7 is addressed.

## Testing Gaps

1. **requestMetricsMiddleware** -- zero coverage (critical integration point)
2. **AbortController timeout path** in `postTelemetry` -- untested
3. **Concurrent `recordRequest` + `flushRequestMetrics`** -- no race condition test
4. **Negative latency clamping** -- untested defensive code
5. **Debug plugin** -- 120 LOC with 0 tests
6. **`disableInternalTelemetry` integration** -- unit test exists via mock but no true integration test with env var
7. **Large payload serialization** -- no stress test for many buckets

---

## Recommended Improvement Plan (Ordered by Priority)

### 1. Fix data loss on flush failure (`reporter.ts`)
- In `flushRequestMetrics()`, keep the old buckets reference and only swap on successful send
- On failure, merge drained data back into `this.#buckets`
- **File:** `packages/appkit/src/internal-telemetry/reporter.ts:150-173`

### 2. Add graceful shutdown cleanup (`server/index.ts`)
- In `_gracefulShutdown()`, call `TelemetryReporter.getInstance()?.stop()`
- Optionally await a final `flushRequestMetrics()` before closing the server
- **File:** `packages/appkit/src/plugins/server/index.ts` (in the `_gracefulShutdown` method)

### 3. Add circuit breaker to reporter (`reporter.ts`)
- Track consecutive failure count in a `#failureCount` field
- After N consecutive failures (e.g., 3), pause sends for a backoff period
- Reset counter on success
- **File:** `packages/appkit/src/internal-telemetry/reporter.ts`

### 4. Add logging to error catch handlers (`reporter.ts`)
- Replace `.catch(() => {})` with `.catch((err) => logger.debug('Telemetry send failed:', err.message))`
- Also for `sendStartup().catch(() => {})` in `appkit.ts`
- **Files:** `reporter.ts:93-99`, `appkit.ts:244`

### 5. Add test coverage for requestMetricsMiddleware
- Test normal flow, undefined `req.route`, latency calculation, status code ranges
- **File:** new test file or addition to existing server plugin tests

### 6. Add timeout test for postTelemetry
- Mock fetch to never resolve, verify AbortError fires after TIMEOUT_MS
- **File:** `packages/appkit/src/internal-telemetry/tests/client.test.ts`

### 7. Add tests for debug plugin
- Cover `redactHeaders`, `toCurl`, 503 when reporter is null, record endpoint
- **File:** new `apps/dev-playground/server/internal-telemetry-debug-plugin.test.ts`

### 8. Add JSDoc to public API surface
- `TelemetryReporter` class, `TelemetrySendRequest/Response/Result` interfaces
- **Files:** `reporter.ts`, `client.ts`

### 9. Guard debug plugin for dev-only use
- Add `NODE_ENV` check or JSDoc warning in the plugin
- **File:** `apps/dev-playground/server/internal-telemetry-debug-plugin.ts`

### 10. Add bucket size cap (`reporter.ts`)
- Add `MAX_BUCKETS` constant (e.g., 1000)
- In `recordRequest()`, skip new entries if limit reached
- **File:** `packages/appkit/src/internal-telemetry/reporter.ts`

### 11. Wrap `authenticate()` in timeout (`client.ts`)
- Move AbortController creation before `authenticate()`
- Or use `Promise.race([authenticate(), timeoutPromise])`
- **File:** `packages/appkit/src/internal-telemetry/client.ts`

### 12. Split unrelated changes into separate commits
- `.superset/config.json` gitignore, `DATABRICKS_JOB_ID` binding, `appkit.plugins.json`

---

## Verdict

**Not ready** -- The core telemetry implementation is well-structured with good test coverage for the happy path, but finding #1 (data loss on flush failure) and #2 (missing shutdown cleanup) are P1 issues that should be fixed before merge. The missing test coverage for the middleware (#4) and timeout path (#5) are also significant gaps. The circuit breaker (#3) and error observability (#7) are important for production reliability.

After addressing items 1-7, this would be **Ready to merge**.

apps/dev-playground/app.yaml
```diff
@@ -5,6 +5,8 @@ env:
     valueFrom: genie-space
   - name: DATABRICKS_SERVING_ENDPOINT_NAME
     valueFrom: serving-endpoint
+  - name: DATABRICKS_JOB_ID
+    valueFrom: job
```
pkosiec, line 9: As long as we don't have databricks.yml it won't help as we need to set the resource manually 🤔 Maybe it's better to skip it?

apps/dev-playground/appkit.plugins.json
pkosiec, line 1: Why do we generate it now for the dev-playground? 🤔

.gitignore
```diff
@@ -11,3 +11,5 @@ coverage
 .turbo
 
 .databricks
+
+.superset/config.json
```
pkosiec: We can ignore the whole dir, right?

BTW, personally I added such user-specific files to my global gitignore - you can consider such approach too 👍

packages/appkit/src/internal-telemetry/reporter.ts
```diff
@@ -0,0 +1,194 @@
...
+        app_startup_event: { placeholder: true },
+      }),
+    ]);
+  }
+
+  async sendHeartbeat(): Promise<TelemetrySendResult | null> {
+    return this.#send([
+      this.#wrap({
+        event_name: "HEARTBEAT",
+        heartbeat_event: { placeholder: true },
```
pkosiec: why `placeholder: true`? shouldn't it be an empty object for now?

Same for startup

packages/appkit/src/internal-telemetry/index.ts
```diff
@@ -0,0 +1,13 @@
...
+// DISABLE_APPKIT_INTERNAL_TELEMETRY=true.
+// Full data inventory: docs/docs/internal-telemetry.mdx.
+
+export type {
+  TelemetrySendRequest,
+  TelemetrySendResponse,
+  TelemetrySendResult,
+} from "./client.js";
+export { isInternalTelemetryEnabled } from "./config.js";
+export { TelemetryReporter } from "./reporter.js";
```
pkosiec: Is there any reason why those exports are for the .js files?

packages/appkit/src/internal-telemetry/client.ts
```diff
@@ -0,0 +1,85 @@
...
+    return {
+      request: { url, method: "POST", headers: headersToObject(headers), body },
+      response: {
+        status: response.status,
+        statusText: response.statusText,
+        body: responseBody,
+      },
+    };
+  } finally {
+    clearTimeout(timer);
```
pkosiec: Something to check, but I believe we could use the client to do an arbitrary request with all the necessary headers already - I think it was client.apiClient.request() but not sure, pls double check if that's possible

packages/appkit/src/internal-telemetry/appkit-log.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export interface RequestMetricsEvent {
+  endpoint?: string;
+  request_count?: number;
+  request_latency_ms_avg?: number;
+  response_count_http4xx?: number;
+  response_count_http5xx?: number;
+}
+
+export interface AppkitLog {
```
pkosiec, line 23: Why `Log`? shouldn't we call it `AppEvent` / `AppKitEvent` etc.? 🤔

pkosiec (commented):
Nice work! Could you share internally a short demo how it works E2E? not sure how to test it properly.

Thanks!

packages/appkit/src/core/appkit.ts
```diff
@@ -220,6 +230,20 @@ export class AppKit<TPlugins extends InputPluginMap> {
...
+    const reporter = TelemetryReporter.initialize({
+      workspaceHost:
+        serviceCtx.client.config?.host || process.env.DATABRICKS_HOST || "",
+      workspaceId: serviceCtx.workspaceId,
+      client: serviceCtx.client,
+      appId: process.env.DATABRICKS_CLIENT_ID || "",
+      appkitVersion: productVersion,
+    });
+    reporter.start();
+    reporter.sendStartup().catch(() => {});
```
MarioCadenas, line 263: we are calling this (which is asynchronous) and basically ignoring the promise, is this on purpose?

docs/docs/internal-telemetry.mdx
calvarjorge: The page is not loading for me

packages/appkit/src/index.ts
calvarjorge: These were needed so we they could be used in the `dev-playground` page and have full observability to test these requests. I've removed the `dev-playground` page so the exports are no longer needed.

apps/dev-playground/server/internal-telemetry-debug-plugin.ts
calvarjorge: Removed this and the `dev-playground` page for internal telemetry

packages/appkit/src/internal-telemetry/appkit-log.ts
calvarjorge: This is the standard naming for this type of telemetry, used across many teams - Observability Log, Databricks CLI log, etc.

docs/docs/internal-telemetry.mdx
calvarjorge: Added `DO_NOT_TRACK=1` as a way to disable it nevertheless

packages/appkit/src/internal-telemetry/index.ts
calvarjorge: Fixed

docs/docs/internal-telemetry.mdx
calvarjorge: Makes sense, done

packages/appkit/src/internal-telemetry/reporter.ts
calvarjorge: Removed. The `placeholder` is part of the event because Lumberjack does not support empty object. So that Appkit is clearner, I removed it from here, and it will show up as `placeholder: null` in the data.

apps/dev-playground/app.yaml
calvarjorge: My deployed `dev-playground` did not start, as it said it was missing this env var. Do you have a different set up that works?

apps/dev-playground/appkit.plugins.json
calvarjorge: Removed

apps/dev-playground/client/src/routes/__root.tsx
calvarjorge: Comments on critical points:
1. True, but we don't make any guarantee about telemetry delivery. We don't want to have heavy error handling for this and of course never cause any trouble to users because of this. If the `flushRequestMetrics` fails, the data will be lost. We don't want to keep the data (potentially increasing a lot in size) hoping the next attempt will succeed.
2. Done
3. I don't think we want a circuit breaker, specially given that we don't retry. If the endpoint is temporarily down, requests will fail during that time (I don't see any benefit in reducing request frequency in that period, in it will increase the reporter complexity)
4. Will add testing
5. N/A anymore

packages/appkit/src/core/appkit.ts
calvarjorge: Yes, we don't want to cause delays or raise errors because of internal telemetry.

packages/appkit/src/core/appkit.ts
calvarjorge: This was the proposal in the design doc. Since we're obtaining some data from customer apps, we wanted to be as transparent as possible about it.

MarioCadenas (approved):

PR #343, opened 2026-05-04
feat(lakebase): add per-user connections via asUser(req)

pkosiec:
## Summary

Adds per-user Lakebase connection pools to the core plugin via the standard `asUser(req)` pattern. Each user gets their own `pg.Pool` authenticated with their Databricks identity, enabling Row-Level Security (RLS) where `current_user` reflects the actual user.

- **Plugin**: Override `asUser(req)` on `LakebasePlugin` to create/reuse per-user pools via `LakebasePoolManager`
- **Pool lifecycle**: Idle connections close automatically (30s); empty pool objects cleaned up every 5min
- **Template**: `POST /todos` uses `asUser(req)` so `created_by` reflects the user; `postgres` scope added to DABs config; better error messages for missing roles
- **Dev playground**: OBO routes demonstrate RLS with side-by-side SP vs user-scoped views
- **Docs**: OBO section in lakebase.md, per-user connections note in execution-context.md, CLI commands for role management

### Usage

```ts
// Service principal (default)
await AppKit.lakebase.query("SELECT * FROM orders");

// Per-user pool (RLS enforced)
await AppKit.lakebase.asUser(req).query("SELECT * FROM orders");
```

### Docs: "Why deploy first?" improvements (incorporates #365)

This PR incorporates the local dev permissions improvements from #365: **Merged recovery guidance into the "Why deploy first?" admonition** — explains what happens if you run `npm run dev` before deploying (credentials own the schema → deployed app hits `permission denied`) and the drop+redeploy recovery path.

Since this PR touches the same `lakebase.md` section, #365 can be closed once this merges.

## Demo

https://github.com/user-attachments/assets/710a011a-5656-4373-951b-56017ae2badb

## Screenshots 

(better error)
Screenshot 2026-05-06 at 11 48 12

Screenshot 2026-05-06 at 11 43 38

## Test plan

- [x] `pnpm build` passes
- [x] `pnpm -r typecheck` passes
- [x] `pnpm check:fix` passes
- [x] `pnpm test` — all tests pass
- [x] `pnpm docs:build` passes
- [x] Manual: deploy dev-playground, verify SP vs OBO panels
- [x] Manual: scaffold app from template, verify `postgres` scope in `databricks.yml`


MarioCadenas:
## Xavier agentic review — verdict: request changes

Ran a three-persona panel (correctness, security, performance) against the diff. All three voted **request changes**. Summary below; happy to break any item into its own thread if useful.

> Note: also flagged `CONFLICTING` on GitHub — needs a rebase regardless.

### Top blockers (flagged by 2+ reviewers — highest confidence)

1. **TOCTOU race in `LakebasePoolManager.getPool`** — `packages/appkit/src/connectors/lakebase/pool-manager.ts` (~L75-81). `has → get → create → set` is non-atomic; two concurrent first-hit requests for the same key both create a pool, the second `set` orphans the first (leaked sockets, timers, DB connections). Flagged by all 3 reviewers. Fix with an in-flight `Promise` map per key, or compare-and-swap.

2. **Per-user pool keyed only on email — ignores token rotation** — `pool-manager.ts` (~L75-81) + `plugins/lakebase/lakebase.ts` (~L183-192). First request pins the pool with its `WorkspaceClient`/OBO token; later requests for the same email with a fresh `x-forwarded-access-token` keep reusing the stale credential. Breaks per-session OBO guarantees (CWE-384 / CWE-639). Include a token fingerprint in the cache key, or compare-and-recreate when it changes.

3. **`local-dev-user` collapse in development** — `plugins/lakebase/lakebase.ts` (~L172-180). When `x-forwarded-email` is absent but a token is forwarded, `effectiveUser` becomes the literal `"local-dev-user"` for everyone — first caller's pool/identity is reused for all subsequent users. Cross-user impersonation if `NODE_ENV=development` ever leaks to a shared host (CWE-284). Require a verified disambiguator (token subject, x-forwarded-user id) in all OBO paths.

4. **Template `todo-routes.ts` ships an insecure mixed-identity sample** — only `POST /todos` uses `asUser(req)`; `GET`/`PATCH`/`DELETE` still use the service principal and bypass RLS entirely (CWE-285). Every scaffolded app starts with cross-user read/update/delete by default — the template is the place where this story has to be safe out of the box.

### Other highs

- **No identity binding between forwarded email and OBO token subject** (CWE-290) — `plugins/lakebase/lakebase.ts` (~L158-192). Postgres role + cache key derived from an unverified header. Derive identity from validated token claims; treat `x-forwarded-email` as a display hint only.
- **Dev fallback to SP on `NODE_ENV=development`** (CWE-15) — `lakebase.ts` (~L162-165). If `NODE_ENV=development` leaks to a reachable deployment, `asUser(req)` silently falls back to the service principal. Gate on an explicit opt-in env flag rather than `NODE_ENV` alone.
- **Template `GRANT ALL ON ALL TABLES … TO PUBLIC`** (CWE-732) — `template/server/routes/lakebase/todo-routes.ts` (~L35-38). Ships least-privilege violation as the default — undermines the RLS story this PR is selling. Grant to a specific app role only.
- **Unbounded per-user pool map** — `pool-manager.ts` (~L59-81). N pools × `OBO_POOL_DEFAULTS.max=3` → RAM / FD / DB connection pressure at scale. Add an LRU cap or TTL eviction.
- **`WorkspaceClient` rebuilt on every `asUser(req)` even on pool-cache hit** — `lakebase.ts` (~L185-192). Per-request SDK allocation in a hot path. Branch on cache miss before building `perPoolConfig`, or expose `getOrCreate(key, factory)`.

### Recurring patterns triggered (from prior appkit reviews)

- **`error-message-leakage`** — raw `err.message` from Postgres returned to clients in `template/server/routes/lakebase/todo-routes.ts` (L76, L104, L130, L156) and `apps/dev-playground/server/index.ts` (L133-138, L154-159). Log full error server-side; return a generic message + stable error code (CWE-209).
- **`stale-documentation`** — docs omit the `local-dev-user` pool-sharing behavior, and the operator log warning still says SQL runs as the SP even though the agent tool now routes through `asUser(req)`.

### Mediums and lows (worth fixing, not blockers individually)

- End-user email logged at info level (CWE-359) — `lakebase.ts` (~L194-200).
- Operator log warning contradicts OBO agent-tool behavior — `lakebase.ts` (~L311-314).
- Docs omit `local-dev-user` pool-sharing behavior — `docs/docs/plugins/lakebase.md` (~L1296-1299).
- Template PK switch (SERIAL → UUID) without migration story for existing demo DBs.
- `idleTimeoutMillis: 30_000` + `max: 3` + `allowExitOnIdle: true` → connect/auth churn for users with inter-request spacing >30s — `lakebase.ts` (~L30-35).
- O(N) cleanup scan every 5 min — `pool-manager.ts` (~L64-71). Pairs badly with unbounded map.
- Wrong error type for missing email header (uses `AuthenticationError.missingToken(...)` even when a token is present) — `lakebase.ts` (~L172-175).
- Sequential `refetchMy(); refetchAll()` after create — `apps/dev-playground/client/src/components/lakebase/OboProductsPanel.tsx` (~L113-126). Use `Promise.all`.
- `exports()` builds new `WorkspaceClient` per call — `lakebase.ts` (~L222-248).
- New `postgres` OAuth scope is scope-appropriate; would benefit from a brief least-privilege call-out in the docs.

### Why this is "request changes" rather than "approve"

Core architecture (per-user `pg.Pool` via `asUser(req)`) is sound and matches the existing AppKit OBO pattern. But the implementation has a real race in pool creation, doesn't track token rotation, has a dev-mode identity-collapse bug, and the template — which is what users copy — ships with a mixed-SP/OBO data plane and a `GRANT … TO PUBLIC`. The four cross-reviewer blockers plus the four remaining `[high]` findings should land before approval.

---

<sub>Generated by [Xavier](https://github.com/pkosiec/xavier) Shark-pattern review (correctness + security + performance personas, run concurrently against the full diff). Not a substitute for a human reviewer — just a fast first pass.</sub>

apps/dev-playground/app.yaml
```diff
@@ -5,8 +5,6 @@ env:
     valueFrom: genie-space
   - name: DATABRICKS_SERVING_ENDPOINT_NAME
     valueFrom: serving-endpoint
```
MarioCadenas, line 7: was this on purpose? (removing the jobs env var)

pkosiec: oops, no, sorry, reverting

MarioCadenas (approved):

PR #347, opened 2026-05-05
feat: add source-map support for editor DOM mapping

fjakobs:
## Summary
- Added `vite-plugin-react-source-loc.ts` — a Vite plugin that injects `data-source="<file>:<line>:<col>"` attributes onto native JSX elements, enabling the editor to map DOM elements back to source locations
- Enabled `sourcemap: true` in the Vite build config
- Added `@babel/parser`, `@babel/traverse`, and `magic-string` as devDependencies

Backported from the app-builder template (`universe/apps/app-builder/standalone/packages/app-under-dev`).

## Test plan
- [ ] Verify `databricks apps init` produces a working template
- [ ] Verify built client output includes sourcemaps
- [ ] Verify `data-source` attributes appear on native DOM elements in dev mode

## dev mode

image

## prod mode

image


MarioCadenas:
general question, should we add this only on dev and not production builds?

template/client/vite.config.ts
pkosiec, line 1: Agentic review:

### P1 — Should Fix

#### 1. Production sourcemaps expose application source code
**File:** `template/client/vite.config.ts:19` (after PR changes)

The PR adds `sourcemap: true` to the template build config. This generates source maps and includes `//# sourceMappingURL` references in production bundles, making full application source code accessible via browser DevTools.

**Suggested fix:** Use `sourcemap: 'hidden'` instead. This generates `.map` files (useful for error monitoring tools like Sentry) but omits the `sourceMappingURL` comment so browsers won't automatically load them. Alternatively, gate behind an environment variable if source maps should only exist in non-production builds.

#### 2. Client directory list duplication creates drift risk
**Files:** `react-source-loc-vite-plugin.ts:18` vs `vite-dev-server.ts:126`

The plugin's `NESTED_CLIENT_DIRS` constant (`Set(["client", "src", "app", "frontend"])`) duplicates knowledge from `ViteDevServer.findClientRoot()` candidates (`["client", "src", "app", "frontend", "."]`). If either list changes, the other may silently produce wrong relative paths.

**Suggested fix:** Either:
- Extract the candidate dirs to a shared constant imported by both modules, or
- Have `reactSourceLocPlugin()` accept an optional `projectRoot` parameter that `ViteDevServer` passes when it already knows the resolved root (simplest, avoids re-derivation entirely)

---

### P2 — Fix if Straightforward

#### 3. Missing public export (inconsistency with other Vite plugins)
**File:** `packages/appkit/src/index.ts`

Both `appKitTypesPlugin` and `appKitServingTypesPlugin` are exported from `index.ts` (lines 116-118) for users who customize their Vite config. `reactSourceLocPlugin` is not exported. Users with custom Vite setups (e.g., the template itself) cannot opt into this feature.

**Suggested fix:** Add `export { reactSourceLocPlugin } from "./plugins/server/react-source-loc-vite-plugin";` to `index.ts`. If it's intentionally internal-only, document that decision with a comment.

#### 4. PR description mentions wrong dependencies
**File:** PR body

The description states *"Added `@babel/parser`, `@babel/traverse`, and `magic-string` as devDependencies"* but:
- The implementation uses `@ast-grep/napi` (already a dependency), not Babel
- `magic-string` is added as a regular `dependency`, not a `devDependency`

**Suggested fix:** Update the PR description to accurately reflect the dependencies.

---

### P3 — Low / Advisory

#### 5. Consider passing `projectRoot` instead of re-deriving
**File:** `react-source-loc-vite-plugin.ts:19-25`

`resolveProjectRoot()` re-derives the project root from `config.root` using hardcoded heuristics. Since `ViteDevServer` already knows the resolved client root, passing it as a parameter would be simpler and eliminate the duplication from finding #2.

```typescript
// Instead of: reactSourceLocPlugin()
// Could be:   reactSourceLocPlugin({ projectRoot: path.resolve(clientRoot, "..") })
```

#### 6. `cleanModuleId` called redundantly in `transform`
**File:** `react-source-loc-vite-plugin.ts:79,82`

`shouldTransform(id)` calls `cleanModuleId(id)` internally (line 35), then `transform` calls it again (line 82). Extract once at the top of `transform`:

```typescript
transform(code, id) {
  const cleanId = cleanModuleId(id);
  if (!shouldTransform(cleanId, id)) return;  // pass both if needed
  // ...use cleanId below
}
```

#### 7. Test helper uses double cast
**File:** `react-source-loc-vite-plugin.test.ts:31`

`reactSourceLocPlugin() as unknown as TestableHooks` is a double-cast that will silently pass if the plugin API changes shape. Consider narrowing the `TestableHooks` type to match Vite's `Plugin` type more closely, or add a brief comment explaining why the cast is needed.

---

## What Looks Good

- **`apply: "serve"` is correct** — `data-source` attributes are injected only in dev mode, not in production builds
- **`@ast-grep/napi` reuse** — leverages an existing dependency instead of adding Babel
- **`hasDataSourceAttribute` guard** — prevents double-injection when attributes already exist
- **Test coverage** — covers native tags, components, member expressions, namespaced tags, existing attributes, and flat project root layouts
- **Source map generation in `transform` return** — correctly returns `map: s.generateMap({ hires: true })` for accurate dev sourcemaps

---

## Verification

After implementing the suggested changes:
1. `pnpm build && pnpm test` — ensure all tests pass
2. `pnpm dev` — verify `data-source` attributes appear on native DOM elements in browser DevTools
3. `pnpm --filter=dev-playground build` — verify production build with `sourcemap: 'hidden'` generates `.map` files without `sourceMappingURL` in bundles
4. `pnpm check:fix && pnpm -r typecheck` — verify lint and type checks pass

template/client/vite.config.ts
```diff
@@ -6,16 +6,14 @@ import path from 'node:path';
...
-    tailwindcss(),
-  ],
+  plugins: [react(), tailwindcss()],
   server: {
     middlewareMode: true,
   },
   build: {
     outDir: path.resolve(__dirname, './dist'),
     emptyOutDir: true,
+    sourcemap: true,
```
pkosiec: Are we sure we want to enable sourcemap for new apps even for production apps? 
Maybe we should do something like

```ts
    sourcemap: process.env.NODE_ENV === 'production' ? 'hidden' : true,
```
?

pkosiec (approved):
LGTM

PR #349, opened 2026-05-06
feat(appkit): scan upward from preferred port in development

MarioCadenas:
## Summary

- In `NODE_ENV=development`, the server plugin now picks the first free TCP port by scanning upward from the preferred port (`server({ port })`, `DATABRICKS_APP_PORT`, or default `8000`) using `get-port`'s `portNumbers(from, to)` helper, bounded to 100 attempts.
- Avoids `EADDRINUSE` failures when running multiple AppKit dev servers locally (e.g. two playgrounds at once).
- Production / non-development behavior is unchanged: the plugin binds the configured / env / default port and lets `listen` errors surface as before.
- Startup log already prints the bound URL via the existing `Server running on http://…` line; no extra fallback log is emitted.

## Implementation

- New private `ServerPlugin.resolveListenPort()` returns:
  - In development: `await getPort({ host, port: portNumbers(requested, requested + 99) })`.
  - Otherwise: the requested port (no scan).
- `start()` awaits the resolved port before `app.listen(...)`.
- `logStartupInfo()` reads the resolved port so logs match the actual socket.
- Adds `get-port@7.2.0` to `@databricks/appkit` dependencies.

## Test plan

- [x] `pnpm --filter @databricks/appkit typecheck`
- [x] `pnpm exec vitest run packages/appkit/src/plugins/server/tests/server.test.ts packages/appkit/src/plugins/server/tests/server.integration.test.ts` (37 tests pass; new dev/prod cases cover the resolution branches via a partial `get-port` mock)
- [x] `pnpm exec biome check --write` on touched files
- [ ] Manual: start two dev playgrounds back-to-back; confirm second one binds `8001` (or next free port) and the URL line reflects it

### DIFF
--- a/packages/appkit/src/plugins/server/index.ts
+++ b/packages/appkit/src/plugins/server/index.ts
@@ -3,6 +3,7 @@ import type { Server as HTTPServer } from "node:http";
 import path from "node:path";
 import dotenv from "dotenv";
 import express from "express";
+import getPort, { portNumbers } from "get-port";
 import type { PluginClientConfigs, PluginPhase } from "shared";
 import { ServerError } from "../../errors";
 import { createLogger } from "../../logging/logger";
@@ -20,6 +21,9 @@ import { ViteDevServer } from "./vite-dev-server";
 dotenv.config({ path: path.resolve(process.cwd(), "./.env") });
 
 const logger = createLogger("server");
+
+/** Dev-only: try `requested` then consecutive ports (see `get-port` `portNumbers`). */
+const devListenPortSpan = 100;
 
 /**
  * Server plugin for the AppKit.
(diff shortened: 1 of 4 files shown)


packages/appkit/src/plugins/server/index.ts
```diff
@@ -306,10 +314,36 @@ export class ServerPlugin extends Plugin {
...
+    }
+
+    const host = this.config.host ?? ServerPlugin.DEFAULT_CONFIG.host;
+    const upper = Math.min(requested + devListenPortSpan - 1, 65_535);
+    const port = await getPort({
+      host,
+      port: portNumbers(requested, upper),
+    });
+    this.resolvedListenPort = port;
+    return port;
```
atilafassina, line 340: I think here is a good opportunity to log and info `${PORT} was busy, picking ${this.resolvedListenPort}`

atilafassina (approved):

packages/appkit/src/plugins/server/index.ts
MarioCadenas: will do!

MarioCadenas: Implemented in 0e2913d8 — `resolveListenPort()` now emits `logger.info("Port %d was busy, picking %d", requested, port)` only when the bound port differs from the preferred one. Two unit tests cover the busy and free cases.

PR #350, opened 2026-05-06
ci: detect breaking-change commits in PRs

MarioCadenas:
## Summary

Adds a breaking-change detector to the existing PR metadata workflow. Every commit in a PR is scanned for Conventional Commits breaking-change markers, and the merge is gated.

- Renames `pr-title.yml` → `pr-metadata.yml` (workflow name `PR Metadata Verification`) and adds `detect-breaking` as a second job alongside the existing title check.
- Walks `git rev-list base..head` over the paths tracked by `.release-it.json` (`packages/appkit`, `packages/appkit-ui`, `packages/shared`) so docs/tooling-only commits don't trigger noise.
- Detection lives in `tools/detect-breaking-commits.ts` (TS, run with `node` via Node 24's native type stripping — no extra install/transpile step), consistent with the rest of `tools/*.ts`.
- Matches both subject markers (`feat!:`, `feat(scope)!:`, etc. across all conventional types) and `BREAKING CHANGE:` / `BREAKING-CHANGE:` footers.
- Upserts a sticky PR comment via `actions/github-script@v7.0.1` (identified by an HTML-comment marker so re-runs update/remove the same comment).
- Fails the job unless the PR carries an `allow-breaking-change` label, providing an explicit escape hatch for intentional major bumps.

Today, `.release-it.json` has `bumpStrict: true`, so a single `BREAKING CHANGE:` or `feat!` commit landing on `main` silently forces a major version on the next release. This check makes that visible at PR time.

## Follow-ups (not in this PR)

- Update `main` branch protection required status checks: the previous `PR Title / Conventional Commit Title` check is now `PR Metadata Verification / Conventional Commit Title`, and add `PR Metadata Verification / Detect Breaking Commits` so the merge button is actually blocked.
- Create the `allow-breaking-change` label so reviewers can opt in when a major bump is intentional.
- If `main` is configured for **squash merge**, the per-commit messages are discarded on merge and `release-it` only sees the squash commit (already covered by the title check); in that case this check is informational. With **merge commit** or **rebase merge** strategies, the check is load-bearing.

## Test plan

- [ ] On the PR for this branch, confirm both jobs run under `PR Metadata Verification` and the breaking-change job reports "no breaking commits found" (no comment posted, job green).
- [ ] In a follow-up scratch PR, push a commit like `feat(appkit)!: test breaking marker` touching `packages/appkit/**` and verify:
  - [ ] Sticky comment appears listing the SHA + subject.
  - [ ] Job fails red.
  - [ ] Adding the `allow-breaking-change` label re-runs the workflow and turns it green; comment updates to acknowledge the intentional bump.
  - [ ] Removing the label flips it back to red.
- [ ] Push a commit with `feat!:` touching `docs/**` only and confirm the check stays green (path filter working).
- [ ] Amend the breaking commit away and verify the sticky comment is deleted on the next workflow run.

image


atilafassina (approved):

.github/workflows/pr-metadata.yml
pkosiec, line 1: Maybe we can keep it within the same workflow (but as a separate job) as the PR title check?

Let's call it PR metadata verification or similar 🤔

.github/scripts/detect-breaking-commits.sh
pkosiec, line 1: consider using ts/js for that, to be consistent with the tools 🤔  

But primarily, do we need that? We squash those commits so maybe we should enforce e.g. empty PR description?

Image

MarioCadenas: we could also do that, that wouldn't cover the case for `feat!: ...` is the only thing

pkosiec: gotcha, so the script is useful - do you mind to unify it with other tools to use ts?

MarioCadenas: totally that's a valid point

pkosiec (approved):

PR #365, opened 2026-05-07
docs: update Lakebase local dev permissions guidance

philip:
Scope databricks_superuser grant to collaborators only (project creators already have access). Merge recovery guidance into the "Why deploy first?" admonition with a drop+redeploy path.


pkosiec:
Thank you @philip for your contribution! I merged the other PR (on agent skills side) but I'd close this one, if you don't mind. The reason is I already reworked that section to use Databricks CLI for role management so. But I incorporated your suggested changes for the "Why deploy first?" section 👍 Thanks again for your help!

PR #366, opened 2026-05-08
feat(appkit): agent chat UI primitives

hubertzub-db:
## Agent chat UI primitives

Adds a small set of headless React building blocks for chatting with an AppKit `agents()` server, and rewires the dev playground's agent route to use them.

### What's new

A new `@databricks/appkit-ui/react/chat` subpath exporting:

- `<Conversation>` — render-prop component that runs the whole chat (streaming, history, scroll-stick).
- `<ChatInput>` — render-prop input that handles Enter / Shift+Enter / IME / streaming-state.
- `useChat` and `useScrollToBottom` — the underlying hooks for custom layouts.
- `ResponsesApiTransport` — the bit that talks to the agents plugin.

The smallest working chat is now ~30 lines of JSX, and threads, tool calls, reasoning ("thinking…"), and destructive-tool approval prompts all stream through without any server-side glue.

Docs at `docs/docs/api/appkit-ui/chat/index.md`.

### Why

Until now, anyone building a chat UI on top of the `agents` plugin had to hand-roll SSE parsing, message reducers, scroll behavior, thread-id plumbing, etc. — the dev playground's agent route was 500+ lines of exactly that. Now it's the same UI, but built out of reusable parts, and any consumer gets the same thing for free.

We're piggy-backing on the Vercel AI SDK (`@ai-sdk/react`, `ai`) for the reducer / state machine — they're declared as optional peer dependencies, so apps that don't use the chat primitives don't pull them in.

### Demo (playground still works)

https://github.com/user-attachments/assets/afdcdd11-dcd8-45b7-a7e4-f2081814737b

### Another demo (tool approval)

approval-minimal

### What changed in the playground

`apps/dev-playground/client/src/routes/agent.route.tsx` was rewritten to use the new primitives. Behavior is the same (chat bubbles, tool approvals, autocomplete, event-stream debug panel), just driven by the shared components instead of inline SSE handling.

### Notes for reviewers

- Tests live next to the transport (`responses-api-transport.test.ts`) and cover the wire-format translation end-to-end.
- The agents plugin's wire format is unchanged — this is a pure client-side addition.
- Only text user messages are supported for now (the agents plugin doesn't accept attachments yet).

### DIFF
--- a/packages/appkit-ui/package.json
+++ b/packages/appkit-ui/package.json
@@ -39,6 +39,10 @@
       "development": "./src/react/beta.ts",
       "default": "./dist/react/beta.js"
     },
+    "./react/chat": {
+      "development": "./src/react/chat/index.ts",
+      "default": "./dist/react/chat/index.js"
+    },
     "./package.json": "./package.json",
     "./styles.css": {
       "development": "./src/react/styles/globals.css",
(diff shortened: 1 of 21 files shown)


packages/appkit-ui/package.json
```diff
@@ -103,13 +107,25 @@
     "vaul": "1.1.2"
   },
   "peerDependencies": {
+    "@ai-sdk/react": "~4.0.0-beta",
```
hubertzub-db, line 110: note: we're importing canary channel since 4.0+ allows for customizing messages transport easily, should be published as stable very soon

PR #371, opened 2026-05-11
refactor(appkit): introduce CoreServiceRegistry and ServiceLocator

ditadi:
## Stack

- **(this PR)** — `CoreServiceRegistry` + `ServiceLocator` (layer 1, base: `main`)
- #372 — split `SQLWarehouseConnector` into submit/get/poll/transform (layer 2, base: this PR)

Review independently — this PR is groundwork, no behaviour change.

---

AppKit's core previously ran a hardcoded boot sequence for cache +
telemetry. This refactor introduces a generic registry so the core no
longer names concrete services:

- `CoreServiceFactory<T>` — interface every core service implements,
  returning `{ instance, shutdown }` (or `null` for intentional opt-outs).
- `CoreServiceRegistry` — boots a list of factories in declared order,
  shuts down in reverse, doubles as a typed locator via `get<T>(name)`.
- `bootstrap-factories.ts` — composes the list `_createApp` boots,
  living next to the concrete service modules so `core/appkit.ts` stays
  free of concrete-class imports.

Cache and Telemetry are migrated to expose `.factory()` static methods
and flow through the registry. Behaviour is identical: same boot order,
same shutdown order, same opt-out semantics.

Plugins now receive a typed `services: ServiceLocator` through
`attachContext`, in addition to the existing per-service fields kept
for backwards compatibility (`cache`, `telemetry`, plus a `taskflow`
placeholder typed as `unknown` for forward compat). New plugins should
consume `services.get<T>(name)`; the named legacy fields are marked
`@deprecated`.

This PR is groundwork — no behaviour change, no public-API removal.
The purpose is to make room for a third core service (TaskFlow) without
re-threading the cross-package `attachContext` contract: adding a new
service now only touches the service module and `bootstrap-factories.ts`.
TaskFlow itself is intentionally NOT in this PR; it lands in a follow-up.

Verified: pnpm -r typecheck, pnpm build, full pnpm test
(122 files, 2255 tests) all green.

Signed-off-by: ditadi <victordperd@gmail.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/appkit/src/cache/index.ts
+++ b/packages/appkit/src/cache/index.ts
@@ -1,6 +1,7 @@
 import { createHash } from "node:crypto";
 import { ApiError, WorkspaceClient } from "@databricks/sdk-experimental";
 import type { CacheConfig, CacheStorage } from "shared";
+import type { CoreServiceFactory } from "@/core/service-registry";
 import { createLakebasePool } from "../connectors/lakebase";
 import { AppKitError, ExecutionError, InitializationError } from "../errors";
 import { createLogger } from "../logging/logger";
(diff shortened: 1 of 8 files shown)


ditadi:
Superseded by #374 — replatformed onto `git stack` with branch name `stack/taskflow/service-registry`.

PR #385, opened 2026-05-14
refactor(appkit): own asUser OBO wrapping in Plugin proxy

MarioCadenas:
Move the OBO-side exports() wrapping out of AppKit#wrapWithAsUser and into Plugin.asUser's proxy. The proxy now intercepts reads of `exports` and returns the underlying exports() result with every function pre-wrapped in the user-context scope. The cross-file USER_CONTEXT_SYMBOL bridge between plugin.ts and appkit.ts is removed, and wrapExportsInUserContext is deleted from appkit.ts.

The two inline Proxy literals in Plugin.asUser (real OBO and dev-mode fallback) collapse into a single _createAsUserProxy(wrapCall) factory that takes a per-call wrap strategy. The real OBO branch uses runInUserContext(userContext, ...); the dev fallback uses otelContext.with(DEV_OBO_FALLBACK_KEY=true, ...).

Behavior preserving: full appkit test suite passes and typecheck is clean. Adds a focused 26-test file covering proxy mechanics: real OBO method calls, exports() interception (class methods, arrows, nested plain objects, class-instance values not recursed, callable exports, null/undefined), AsyncLocalStorage propagation across Promise.all, concurrent user isolation, error cleanup, dev-fallback exports(), and escape-the-proxy boundaries (returned functions not auto-wrapped, `return this` returns the unwrapped target).

Also drop the now-unused UserContext type re-export from the internal context barrel (not part of the public package entry).

### DIFF
--- a/packages/appkit/src/plugin/plugin.ts
+++ b/packages/appkit/src/plugin/plugin.ts
@@ -14,12 +14,7 @@ import type {
 } from "shared";
 import { AppManager } from "../app";
 import { CacheManager } from "../cache";
-import {
-  getCurrentUserId,
-  runInUserContext,
-  ServiceContext,
-  type UserContext,
-} from "../context";
+import { getCurrentUserId, runInUserContext, ServiceContext } from "../context";
 import type { PluginContext } from "../core/plugin-context";
 import { AppKitError, AuthenticationError } from "../errors";
 import { createLogger } from "../logging/logger";
@@ -44,17 +39,44 @@ import type {
 const logger = createLogger("plugin");
 
 /**
- * Symbol used to expose the UserContext from an asUser() proxy.
- * Allows wrapWithAsUser in appkit.ts to retrieve the context and
- * wrap export methods in runInUserContext().
- */
-export const USER_CONTEXT_SYMBOL = Symbol("appkit.userContext");
-
-/**
  * OTel context key for marking OBO dev mode fallback.
  * Set when asUser() is called in development mode without a user token.
  */
 const DEV_OBO_FALLBACK_KEY = createContextKey("appkit.devOboFallback");
+
+/**
+ * Returns true if `value` is a plain object literal (not an array, Date,
+ * class instance, etc.). Used to decide whether to recurse into nested
+ * export shapes when wrapping functions.
+ */
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  if (typeof value !== "object" || value === null) return false;
+  const proto = Object.getPrototypeOf(value);
+  return proto === Object.prototype || proto === null;
+}
+
+/**
+ * Recursively replaces every function in `exports` with `wrap(fn)`,
+ * walking into nested plain objects. Mutates and returns `exports`.
+ *
+ * Used by the asUser proxy to make the user context follow function
+ * references that escape the proxy via `exports()`.
+ */
+function wrapExportFunctions(
+  exports: Record<string, unknown>,
+  wrap: (fn: (...a: unknown[]) => unknown) => (...a: unknown[]) => unknown,
+): Record<string, unknown> {
+  for (const key of Object.keys(exports)) {
+    if (!Object.hasOwn(exports, key)) continue;
+    const val = exports[key];
+    if (typeof val === "function") {
+      exports[key] = wrap(val as (...a: unknown[]) => unknown);
+    } else if (isPlainObject(val)) {
+      wrapExportFunctions(val as Record<string, unknown>, wrap);
+    }
+  }
+  return exports;
+}
 
 /**
  * Returns true if the current execution is an OBO dev mode fallback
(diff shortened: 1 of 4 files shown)


packages/appkit/src/plugin/plugin.ts
pkosiec, line 1: # Code Review: PR #385 — refactor(appkit): own asUser OBO wrapping in Plugin proxy

## Context

PR #385 moves OBO-side `exports()` wrapping out of `AppKit#wrapWithAsUser` and into `Plugin.asUser`'s proxy. The cross-file `USER_CONTEXT_SYMBOL` bridge is removed, and two inline Proxy literals (real OBO + dev fallback) collapse into a single `_createAsUserProxy(wrapCall)` factory. A 26-test file covering proxy mechanics is added.

**Scope:** 4 files changed in `packages/appkit/` — context/index.ts, core/appkit.ts, plugin/plugin.ts, plugin/tests/asUser-proxy.test.ts (new)
**Reviewers dispatched:** correctness, testing, maintainability, project-standards, agent-native, learnings-researcher, kieran-typescript, reliability, api-contract

---

## Review Findings

### P2 — Moderate

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 1 | `packages/appkit/src/plugin/plugin.ts:52` | **Duplicate `isPlainObject` implementation.** Identical logic exists at `plugin.ts:52` (module function) and `appkit.ts:180` (static method). Two sources of truth for the same check — if logic needs to change, both must be updated independently. | maintainability | 0.95 | `manual -> human` |
| 2 | `packages/appkit/src/plugin/plugin.ts:65` | **`wrapExportFunctions` mutates input object in-place.** Currently safe because `exports()` returns a fresh object per call, but fragile if a plugin ever caches the result. The docstring says "Mutates and returns" but the `wrap` naming suggests a non-mutating operation. | maintainability | 0.80 | `advisory -> human` |
| 3 | `packages/appkit/src/plugin/plugin.ts:75` | **No circular reference protection in `wrapExportFunctions`.** Recursive descent into nested plain objects has no cycle detection. Extremely unlikely with real plugin exports (shallow structures, `isPlainObject` rejects arrays/class instances), but undocumented limitation. | testing | 0.70 | `advisory -> human` |

### P3 — Low

| # | File | Issue | Reviewer(s) | Confidence | Route |
|---|------|-------|-------------|------------|-------|
| 4 | `packages/appkit/src/plugin/plugin.ts:70` | **Redundant `Object.hasOwn()` check.** `Object.keys()` on line 69 already returns only own enumerable keys, so the `hasOwn` guard on line 70 can never be true — dead code. | testing | 0.95 | `safe_auto -> review-fixer` |

### Pre-existing (not introduced by this PR)

| # | File | Issue | Reviewer(s) | Confidence |
|---|------|-------|-------------|------------|
| — | `packages/appkit/src/core/appkit.ts:172` | `(plugin as any).asUser(req).exports()` cast loses type safety. Pre-existing pattern — the old code also used `(plugin as any)` with `USER_CONTEXT_SYMBOL`. | kieran-typescript | 0.92 |

---

## Testing Gaps

- No test for `exports()` returning non-plain, non-function values (Map, Set, Array as top-level return) — the fallthrough at `plugin.ts:502` is uncovered
- No test for deeply nested exports (5+ levels) to confirm recursion terminates cleanly
- No integration test at AppKit layer for when `plugin.exports()` throws during `wrapWithAsUser`
- `wrapExportFunctions` and `isPlainObject` have no isolated unit tests (only tested indirectly through proxy integration)

---

## Coverage Notes

- **Suppressed:** 0 findings below confidence threshold
- **Learnings:** No `docs/solutions/` directory exists — no institutional learnings to surface
- **Agent-Native:** No gaps — this is an internal refactor with no new user-facing features
- **Schema Drift / Deployment:** N/A (no migrations)
- **Failed reviewers:** 0 of 9

---

## Verdict: Ready to merge (with optional improvements)

The core refactoring is sound — consolidating two Proxy literals into `_createAsUserProxy(wrapCall)` is a clear improvement that eliminates the cross-file `USER_CONTEXT_SYMBOL` bridge. The 26-test file is thorough and well-organized, covering real OBO, dev fallback, exports interception, async propagation, concurrent user isolation, and boundary cases.

No P0 or P1 issues found in the PR's committed changes. The findings below are improvements, not blockers.

---

## Improvement Plan

### Fix 1: Remove redundant `Object.hasOwn()` check (P3)

**File:** `packages/appkit/src/plugin/plugin.ts:70`

Remove the dead `if (!Object.hasOwn(exports, key)) continue;` guard since `Object.keys()` on line 69 already returns only own enumerable keys.

### Optional improvements (not blocking merge)

2. **Deduplicate `isPlainObject`** — Extract to a shared utility (e.g. `packages/appkit/src/utils.ts`) or have `plugin.ts` import `AppKit.isPlainObject` from `appkit.ts`. Both files (`plugin.ts:52` and `appkit.ts:180`) have identical implementations.

3. **Add test for non-plain-object top-level exports** — Add a test case where `exports()` returns a Map or Array as the top-level value, verifying it's returned as-is without wrapping (the `plugin.ts:502` fallthrough path).

4. **Add test for mixed async/sync exports** — Verify that an exports object with both async and sync functions preserves their semantics when wrapped.

---

## Verification

After making changes:
```bash
pnpm build && pnpm test && pnpm check:fix && pnpm -r typecheck
```

pkosiec (approved):
Nice one! Tested Lakebase OBO - still works 👍 

Before merge, please take a look at the agentic review summary.

PR #386, opened 2026-05-14
feat(appkit): one chat integration

MarioCadenas:
## Why

The OneChat platform integrates with AppKit agent apps by calling **`POST <app-url>/responses`** at the root level, sending the OpenAI Responses API request shape, and expecting a single JSON envelope back (not SSE). Until this PR, AppKit only exposed `POST /invocations` (SSE), so OneChat got `404 Cannot POST /responses` from Express and had no compatible single-shot endpoint to consume.

This integration doesn't support HITL either so we handle that too.


pkosiec (commented):
Code LGTM - could you please share a short demo or at least a testing instruction to see how it works? Thanks!

atilafassina (approved):

pkosiec (approved):

PR #390, opened 2026-05-19
fix(appkit): backpressure inline-Arrow stash before EXTERNAL_LINKS fallback

jamesbroadhead:
## Summary

Follow-up to #329. On R-backed warehouses, the analytics plugin's stash-full fallback target (`EXTERNAL_LINKS + ARROW_STREAM`) does not exist — R returns HTTP 501 `NOT_IMPLEMENTED: "ExternalLinks disposition is not yet implemented."` — so any stash overflow on a R app would surface that 501 to the caller with no further recovery.

This PR adds brief FIFO backpressure inside the stash so transient capacity pressure is absorbed by waiting for the next `/arrow-result` drain (typically tens of ms — the stash is drain-on-read) rather than immediately escalating to EXTERNAL_LINKS. The existing EXTERNAL_LINKS fallback is preserved for true sustained overload on warehouses that accept it.

## What's the fix

- `InlineArrowStash`: new `putWaitMs` option and `async putBlocking(userId, bytes, signal?)`. On overflow it waits FIFO up to `putWaitMs` for a `take()` / TTL-gc to free a slot, then retries. Honors `AbortSignal` (aborted request drops out of the queue immediately). When the wait elapses without capacity, returns `null` and the caller falls back as before.
- `AnalyticsPlugin`: stash now constructed with `putWaitMs: 500`; the call site `await`s `putBlocking` and passes the request's abort signal.
- Wait-queue is strictly FIFO. `wakeWaiters()` walks from the head and shifts each fitting waiter off *before* invoking it, so a re-entrant `gc()` inside the waker can't pick up the same waiter twice.

Surface area is small: the synchronous `put()` is unchanged, the EXTERNAL_LINKS fallback path is unchanged, and `putWaitMs = 0` (the default) reduces `putBlocking` to the synchronous behavior.

## Why not the alternatives

- **Probe warehouse capability up-front, skip EXTERNAL_LINKS on R**: more state, more code, doesn't help if a different warehouse changes behavior later.
- **Catch the 501 and surface a clearer error**: improves diagnostics, but the request still fails — backpressure prevents the failure.
- **Drop the stash entirely and content-negotiate Arrow bytes inline in the HTTP response**: structurally cleaner, but a wire-protocol shift better done as a separate PR.

## Tests

`inline-arrow-stash.test.ts` (+8 cases):

- [x] `putBlocking` succeeds immediately when capacity is available
- [x] waits for a `take()` to free a slot, then succeeds
- [x] returns `null` when `putWaitMs` elapses without a slot freeing
- [x] preserves FIFO order across waiters
- [x] rejects later waiters when the head consumes the freed capacity
- [x] settles `null` when `AbortSignal` aborts mid-wait
- [x] pre-aborted signal short-circuits (no wait)
- [x] `putWaitMs=0` (default) behaves like the synchronous `put()`

`analytics.test.ts`: existing "stash full → EXTERNAL_LINKS" integration test now stubs `putBlocking` directly instead of `put`, so the suite doesn't bake the 500ms wait into the run time.

Full suite: **2,682 tests, all green** (was 2,674 on the base).

## Stack

- #329 — inline Arrow IPC decode + warehouse-compat fallback (base for this PR)
- **(this PR)** — backpressure the stash so the fallback isn't reached on R under transient pressure

Targets `stack/arrow-3-inline-arrow-fix`; will rebase onto `main` after #329 merges.

## Test plan

Verified live against e2-dogfood `Test R` (warehouse_type=`R`, id `0000000002e475d4`):

- [x] `JSON_ARRAY` caller: single SSE `result` event with row data, no fallback (unchanged)
- [x] `ARROW_STREAM` caller: single SSE `arrow` event → `GET /arrow-result/inline-<uuid>` returns 776 bytes of valid Arrow IPC, `application/vnd.apache.arrow.stream` (unchanged — new code only activates under capacity pressure)

The new code path (backpressure under contention) is covered by the unit tests; reproducing it live requires saturating the stash, which is outside the scope of this PR.

### DIFF
--- a/packages/appkit/src/plugins/analytics/analytics.ts
+++ b/packages/appkit/src/plugins/analytics/analytics.ts
@@ -60,7 +60,15 @@ export class AnalyticsPlugin extends Plugin implements ToolProvider {
    * the bytes through the existing `/arrow-result/:jobId` endpoint with
    * a real binary content-type.
    */
-  protected inlineArrowStash: InlineArrowStash = new InlineArrowStash();
+  // Short put-wait so that a momentarily full stash backpressures rather
+  // than immediately falling back to EXTERNAL_LINKS — important on
+  // warehouses (e.g. Reyden) that refuse EXTERNAL_LINKS outright. The
+  // stash is drain-on-read, so an in-flight `/arrow-result` GET from any
+  // concurrent query usually frees a slot well within this window. True
+  // sustained overload still falls back via the existing path below.
+  protected inlineArrowStash: InlineArrowStash = new InlineArrowStash({
+    putWaitMs: 500,
+  });
 
   constructor(config: IAnalyticsConfig) {
     super(config);
@@ -327,7 +335,8 @@ export class AnalyticsPlugin extends Plugin implements ToolProvider {
    *   `JSON_ARRAY` contract is preserved.
    * - **ARROW_STREAM** first tries `INLINE + ARROW_STREAM`. If the
    *   warehouse refuses (most classic + some serverless variants), or the
-   *   inline stash is full, falls back to `EXTERNAL_LINKS + ARROW_STREAM`.
+   *   inline stash is full after a brief backpressure wait, falls back to
+   *   `EXTERNAL_LINKS + ARROW_STREAM`.
    *
    * INLINE Arrow attachments under the ARROW_STREAM path are decoded once
    * and put on the plugin's `inlineArrowStash`; the SSE message carries the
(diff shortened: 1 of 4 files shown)


jamesbroadhead:
Superseded by the overflow-pool fix landed on #329 in 4afce1f6 (see https://github.com/databricks/appkit/pull/329#issuecomment-4510512570). The `putWait` half-measure here is no longer needed — the overflow pool eliminates the double-execution path entirely.

PR #393, opened 2026-05-19
fix(appkit): use retry client for Lakebase V1 transaction retry

### LINKED ISSUE #392: Lakebase V1 transaction auth retry uses original client for transaction boundaries
## Summary
`LakebaseV1Connector.transaction()` rotates credentials on `28P01`, opens a retry client, and passes that retry client to the transaction callback. The retry path still executes `BEGIN` and `COMMIT` on the original client, which has already been released.

## Impact
The callback work and transaction boundaries can run on different PostgreSQL clients. In practice the retry path may fail immediately because the original client was released, or it may leave the retry callback work outside the intended transaction.

## Expected
After credential rotation, `BEGIN`, callback work, `COMMIT`, and rollback handling should all use the retry client.

## Repro Pointer
See `packages/appkit/src/connectors/lakebase-v1/client.ts` in the auth retry branch of `transaction()`.

fallintoplace:
## Summary
- use the retry PostgreSQL client for retry transaction `BEGIN` and `COMMIT` after credential rotation
- guard the original transaction client release so retry paths do not release it twice
- add a regression test covering auth-failure transaction retry behavior

## Root Cause
The auth retry path opened a retry client and passed it to the transaction callback, but still ran transaction boundaries on the original client after releasing it. That could put callback work and transaction control on different clients, or fail because the original client was already released.

## Validation
- `pnpm exec vitest run packages/appkit/src/connectors/tests/lakebase-v1.test.ts`
- `pnpm exec biome check packages/appkit/src/connectors/lakebase-v1/client.ts packages/appkit/src/connectors/tests/lakebase-v1.test.ts`
- `pnpm --filter @databricks/appkit typecheck`
- `git diff --check`

Closes #392

### DIFF
--- a/packages/appkit/src/connectors/lakebase-v1/client.ts
+++ b/packages/appkit/src/connectors/lakebase-v1/client.ts
@@ -208,6 +208,14 @@ export class LakebaseV1Connector {
       async (span) => {
         const pool = await this.getPool();
         const client = await pool.connect();
+        let clientReleased = false;
+        const releaseClient = () => {
+          if (!clientReleased) {
+            client.release();
+            clientReleased = true;
+          }
+        };
+
         try {
           await client.query("BEGIN");
           const result = await callback(client);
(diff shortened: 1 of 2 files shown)


pkosiec:
Hi @fallintoplace! Thank you so much for taking the time to put this contribution together.

We actually deprecated the Lakebase v1 connector a few months ago and are planning to remove it completely, so I am going to close this PR for now. If you are interested in applying your changes to our current setup, the new connector lives under packages/lakebase.

Thanks again for your effort and for being part of the community!

PR #399, opened 2026-05-21
feat(appkit): allowlist DBSQL error codes whose messages are safe to forward

jamesbroadhead:
## Summary

Follow-up to the error-sanitization work in #329. Restores user-facing SQL error messages (e.g. \`Table 'foo' not found\`, \`Syntax error near ','\`) without re-introducing the CWE-209 leak from raw control-plane wording.

#329 collapses every \`ExecutionError.statementFailed\` to a generic \"Query execution failed\" client message. That was the right safe-default for the sanitization pass, but it loses real user value — anyone with a typo in their SQL now debugs blind, because the error is the user's own SQL artifact and was authored by DBR for them to read.

This PR splits the two error sources by \`ServiceErrorCode\`:

1. **DBR (data plane)** — populates \`errorDisplayMessage\` for user-authored SQL errors. Source of truth: \`sqlgateway/scheduler/src/main/scala/DriverRequests.scala:181-208\` — \`ErrorInfo.withMessage(command.errorDisplayMessage).withErrorReturnedFromDataPlane(true)\`. Codes: \`BAD_REQUEST\`, \`NOT_FOUND\`, \`ALREADY_EXISTS\`, \`DEADLINE_EXCEEDED\`, \`CANCELLED\`, \`UNAUTHENTICATED\`. These messages **are** the user's own SQL — passthrough is correct.

2. **Control plane (proxy / scheduler / metaservice)** — wraps internal RPC failures, scheduler exceptions, capacity rejections. Source: \`sqlgateway/scheduler/src/main/scala/utils/CommandUtils.scala\`, \`exceptions.scala\`. Codes: \`INTERNAL_ERROR\`, \`IO_ERROR\`, \`UNKNOWN\`, \`RESOURCE_EXHAUSTED\`, \`SERVICE_UNDER_MAINTENANCE\`, \`TEMPORARILY_UNAVAILABLE\`, \`WORKSPACE_TEMPORARILY_UNAVAILABLE\`, \`ABORTED\`. These can carry correlation IDs, RPC paths, stack traces — **must not** pass through.

## Design

- New module \`packages/appkit/src/errors/dbsql-error-allowlist.ts\` exports \`isSqlErrorPassthrough(errorCode)\` — a \`ReadonlySet\`-backed predicate over \`ServiceErrorCode\`.
- \`ExecutionError.statementFailed\` consults the allowlist before defaulting \`clientMessage\`. Allowlisted codes get the verbatim warehouse text; anything else (including future SDK additions) collapses to the generic \"Query execution failed\".
- An explicit \`clientMessage\` argument always wins — the allowlist is a default, not a ceiling.
- Server-side \`.message\` always preserves the raw upstream text for \`logger.error\` capture; sanitization is purely at the wire boundary.

## Stack

Stacks on #329. Targets \`stack/arrow-3-inline-arrow-fix\` (not \`main\`) so it merges as the next layer once the parent lands.

## Test plan

- [x] \`describe(\"statementFailed clientMessage passthrough\")\` — 13 new cases covering each allowlisted code, key denied codes, default-deny on unknown / undefined / lowercased input, explicit clientMessage override, and raw-message preservation for log debugging.
- [x] \`dbsql-error-allowlist.test.ts\` — focused tests on the predicate itself (allowlist members, explicit denylist members, undefined / empty / unknown / lowercase).
- [x] Full \`appkit\` suite: 2,129 tests pass (+31 new).
- [x] \`tsc --noEmit\` clean; \`biome check\` clean.


jamesbroadhead:
Pushed `a5c8e4a4` — expanded the allowlist after reviewer pushback on the 5xx-class codes.

**Newly allowlisted** (all verified against universe construction sites):
- `TEMPORARILY_UNAVAILABLE` — stable template *"DBSQL temporarily unavailable. Please try again in a few minutes."* (`SchedulerRpcClient.scala`, `SqlGatewayClerk.scala`)
- `RESOURCE_EXHAUSTED` — explicit user-actionable templates: *"The maximum number of warehouses has been reached. Please contact Databricks support."*, *"Stop or delete existing warehouses to free up capacity."* (`SchedulerRpcValidatorHook.scala`)
- `WORKSPACE_TEMPORARILY_UNAVAILABLE` — shard-routing messages (`proxy/thrift/Exceptions.scala`); names a shard ID, low-sensitivity, retry-relevant
- `SERVICE_UNDER_MAINTENANCE` — maintenance-window templates
- `ABORTED` — short reason strings (`"version mismatch"`)

**Still denied** — these wrap raw `ex.getMessage` or interpolate internal identifiers, so leak risk is concrete:
- `INTERNAL_ERROR` — `EndpointModel.scala` emits *"SQL ${conf.warehouse} cannot be created due to repeated collisions on unique IDs."*; `SchedulerRpcContextValidatorHook.workspaceNotOwned(orgId)` interpolates orgId
- `IO_ERROR` — storage / network paths
- `UNKNOWN` — unclassified by definition

Also reframed the module-level docs from "DBR vs control plane" to "designed-for-user vs designed-for-debugging" — same construction sites can land in either bucket, and the latter framing matches how the codes are actually used in practice.

PR #408, opened 2026-06-02
feat(lakebase): support CJS consumers and validate exports via attw

### LINKED ISSUE #404: Support for CJS
We are trying to use the `lakebase` package in our codebase, unfortunately,  our codebase builds to CJS rather than ESM. The problem is that the said package builds to ESM, which means we can't use it :( Is there room to support CJS codebases too?

Thanks so much!

MarioCadenas:
Closes #404.

- Build @databricks/lakebase as dual ESM + CJS via tsdown, emitting .js/.d.ts and .cjs/.d.cts artifacts. Update package.json exports with conditional import/require entries; the development condition is preserved for source consumers.
- Enable tsdown's built-in attw integration in @databricks/lakebase, @databricks/appkit, and @databricks/appkit-ui so type-resolution correctness across node10, node16, and bundler is validated on every build (level: error). appkit and appkit-ui use the esm-only profile; lakebase uses strict.


package.json
```diff
@@ -55,6 +55,7 @@
     ]
   },
   "devDependencies": {
+    "@arethetypeswrong/core": "^0.18.4",
```
pkosiec: is it actually used? If so, we need to put an exact version 🙏

pkosiec (approved):

PR #415, opened 2026-06-03
feat(analytics): auto-start SQL warehouse and stream warehouse_status SSE events

MarioCadenas:
## Summary

The analytics `/query` route used to fail or freeze the UI on a stalled spinner when the configured SQL warehouse was `STOPPED`/`STARTING`. This wires up auto-start + readiness streaming so the frontend can give users feedback during cold starts (which can take 30s–2min).

## What changed

- **`SqlWarehouseClient.ensureWarehouseRunning()`** — new method that checks warehouse state, calls `warehouses.start` when `STOPPED`, polls until `RUNNING` (configurable timeout, default 5min), and reports `DELETED`/`DELETING` as a hard error.
- **`/query` route handler is now an async generator** that yields `warehouse_status` events over SSE while the warehouse warms up, followed by the regular `result`/`arrow`/`error` event for the SQL itself.
- **Caching/retries** are kept on the inner SQL `execute()` call only — the outer stream isn't cached because async generators yield once per consumer.
- **New analytics plugin config:** `warehouseStartupTimeoutMs` (default `300000`).
- **Test helpers** default `workspaceClient.warehouses.{get,start}` to a `RUNNING`-warehouse mock so existing tests keep passing without modification.

## Follow-up

A follow-up PR exposes `warehouseStatus` on `useAnalyticsQuery`, suppresses misleading chart skeletons during warmup, and adds a generic `ResourceStatusProvider` + drop-in indicator (also wired into the template + playground). That PR is independent of this one — the new SSE event simply stays unused until both land.

## Test plan

- [x] Unit tests for `ensureWarehouseRunning` covering `RUNNING`, `STOPPED→start→poll→RUNNING`, `STARTING→poll→RUNNING`, `DELETED`/`DELETING`, abort, and timeout (with `vi.useFakeTimers`).
- [x] Analytics SSE test verifying `warehouse_status` events arrive before `result`.
- [x] Existing analytics suite passes with the new test-helpers default mocks.
- [x] Manual smoke against a real STOPPED warehouse before merge.

### DIFF
--- a/packages/appkit/src/connectors/sql-warehouse/client.ts
+++ b/packages/appkit/src/connectors/sql-warehouse/client.ts
@@ -23,6 +23,8 @@ import {
   TelemetryManager,
 } from "../../telemetry";
 import { executeStatementDefaults } from "./defaults";
+import { WarehousePollBackoff } from "./warehouse-poll-backoff";
+import { WarehouseStatusEmitter } from "./warehouse-status-emitter";
 
 const logger = createLogger("connectors:sql-warehouse");
 
@@ -39,13 +41,6 @@ interface SQLWarehouseConfig {
  * serverless typically reaches RUNNING within ~30s.
  */
 export const DEFAULT_WAREHOUSE_STARTUP_TIMEOUT_MS = 5 * 60 * 1000;
-
-/**
- * Polling backoff bounds for the warehouse readiness loop. Backoff is
- * exponential with ±15% jitter to spread retries across concurrent waiters.
- */
-const WAREHOUSE_POLL_INITIAL_MS = 1_000;
-const WAREHOUSE_POLL_MAX_MS = 30_000;
 
 /**
  * Window during which a recent `RUNNING` observation lets subsequent calls
@@ -102,10 +97,9 @@ export class SQLWarehouseConnector {
   private _arrowProcessor: ArrowStreamProcessor | null = null;
 
   /**
-   * Per-warehouse "recently observed RUNNING" cache. Lets warm-path
-   * requests skip the readiness loop (and its `warehouses.get` round-trip)
-   * entirely. Map values are the timestamp of the last RUNNING observation;
-   * see {@link WAREHOUSE_RUNNING_CACHE_TTL_MS}.
+   * Per-warehouse cache of the last RUNNING observation timestamp. Used by
+   * {@link ensureWarehouseRunning} to short-circuit warm-path callers; see
+   * {@link WAREHOUSE_RUNNING_CACHE_TTL_MS}.
    */
   private _recentlyRunning = new Map<string, number>();
   // telemetry
@@ -351,19 +345,27 @@ export class SQLWarehouseConnector {
 
   /**
    * Wait until the SQL warehouse is in the `RUNNING` state, auto-starting it
-   * if currently `STOPPED`. Emits one {@link WarehouseStatusUpdate} per
-   * observation so callers can surface progress (e.g. over SSE) instead of
-   * letting the UI freeze on a cold warehouse.
+   * if currently `STOPPED`. Emits a {@link WarehouseStatusUpdate} whenever
+   * the observed state changes, so callers can surface progress (e.g. over
+   * SSE) instead of letting the UI freeze on a cold warehouse. Equal
+   * successive observations are de-duplicated.
+   *
+   * Fast path: if this connector recently observed the warehouse RUNNING
+   * (within {@link WAREHOUSE_RUNNING_CACHE_TTL_MS}), the call returns
+   * immediately without any SDK round-trip or status emission. This keeps
+   * cache-hit analytics requests off the Databricks control plane.
    *
    * Behaviour by initial state:
-   * - `RUNNING`: emits one update and returns immediately.
-   * - `STOPPED`: emits a `STARTING` update, calls
+   * - `RUNNING`: emits one update, caches the observation, returns.
+   * - `STOPPED`: emits a synthetic `STARTING` update, calls
    *   `workspaceClient.warehouses.start`, then polls until `RUNNING`.
+   *   When `autoStart: false`, throws `ConfigurationError` instead.
    * - `STARTING` / `STOPPING`: polls until `RUNNING`.
    * - `DELETED` / `DELETING`: throws `ConfigurationError.resourceNotFound`.
    *
-   * Aborts and timeouts surface as `ExecutionError`. The whole loop runs
-   * inside a `sql.warehouseReady` telemetry span.
+   * Aborts and timeouts surface as `ExecutionError`. The poll loop uses
+   * exponential backoff with ±15% jitter (1s → 30s cap) and runs inside a
+   * `sql.warehouseReady` telemetry span.
    */
   async ensureWarehouseRunning(
     workspaceClient: WorkspaceClient,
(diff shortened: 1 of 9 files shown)


fjakobs (approved):

packages/appkit/src/connectors/sql-warehouse/client.ts
pkosiec, line 1: # Code Review: feat(analytics): auto-start SQL warehouse and stream warehouse_status SSE events

**PR:** [#415](https://github.com/databricks/appkit/pull/415)
**Branch:** `mario/analytics-warehouse-readiness-sse`
**Base:** `main` (24b8da2f)
**Scope:** 9 files, ~500 changed lines (executable)

## Context

The analytics `/query` route failed or froze the UI when the SQL warehouse was STOPPED/STARTING. This PR wires up auto-start + readiness streaming so the frontend can give users feedback during cold starts (30s-2min).

## Overall Assessment

The code is **well-structured and thoughtful**. The decomposition into `WarehousePollBackoff`, `WarehouseStatusEmitter`, and `streamCallbacks` is clean. Error sanitization prevents SDK internals from leaking to clients. The `_recentlyRunning` cache keeps hot-path requests off the control plane. Test coverage is strong for the core `ensureWarehouseRunning` logic.

**Verdict: Ready with fixes** -- P2 findings should be addressed before merge; P3 items are at your discretion.

---

## Findings

### P2 -- Moderate

| File | Issue | Confidence |
|------|-------|------------|
| `client.ts:432-436` | Wrong error factory for STOPPED + autoStart=false | 100 |
| `client.ts:490-508` | Original error lost without OTel -- no debug log before sanitization | 75 |

- **client.ts:432-436** -- `ConfigurationError.resourceNotFound("SQL warehouse", ...)` is used when the warehouse exists but is STOPPED and `autoStart` is disabled. The resulting message "SQL warehouse not found. Warehouse ... is STOPPED" is semantically wrong -- the resource was found, it's in the wrong state. This will confuse operators reading error logs. Should use `new ConfigurationError(...)` directly with a clear message, or add a new `ConfigurationError.invalidState()` factory.

- **client.ts:490-508** -- `_throwSanitizedReadinessError` replaces non-AppKitError messages with a generic string. The original error is recorded on the OTel span, but if OTel isn't configured (common in dev/early deployments), the original SDK error is lost entirely. Adding a `logger.debug(...)` before wrapping would preserve debuggability without leaking to clients.

### P3 -- Low

| File | Issue | Confidence |
|------|-------|------------|
| `client.ts:405` | Effective timeout can overshoot by up to 30s | 100 |
| `analytics.ts:300` | Type cast bypasses compile-time safety | 75 |
| `sql-warehouse.test.ts` | Missing DELETING state test | 100 |
| `warehouse-status-emitter.ts` | No dedicated unit tests | 75 |

- **client.ts:405** -- Timeout is checked at loop top, after backoff sleep. Worst case: the sleep is 30s (max backoff), timeout expires mid-sleep, detected on next iteration. Effective wait = `timeoutMs + 30s`. Fix: check elapsed time inside `_sleepRespectingAbort` or use `Math.min(backoff, remaining)` for the sleep duration.

- **analytics.ts:300** -- `update.state as WarehouseStatus["state"]` casts `sql.State` to the narrower `WarehouseState` union. Currently safe because unknown states hit the `default` throw in the switch. A type guard (`isWarehouseState(state)`) would make this invariant explicit and catch regressions if the switch cases change.

- **sql-warehouse.test.ts** -- DELETED is tested but DELETING is not. Both should throw `ConfigurationError.resourceNotFound` (same branch), but DELETING is a distinct runtime state worth covering.

- **warehouse-status-emitter.ts** -- `WarehouseStatusEmitter` is only tested indirectly through `ensureWarehouseRunning` integration tests. Dedicated tests would cover edge cases: callback throws, dedup with different states, attempt counter accuracy.

---

## Pre-existing (not introduced by this PR)

None identified.

## Testing Gaps

- No test for concurrent `ensureWarehouseRunning` calls on the same warehouse (thundering herd on cold start -- not a bug, just documents behavior)
- No test for `streamCallbacks` generator cleanup on abort (covered by abort signal propagation but not explicitly tested)

## Residual Risks

- Multiple concurrent requests to a STOPPED warehouse all call `warehouses.start()` (idempotent, so safe, but noisy). A single-flight mechanism could optimize this but adds complexity -- acceptable for v1.
- The `_recentlyRunning` cache TTL (30s) means one `warehouses.get()` call per 30s per warehouse. Acceptable for typical traffic.

---

## Fix Plan

### Fix A: Wrong error factory for STOPPED + autoStart=false (P2)

**File:** `packages/appkit/src/connectors/sql-warehouse/client.ts:432-436`

Replace:
```typescript
throw ConfigurationError.resourceNotFound(
  "SQL warehouse",
  `Warehouse ${warehouseId} is STOPPED and analytics auto-start is disabled. Start the warehouse manually or set analytics.autoStartWarehouse=true.`,
);
```

With a direct `ConfigurationError` construction:
```typescript
throw new ConfigurationError(
  `Warehouse ${warehouseId} is STOPPED and auto-start is disabled. Start the warehouse manually or set analytics.autoStartWarehouse=true.`,
);
```

This avoids the misleading "not found" phrasing while keeping the same error class and status code.

### Fix B: Add debug log before error sanitization (P2)

**File:** `packages/appkit/src/connectors/sql-warehouse/client.ts:499-507`

Add a debug log of the original error before wrapping:
```typescript
// After: span.recordException(error instanceof Error ? error : new Error(String(error)));
// Add:
logger.debug(
  "Warehouse readiness check raw error for %s: %O",
  warehouseId,
  error,
);
```

This preserves debuggability in non-OTel environments. The `debug` level keeps it out of production logs by default.

### Fix C (optional): Clamp backoff sleep to remaining timeout (P3)

**File:** `packages/appkit/src/connectors/sql-warehouse/client.ts:459`

Replace:
```typescript
await this._sleepRespectingAbort(backoff.next(), signal);
```

With:
```typescript
const remaining = timeoutMs - (Date.now() - startTime);
await this._sleepRespectingAbort(Math.min(backoff.next(), remaining), signal);
```

This ensures the effective timeout never exceeds the configured value.

### Fix D (optional): Add DELETING state test (P3)

**File:** `packages/appkit/src/connectors/tests/sql-warehouse.test.ts`

Add after the DELETED test (~line 225):
```typescript
test("rejects when warehouse is DELETING", async () => {
  const get = vi.fn().mockResolvedValue({ state: "DELETING" });
  const start = vi.fn();
  const wsClient = { warehouses: { get, start } };
  const updates: any[] = [];

  await expect(
    connector.ensureWarehouseRunning(wsClient as any, "wh-del", {
      onStatus: (u) => updates.push(u),
    }),
  ).rejects.toThrow(/wh-del is DELETING/);

  expect(start).not.toHaveBeenCalled();
  expect(updates).toHaveLength(0);
});
```

## Verification

After applying fixes:
1. `pnpm build && pnpm docs:build`
2. `pnpm check:fix && pnpm -r typecheck`
3. `pnpm test` -- all existing + new tests pass
4. Manual smoke: configure a STOPPED warehouse, hit `/query/:key`, confirm `warehouse_status` SSE events stream before the result

pkosiec (approved):
LGTM overall, but please check the suggestions from the agentic review 🙏 thanks!

PR #428, opened 2026-06-09
ci: trigger dogfood eval pipeline on run-evals PR label

calvarjorge:
## What

Adds a GitHub Actions workflow that launches the dogfood eval pipeline ([job 398185277057549](https://dogfood.staging.databricks.com/jobs/398185277057549?o=6051921418418893)) for a PR when the **`run-evals`** label is present, and re-launches it on every new commit while the label stays on.

## How it works

- **Trigger:** `pull_request` with `types: [labeled, synchronize]`.
  - `labeled` + label is `run-evals` → run.
  - `synchronize` (new commit) + PR already has `run-evals` → run.
- **Commit:** passes `github.event.pull_request.head.sha` (the real PR head commit, never the synthetic merge commit) as the `appkit_ref` job param, so the pipeline can pull the code. Also sets `prompt_preset=custom-pr` and `tags=appkit_pr:<number>`.
- **Latest wins:** `concurrency` with `cancel-in-progress: true` (grouped per PR) guarantees the sticky `⏳ Eval running` comment always reflects the most recently triggered commit, even if an earlier run's job finishes first.
- **Comment:** a sticky comment (`.github/scripts/upsert-eval-comment.cjs`) links the evals-monitor PR page (`/prs/appkit/<number>`) and the triggered job run (`run_id` from the `run-now` response — no extra API call).
- **Auth:** OAuth M2M as the `apps-mcp-evals-runner` service principal. Credentials are scoped to the trigger step only, so the PR-authored comment script never sees them.

### Security notes
- Uses `pull_request` (not `pull_request_target`): repo secrets are withheld from fork PRs, so an external contributor cannot exfil the credentials even by editing the workflow. The `run-evals` label gate also requires write/triage access.
- Action SHAs are pinned.

## Required setup before this works
1. Create the **`run-evals`** label in the repo.
2. Generate an OAuth secret on the `apps-mcp-evals-runner` SP and add two repo secrets:
   - `EVALS_DATABRICKS_CLIENT_ID_DOGFOOD`
   - `EVALS_DATABRICKS_CLIENT_SECRET_DOGFOOD`
3. Grant the SP **CAN MANAGE RUN** on job `398185277057549`.

## Testing
Because this is a `pull_request`-triggered workflow, it runs the PR branch's version of the workflow, so it can be exercised on this PR once the label and secrets exist.

### DIFF
--- a/.github/scripts/upsert-eval-comment.cjs
+++ b/.github/scripts/upsert-eval-comment.cjs
@@ -1,0 +1,72 @@
+/**
+ * Upserts a sticky "Eval running" comment on a PR after the dogfood eval
+ * pipeline has been launched.
+ *
+ * Invoked via `actions/github-script`. Inputs come from environment vars:
+ *   PR_NUMBER - the pull request number
+ *   HEAD_SHA  - the commit the eval was launched for
+ *   RUN_JSON  - raw JSON from `databricks jobs run-now` (used to link the run)
+ */
+
+const MARKER = "<!-- pr-eval-run -->";
+const EVALS_MONITOR_URL =
+  "https://evals-monitor-6051921418418893.staging.aws.databricksapps.com";
+const DATABRICKS_HOST = "https://dogfood.staging.databricks.com";
+const JOB_ID = "398185277057549";
+const WORKSPACE_ID = "6051921418418893";
+
+module.exports = async ({ github, context }) => {
+  const { owner, repo } = context.repo;
+  const issue_number = Number(process.env.PR_NUMBER);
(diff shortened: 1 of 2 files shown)


calvarjorge:
Closing for now — GitHub Actions runners can't reach dogfood.staging (network perimeter), so we can't trigger the eval job directly from CI. Keeping the branch `jorge_calvar/eval_trigger` in case we revisit this with a staging-capable runner or inverted trigger. Replacing with a lightweight approach: post a link to the evals-monitor PR page where the eval can be started.

PR #431, opened 2026-06-09
fix: clean up dev-server warnings for agentic flow

pkosiec:
## Problem

In agentic flow, console shows two misleading warnings on a perfectly healthy scaffold, making it look broken:

1. The serving endpoint is listed **twice** in the `MISSING REQUIRED RESOURCES` banner (`serving_endpoint:Serving Endpoint` and `serving_endpoint:serving-default`), both asking to set `DATABRICKS_SERVING_ENDPOINT_NAME`.
2. `./.env not found. Continuing without it.` is printed **twice**.

## Root causes & fixes

**Serving double-registration (real bug).** `ServingPlugin.getResourceRequirements()` emitted a `serving-default` resource that never deduped against the static manifest `serving-endpoint` (different `resourceKey`), so the same env var was reported twice. It now returns `[]` in default/unnamed mode — the static manifest resource already covers the default endpoint for both runtime validation **and** scaffolding (`databricks apps init` reads only the manifest) — and only emits dynamic resources for explicitly *named* endpoints. No change to `appkit.plugins.json` / init.

**`.env not found` noise.** The `--env-file-if-exists=./.env` Node flag is redundant — AppKit already loads `.env` silently via `dotenv.config()` at import. Node 26 prints the "not found" notice for that flag, and `tsx watch` doubles it (supervisor + child). Removed from the template `dev` script.

**Agentic-mode gate.** Adds a `DATABRICKS_APPS_AGENTIC_MODE` flag: when set, `ResourceRegistry.enforceValidation()` skips the missing-resource warning/throw entirely. In agentic flows, resource env vars are injected out-of-band — resolved from `app.yaml` `valueFrom` against bound resources, only once a resource is bound — so the banner is premature noise during the build phase. The flag is **set by the agent runtime**.

## Notes
- Validation logic is unchanged/correct — suppression is display-only and cannot mask a value AppKit should have seen.
- Deployed apps continue to get `DATABRICKS_SERVING_ENDPOINT_NAME` via the platform's `valueFrom` resolution.

## Tests
- `resource-registry.test.ts`: agentic mode does not throw/warn on missing required resources (even in production), and takes precedence over `APPKIT_STRICT_VALIDATION`.
- `serving.test.ts`: default mode now emits zero dynamic requirements.
- All 47 tests in the two suites pass; typecheck + biome clean.

### DIFF
--- a/packages/appkit/src/plugins/serving/serving.ts
+++ b/packages/appkit/src/plugins/serving/serving.ts
@@ -85,11 +85,16 @@ export class ServingPlugin extends Plugin {
   static getResourceRequirements(
     config: IServingConfig,
   ): ResourceRequirement[] {
-    const endpoints = config.endpoints ?? {
-      default: { env: "DATABRICKS_SERVING_ENDPOINT_NAME" },
-    };
+    // Default (unnamed) mode is already covered by the static manifest resource
+    // (serving-endpoint), which also drives scaffolding via `databricks apps
+    // init`. Emitting a default resource here too would register a second,
+    // non-deduping entry (serving-default) for the same env var. Only emit
+    // dynamic resources for explicitly named endpoints.
+    if (!config.endpoints) {
+      return [];
+    }
 
-    return Object.entries(endpoints).map(([alias, endpointConfig]) => ({
+    return Object.entries(config.endpoints).map(([alias, endpointConfig]) => ({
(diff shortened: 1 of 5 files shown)


pkosiec:
Continued in https://github.com/databricks/appkit/pull/432?timeline_per_page=5

PR #432, opened 2026-06-09
fix: suppress resource-validation warnings in agentic flows

pkosiec:
## Problem

When an AppKit app runs inside an agentic flow, the dev-server console shows two misleading warnings on a perfectly healthy scaffold:

1. A `MISSING REQUIRED RESOURCES` banner for resource env vars (e.g. `DATABRICKS_SERVING_ENDPOINT_NAME`) that are provisioned out-of-band and simply aren't present yet at boot.
2. `./.env not found. Continuing without it.` — printed twice.

## Changes

**Agentic-mode gate.** Adds a `DATABRICKS_APPS_AGENTIC_MODE` flag. When set, `ResourceRegistry.enforceValidation()` skips the missing-resource warning/throw entirely. In an agentic flow, resource env vars are provisioned out-of-band and may be absent when the dev server first boots, so the banner is premature noise. The check runs before the dev/strict branch, so it takes precedence over `APPKIT_STRICT_VALIDATION`. The flag is set by the surrounding agentic environment and is intentionally **undocumented** (internal, not a public config knob).

**Remove redundant `--env-file-if-exists`.** AppKit already loads `.env` silently via `dotenv.config()` at import, so the flag in the template `dev` script is redundant — it only produces Node's "not found" notice (doubled under `tsx watch`). Removed from the `dev` script.

## Notes
- Validation logic itself is unchanged and correct — suppression is display-only and cannot mask a value AppKit should have seen.
- Default resource behavior is unchanged; the only new behavior is gated on `DATABRICKS_APPS_AGENTIC_MODE`.

## Tests
- `resource-registry.test.ts`: agentic mode does not throw/warn on missing required resources (even with `NODE_ENV=production`), and takes precedence over `APPKIT_STRICT_VALIDATION`.
- All 47 tests in the touched suites pass; typecheck + biome clean.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "start": "NODE_ENV=production node --env-file-if-exists=./.env ./dist/server.js",
-    "dev": "NODE_ENV=development tsx watch --tsconfig ./tsconfig.server.json --env-file-if-exists=./.env ./server/server.ts",
+    "dev": "NODE_ENV=development tsx watch --tsconfig ./tsconfig.server.json ./server/server.ts",
     "build:client": "tsc -b tsconfig.client.json && vite build --config client/vite.config.ts",
     "build:server": "tsc -b tsconfig.server.json && tsdown -c tsdown.server.config.ts",
     "build": "npm run build:server && npm run build:client",
--- a/packages/appkit/src/registry/resource-registry.ts
+++ b/packages/appkit/src/registry/resource-registry.ts
@@ -373,6 +373,13 @@ export class ResourceRegistry {
    */
   public enforceValidation(): ValidationResult {
     const validation = this.validate();
+
+    // Agentic flows provision resource env vars out-of-band, so a missing
+    // resource at dev-server boot is expected — skip the warning/throw.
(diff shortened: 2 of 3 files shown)


fjakobs (commented):
The env vars should be set in builder mode. I want to double check that before merging this PR.

MarioCadenas (changes requested):
I don't think this is the right approach, we are basically enforcing everything for 1 environment and completely breaking the other one, let's talk about it

template/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "start": "NODE_ENV=production node --env-file-if-exists=./.env ./dist/server.js",
-    "dev": "NODE_ENV=development tsx watch --tsconfig ./tsconfig.server.json --env-file-if-exists=./.env ./server/server.ts",
+    "dev": "NODE_ENV=development tsx watch --tsconfig ./tsconfig.server.json ./server/server.ts",
```
MarioCadenas, line 8: this breaks local dev

pkosiec: I believe it doesn't as we use `dotenv` but will confirm that 👍

PR #436, opened 2026-06-11
fix(agents): stop DatabricksAdapter duplicating answer text across tool calls (#421)

### LINKED ISSUE #421: DatabricksAdapter yields text content during tool-calling rounds, causing response duplication
### Problem

When using `DatabricksAdapter` with Claude models via Databricks Model Serving (OpenAI-compatible endpoint), the model produces substantial text content **alongside** tool calls in the same response turn. The adapter's `streamCompletion()` method yields all `delta.content` text as `message_delta` events regardless of whether the same response also contains tool calls.

In a multi-step ReAct loop, this causes the model's intermediate text (which often includes the full answer) to be emitted and displayed to the user on **every** tool-calling turn — resulting in the same answer appearing 3-4× in the final output.

### Root Cause Analysis

Claude's native API uses a trained-in tool-use system prompt that shapes the model to produce brief commentary before `tool_use` blocks. The Databricks OpenAI-compatible translation layer does not include this trained-in prompt, so Claude produces more verbose text during tool-calling rounds.

The adapter loop (`run()` method, line ~196) correctly breaks when no tool calls are present, but during tool-calling rounds, `streamCompletion()` yields ALL `delta.content` as `message_delta` events — there's no mechanism to suppress or buffer text that arrives in a turn that also contains tool calls.

### Expected Behavior

Text content from tool-calling rounds should either:
1. Be suppressed (not yielded as `message_delta`) when the same response also contains `tool_calls`
2. Be buffered until `streamCompletion` returns, and only yielded if no tool calls were detected
3. Be marked with metadata so consumers can distinguish intermediate text from final-answer text

### Current Workaround

We added:
- A system prompt instruction telling the model to only write brief status notes during tool-calling turns
- Client-side text buffering in the SSE transport layer that discards accumulated text when a `function_call` event arrives in the same round

### Environment

- AppKit version: 0.38.1
- Model: `databricks-claude-opus-4-6` via Model Serving
- Adapter: `DatabricksAdapter.fromServingEndpoint()`

### Reproduction

1. Create an agent with `maxSteps: 10` and multiple tools (e.g., `describeTable`, `executeSql`)
2. Ask a question that triggers 3+ tool calls
3. Observe that the assistant's text response contains the full answer repeated once per tool-calling round

MarioCadenas:
## Summary

Closes #421. Claude on Databricks Model Serving emits a full draft answer **alongside** its tool calls in the same turn, because the OpenAI-compatible layer lacks Claude's native tool-use prompt. `DatabricksAdapter` yielded every `delta.content` as a `message_delta` the instant it arrived, so in a ReAct loop the draft surfaced on every tool-calling step — the same answer appearing 3-4x. Because `consumeAdapterStream` *appends* `message_delta` content, it was also duplicated in the aggregated server-side output, not just the live display.

## Fix

Surface assistant text only on the **terminal** turn:

- `streamCompletion()` streams text live **only when no tools are offered** (such a turn is always terminal, so its text is the final answer). When tools are present it buffers the text into `fullText` instead of yielding it.
- `run()` emits the buffered text as a single `message_delta` on the terminal turn (no structured or text-encoded tool calls). Tool-calling turns keep their text in message history for model context but never surface it to the user.

## Tradeoff

With tools enabled, the final answer arrives as one delta rather than streaming token-by-token — there's no way to know a turn is terminal until its stream ends without tool calls. Pure-chat (no tools) still streams live, exactly as before.

## Verification

- `pnpm --filter=@databricks/appkit typecheck` ✅
- 2 new tests (text + tool-call in one turn → draft suppressed; no duplication across multiple steps) ✅
- Full appkit suite — **2141 tests** ✅

### DIFF
--- a/packages/appkit/src/agents/databricks.ts
+++ b/packages/appkit/src/agents/databricks.ts
@@ -398,6 +398,15 @@ export class DatabricksAdapter implements AgentAdapter {
 
     yield { type: "status", status: "running" };
 
+    // When tools are offered, streamCompletion buffers assistant text instead of
+    // streaming it live (#421), and run() surfaces it only on a terminal turn.
+    // If the loop instead exits by exhausting maxSteps while still tool-calling,
+    // there is no terminal turn — so we'd drop the last turn's text entirely and
+    // the caller would see nothing. Track the last buffered text and whether a
+    // terminal turn surfaced it, then flush it after the loop as a fallback.
+    let lastBufferedText = "";
+    let surfacedFinalText = false;
+
     for (let step = 0; step < this.maxSteps; step++) {
       if (context.signal?.aborted) break;
 
@@ -406,6 +415,7 @@ export class DatabricksAdapter implements AgentAdapter {
         tools,
         context,
       );
+      lastBufferedText = text;
 
       if (toolCalls.length === 0) {
         const parsed = parseTextToolCalls(text);
@@ -422,6 +432,7 @@ export class DatabricksAdapter implements AgentAdapter {
         if (tools.length > 0 && text) {
           yield { type: "message_delta", content: text };
         }
+        surfacedFinalText = true;
         break;
       }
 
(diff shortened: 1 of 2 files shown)


MarioCadenas:
Superseded by #446, which fixes the same #421 duplication but **preserves live streaming**.

The buffer-until-terminal approach here made the final answer arrive all at once (no token-by-token streaming with tools enabled). #446 instead keeps the adapter streaming every round and de-duplicates in the two consumers that were flattening the translator's per-round output items (`consumeAdapterStream` keeps only the terminal round; `useAgentChat` exposes ordered per-round `items` with `content` = the final message), plus a collapsible-steps UX in the playground. Closing in favor of #446.

PR #444, opened 2026-06-12
ci(template): verify lockfile uses only public npm registry

MarioCadenas:
## What

Adds a required CI check that fails if `template/package-lock.json` resolves any dependency from a non-public registry.

## Why

The `template/` directory is the app scaffold shipped to users via `databricks apps init`, and its `package-lock.json` pins exactly where each dependency is fetched from (the `resolved` field on every package entry). If a private/internal registry (Artifactory, JFrog, GitHub Packages, Verdaccio, an internal mirror) ever leaks into that lockfile — for example because it was regenerated on a machine with a custom `.npmrc` — scaffolded apps would either fail `npm install` (no access) or silently pull from a non-public source.

Today the lockfile is clean (1003 `resolved` entries, all `https://registry.npmjs.org/`), but nothing enforced it. This check closes that gap.

## How

New `tools/check-template-lock-registry.ts` (follows the existing `check-template-deps.ts` pattern):

- Iterates the `packages` map of `template/package-lock.json` (lockfileVersion 3) and requires every `resolved` URL to be `https://` with host `registry.npmjs.org`. Entries without `resolved` (root + workspace/link) are skipped. The single https+host rule also rejects non-registry sources like `git+ssh://`, `file:`, and `http://`.
- If `template/.npmrc` is ever added (none exists today), flags any private `registry=`/scoped-registry line or `_authToken=` line; absence is handled silently.

Wired as a new step in the existing required `lint_and_typecheck` job, next to `check-template-deps`, so it runs on every PR and blocks merge.

## Verification

- Happy path: current lockfile → exit 0.
- Injected JFrog URL and `git+ssh://` entry → both flagged, exit 1.
- Private `.npmrc` `registry=`/`_authToken=` flagged; legitimate `@scope:registry=https://registry.npmjs.org/` not flagged → exit 1.
- Missing `.npmrc` → exit 0, no error.
- Biome clean.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,8 @@ jobs:
         run: pnpm run check:licenses
       - name: Check template deps are pinned
         run: pnpm exec tsx tools/check-template-deps.ts
+      - name: Check template lockfile uses only public npm registry
+        run: pnpm exec tsx tools/check-template-lock-registry.ts
 
   test:
     name: Unit Tests
--- a/template/package-lock.json
+++ b/template/package-lock.json
@@ -8218,7 +8218,7 @@
     },
     "node_modules/clsx": {
       "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/clsx/-/clsx-2.1.1.tgz",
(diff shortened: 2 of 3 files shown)


pkosiec (approved):
LGTM

PR #450, opened 2026-06-13
feat(agents): forward generation params (temperature/top_p/stop/...) to the serving adapter

MarioCadenas:
## The gap

The Databricks agent adapter (`DatabricksAdapter`) sent only `max_tokens` to the OpenAI-compatible serving endpoint. App engineers had no way to tune generation (temperature, nucleus sampling, stop sequences, penalties) — a basic parity gap vs. the Vercel AI SDK / LangChain.

## What changed

Forward the standard OpenAI-compatible generation params into the serving request body:

- `temperature`
- `top_p`
- `stop` (string or string[])
- `frequency_penalty`
- `presence_penalty`

### Where they're declared (all optional / opt-in)

- `GenerationParams` interface in `packages/appkit/src/agents/databricks.ts`, accepted on every `DatabricksAdapter` options shape (raw fetch, stream-body, `fromServingEndpoint`, `fromModelServing`) and exported from `@databricks/appkit/beta`.
- `AgentDefinition.generationParams` and `RegisteredAgent.generationParams` in `packages/appkit/src/core/agent/types.ts`, wired through `packages/appkit/src/plugins/agents/agents.ts` into the adapter options (same path as `maxTokens`, only applied when AppKit builds the adapter from a string/omitted `model`).
- Markdown frontmatter `generationParams:` in `packages/appkit/src/core/agent/load-agents.ts`, parsed defensively (only known keys with the expected wire types are picked).

Names match the serving API wire keys. Only keys that are **set** are written to the request body; `undefined` keys are omitted so the endpoint applies its own defaults. No range validation is added — the endpoint validates.

## Verification

- `pnpm --filter=@databricks/appkit typecheck` — clean.
- `vitest run --project appkit` for the adapter, core agent, and agents plugin suites — 348 tests pass, including two new adapter tests asserting (a) set params are forwarded and (b) unset params are omitted from the body.
- `biome check` on the edited files introduces no new warnings (the two `agents.ts` warnings are pre-existing on `main`).


packages/appkit/src/core/agent/load-agents.ts
```diff
@@ -340,6 +348,32 @@ export function parseFrontmatter(
...
+  if (typeof raw.frequency_penalty === "number")
+    out.frequency_penalty = raw.frequency_penalty;
+  if (typeof raw.presence_penalty === "number")
+    out.presence_penalty = raw.presence_penalty;
+  if (
+    typeof raw.stop === "string" ||
+    (Array.isArray(raw.stop) && raw.stop.every((s) => typeof s === "string"))
+  ) {
+    out.stop = raw.stop as string | string[];
+  }
```
atilafassina: if user goes for something like `temperature: "0.5"` the code will silently ignore and uses the default value. Making the user believe temperature is not applying since that's valid YAML and we're not validating it. Right? 

Same goes for `top-p` or any other invalid combination.

How about we add another validation step and surface this issue more loudly?
or some log like "<key> is undefined or not a <right type>, using `<default>`"

wdyt?

atilafassina (commented):
Great work!
~2~ 1 quick question~s~  about the params parsing before I can approve this :)

packages/appkit/src/core/agent/load-agents.ts
MarioCadenas: sounds good, done

atilafassina (approved):

PR #459, opened 2026-06-22
chore: CLI typegen for Metric Views

atilafassina:
Adds Metric View type generation to **`generate-types`** (the same command + Vite plugin that produces query types — no separate command). When `config/queries/metric-views.json` is present, the run DESCRIBEs each declared UC Metric View and emits **`metric-views.d.ts`**, which augments `MetricRegistry` so `useMetricView('<key>', …)` is autocompleted and type-checked. Absent that file, the path stays dormant and nothing is emitted.

> [!NOTE]
> Dormant, not yet user-facing, **types only**. Apps opt in by adding `metric-views.json`. The runtime metric path (route, SQL engine, `useMetricView` hook) lands in later PRs of the stack.

## What it delivers

- **One entry point** — metric-view typegen is folded into `generate-types`, sharing its warehouse round-trip and cache. `shared` reaches appkit's `syncMetricViewsTypes` through the existing dynamic `import("@databricks/appkit/type-generator")`, keeping **no static dependency** on appkit.
- **One artifact** — `metric-views.d.ts` (compile-time only). Encodes each view's measures, dimensions, key unions (`measureKeys`/`dimensionKeys`/`timeGrains`), and per-column semantic metadata (`type`, `display_name`, `format`, `time_grain`). No runtime metadata bundle — that's deferred to the runtime PR where a real consumer drives its shape.
- **Warehouse-readiness contract** — identical to query types: non-blocking runs write permissive types + warn on a view that can't be described; `--wait` fails the build so CI never ships incomplete types.
- **Cache holds only good describes** — keyed on `source|lane`; a degraded outcome is rendered into the `.d.ts` (so `tsc` never breaks) but never cached, so `--wait` re-describes a still-bad source every run instead of serving a sticky degrade. A malformed `metric-views.json` fails fast in every mode.

## Versioning

Everything under `shared/appkit-types/` is a build artifact; the versioned input is `config/queries/*`. The template gitignores `metric-views.d.ts` (regenerated at build via `prebuild`); dev-playground tracks it as a demo. Documented in `docs/docs/development/type-generation.md`.


packages/shared/src/cli/commands/metric-views/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { metricViewsSyncCommand } from "./sync/sync";
+
+/**
+ * Parent command for UC Metric View operations.
+ *
+ * Exposes a single subcommand (`sync`).
+ * Future subcommands (`list` / `validate` / `describe`) plug in here so users have one top-level surface for everything related to Metric Views.
+ * Sibling of `plugin`, `setup`, `generate-types`, `lint`, `docs`, `codemod`.
+ */
+export const metricViewsCommand = new Command("mv")
```
pkosiec: I'm not sure about `mv`; as an additional alias - maybe (however I'd still think at first that it's something functionally close to unix `mv`).

The `metric-view` command would be consistent with what we have (like `plugin`). What do you think?
(Another thing is that we have `docs` (plural) and `plugin` (singular) - so we're already quite inconsistent 😞 not worth to change it now though)

What do you think?

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+/**
+ * Options parsed by commander for `appkit mv sync`.
+ *
+ * interactive clack flow:
+ *   - `--warehouse-id` (+ `DATABRICKS_WAREHOUSE_ID` fallback)
+ *   - `--metric-views-json-path` (canonical config path)
+ *   - `--output-dir` (artifact output directory; replaces Phase 1's `--out-dir`)
+ *   - `--no-cache` (commander negation → `cache === false` disables the metric type-generation cache)
+ */
+export interface MetricViewsSyncOptions {
```
pkosiec: I'm a fan of the approach you suggested here (explicit flags) but I think we should consider being consistent across all commands. And the `generate-types` command uses positional args:
```
appkit generate-types [rootDir] [outFile] [warehouseId]  --no-cache  --wait
```
 
Personally, I'd rather prefer updating the `generate-types` command to use explicit flags but that would be a breaking change for users 🤔 maybe it's worth to do it sooner than later?


It's not a blocker but at least we should plan a follow-up to unify the commands.

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+      }
+      process.exit(EXIT_FAILURE);
+      return;
+    }
+
+    // Degraded-but-not-failed (e.g. a not-ready warehouse returned no schema for
+    // some keys): the permissive types ARE written, so unlike `result.failures`
+    // above this is a warning — not a hard failure — and the command still exits
+    // 0. The degraded entries refresh on a rerun once the warehouse is available.
+    const degradedKeys = result.schemas
```
pkosiec: **[P2 · parity, optional]** No `--wait` equivalent to `generate-types`. On a cold warehouse `mv sync` degrades and exits 0 — which is *consistent* with `generate-types`' default (its `--wait` is documented as "wait for warehouse readiness instead of degrading"), so this is **not a bug**.

The only gap is the missing CI opt-in: consider adding `--wait` → `mode: "blocking"` (default unchanged) and exiting non-zero on a still-degraded result under `--wait`.

For context, the mechanics: `describe-now` has no preflight; a cold `DESCRIBE` times out at ~30s, and `describeOne` returns a degraded schema with **no failure** (`mv-registry/sync.ts:83-85`), so `result.failures` is empty and this branch warns + exits 0.

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+      return;
+    }
+
+    onProgress?.succeed(`Generated metric types: ${metricOutFile}`);
+    console.log(`Generated metric types: ${metricOutFile}`);
+    console.log(`Generated metric metadata: ${metricMetadataOutFile}`);
+  } catch (error) {
+    if (
+      error instanceof Error &&
+      error.message.includes("Cannot find module")
```
pkosiec: **[P2]** This `catch` wraps both the dynamic `import("@databricks/appkit/type-generator")` and the entire `syncMetricViewsTypes` call, so *any* nested error whose message contains `Cannot find module` (a broken transitive dep, a bad subpath export inside appkit) is misreported as "appkit not installed" — sending the user to reinstall a package that's present.

Narrow the match, e.g. `(err as NodeJS.ErrnoException).code === "ERR_MODULE_NOT_FOUND" && err.message.includes("@databricks/appkit")`, or scope the heuristic to just the `import` call.

packages/shared/src/cli/commands/type-generator.d.ts
```diff
@@ -25,4 +24,60 @@ declare module "@databricks/appkit/type-generator" {
...
+   *
+   * Reads `metric-views.json` from `queryFolder`, DESCRIBEs any metric views
+   * that are missing from the cache, then writes `metric-views.d.ts` and
+   * `metric-views.metadata.json`. This only syncs metric-view types; analytics query
+   * types are generated separately.
+   *
+   * The cache is enabled by default. Pass `cache: false` to force fresh
+   * DESCRIBE results, matching the CLI's `--no-cache` flag.
+   */
+  export function syncMetricViewsTypes(options: {
```
pkosiec: **[P2]** This ambient module hand-mirrors appkit's real export across the package boundary, with no compile-time link to it: `measures`/`dimensions` are `unknown[]` here vs typed arrays in appkit, and `mode`/`metricFetcher` are omitted. Nothing fails to compile if the real `SyncMetricViewsTypesResult`/`MetricSchema` contract drifts.

At minimum add a comment that this intentionally narrows the signature; ideally a type-assignability check. (Adding `mode?:` here is also the prerequisite for the `--wait` parity suggestion in `sync.ts`.)

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+      );
+      console.warn(
+        `Warning: ${degradedKeys.length} metric view(s) (${degradedKeys.join(", ")}) could not be described — the warehouse wasn't ready, so permissive types were written. Rerun \`appkit mv sync\` once the warehouse is available.`,
+      );
+      console.log(`Generated metric metadata: ${metricMetadataOutFile}`);
+      return;
+    }
+
+    onProgress?.succeed(`Generated metric types: ${metricOutFile}`);
+    console.log(`Generated metric types: ${metricOutFile}`);
```
pkosiec: **[P3]** In interactive mode this prints twice — `onProgress.succeed(...)` on the line above already stops the spinner with the same "Generated metric types: …" text, then this `console.log` repeats it. Gate the `console.log` on `onProgress === undefined` (only the flag-driven path needs the plain logs).

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+  };
+
+  const s = spinner();
+  await runMetricViewsSync(options, {
+    start: () => s.start("Describing metric views…"),
+    succeed: (msg) => s.stop(msg),
+    fail: () => s.stop("Failed."),
+  });
+
+  outro("Metric types synced.");
```
pkosiec: **[P3]** When `metric-views.json` is absent, `runMetricViewsSync` logs "Nothing to sync" and returns normally (~line 124), but this path still prints `outro("Metric types synced.")` — a contradictory success message. Have `runMetricViewsSync` signal the dormant/no-op outcome so the outro can match (or be skipped).

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+async function runInteractive(): Promise<void> {
+  intro("Sync UC Metric View types");
+
+  // A cancelled prompt (Ctrl-C) is a graceful, non-zero exit. The explicit
+  // `return` after `process.exit` keeps control flow correct under a no-op exit
+  // (tests) — without it, a cancelled flow would fall through to the next
+  // prompt and eventually run the sync.
+  const cancelled = (): never => {
+    cancel("Cancelled.");
+    process.exit(1);
```
pkosiec: **[P3 · consistency]** Cancel exits `1` here, but every cancel in `plugin create` / `add-resource` exits `0`, and clack's docs use `process.exit(0)` after `cancel(...)`. A Ctrl-C out of `mv sync` returns a failure code while the same gesture elsewhere is graceful, and wrapper scripts treating non-zero as error will misread it. Suggest `process.exit(0)` (and flip the two cancel assertions in `sync.test.ts`).

packages/shared/src/cli/commands/metric-views/index.ts
pkosiec: **[P3 · naming, your call]** `mv` is the only abbreviated subcommand — `plugin` / `setup` / `generate-types` / `lint` / `docs` / `codemod` are all full words — and it collides with the Unix `mv` (move) command. `metrics` or `metric-views` would match the family; the short `mv` is more ergonomic. Flagging for a deliberate decision rather than drift.

packages/shared/src/cli/commands/metric-views/sync/sync.ts
```diff
@@ -0,0 +1,392 @@
...
+  } else {
+    await runInteractive();
+  }
+}
+
+export const metricViewsSyncCommand = new Command("sync")
+  .description(
+    "Sync UC Metric View schemas: DESCRIBE every entry in metric-views.json, then emit metric-views.d.ts + metric-views.metadata.json.",
+  )
+  .option(
```
pkosiec: **[P3 · UX consistency, your call]** The two type-gen commands now diverge: `generate-types [rootDir] [outFile] [warehouseId] --no-cache --wait` (positional) vs `mv sync --warehouse-id --output-dir … --no-cache` (named flags); and on a missing warehouse `generate-types` skips with a message while `mv sync` hard-fails (exit 1). The flag-based UX here is arguably the better design — flagging the divergence for a conscious call (is the added strictness intended?).

pkosiec (commented):
Reviewed the `mv-cli` typegen changes — solid refactor, and behavior parity for the dev/Vite (`non-blocking`/`blocking`) paths reads as preserved 1:1 (the extracted `syncMetricViewsTypes` keeps `serveDegraded = mode !== "describe-now"`, the preflight, the #406 gate, the sticky-degraded notice, cache save/prune, and the `fatalErrors`→throw ordering). Findings inline below — mostly P3 polish; nothing blocks correctness of the extraction. Verified the `--no-cache` / `getOptionValueSource` handling and the `@clack/prompts` usage against their docs — both correct.

Two nits not worth their own threads:
- Stale `metric.d.ts` / `metrics.metadata.json` aren't cleaned up on the rename to `metric-views.*` (`removeOldGeneratedTypes` only removes `appKitTypes.d.ts`) — low impact for a pre-release feature, but an upgraded app could keep an orphaned `metric.d.ts`.
- The CLI's Zod pre-validation doesn't mirror `resolveMetricConfig`'s 200-entry / 255-char-segment caps, so an over-cap config passes the "fail fast before the dynamic import" stage and throws only after it (still exit 1, just later).

Verdict: ready with minor fixes.

packages/shared/src/cli/commands/type-generator.d.ts
```diff
@@ -1,4 +1,20 @@
...
+ * `shared` must not statically depend on `appkit` (it is a leaf package), so the
+ * CLI reaches appkit's type-generator through a dynamic
+ * `import("@databricks/appkit/type-generator")`; this declaration types that
+ * import without a build-time dependency on appkit.
+ *
+ * The mirror is deliberately narrower than the real export: array element types
+ * are widened to `unknown[]`, options the CLI never passes (e.g. `metricFetcher`)
+ * are omitted, and only the surface the CLI actually uses is declared.
+ *
+ * DRIFT WARNING: there is NO compile-time link to appkit's real types — if the
```
calvarjorge: maybe we should move the cli to the appkit package?

calvarjorge (approved):

packages/shared/src/cli/commands/type-generator.d.ts
```diff
@@ -1,14 +1,35 @@
-// Type declarations for optional @databricks/appkit/type-generator module
+/**
```
MarioCadenas, line 1: can we remove this whole comment or make it shorter? I don't know how much value it adds 😅

packages/appkit/tsdown.config.ts
```diff
@@ -4,7 +4,14 @@ export default defineConfig([
...
     name: "@databricks/appkit",
-    entry: ["src/index.ts", "src/beta.ts"],
+    // `./type-generator` is a public subpath export consumed cross-package by the
+    // `appkit generate-types` CLI via a dynamic import Rolldown can't see. It must
+    // be its own entry so the names the CLI imports at runtime
+    // (generateFromEntryPoint — which additively emits metric-view types — and
+    // generateServingTypes) are preserved under unbundle tree-shaking. Without it,
+    // the subpath's runtime exports collapse to only the names appkit's own Vite
+    // plugins import — silently dropping the CLI's.
+    entry: ["src/index.ts", "src/beta.ts", "src/type-generator/index.ts"],
```
MarioCadenas: do we still need this?

packages/shared/src/cli/commands/type-generator.d.ts
atilafassina: it looks like it should. But I need first to have look at why it hasn't been there in the first place. I'm sure there's a tradeoff to consider

MarioCadenas (approved):

PR #461, opened 2026-06-24
feat: split out @databricks/lakebase-auth with eager refresh and retry

jawj:
Extract OAuth credential/token-refresh logic from @databricks/lakebase.

Move these to a new, lightweight @databricks/lakebase-auth package.

Add eager (default) and lazy refresh, plus retry on transient failures.


.github/workflows/prepare-release-lakebase-auth.yml
pkosiec, line 1: This is just a half of the changes required for the release - we'll need to:
- do the second part in the secure release repo (polling + actual release)
- create and configure the new package on npm side (with Fabian)
- allowlist the package in both internal npm proxies

I can take care of those items after my OOO 👍

packages/lakebase-auth/package.json
pkosiec, line 1: What are the reasons to extract this as a separate package already? Is it only about the size or do you have other concerns?

I totally get the idea of exporting such useful auth-related helpers - but wouldn't it be easier if we did it as a part of the `@databricks/lakebase` package from the maintenance point of view?

pkosiec (commented):
Thank you George for your contribution - this looks awesome!

Took a quick look (didn't test) but my general question would be: does it make sense to introduce a separate package right now? It brings more maintenance burden and I'm not sure if the benefits outweighs the increased complexity.

I'd vote for making all the exports part of the Lakebase package directly, and later on we could split it if there will be such requests from the community. What do you think?

packages/lakebase/src/pool-config.ts
```diff
@@ -44,36 +163,7 @@ export function getLakebasePgConfig(
...
-    port: poolConfig.port,
-    user: username,
-    database: poolConfig.database,
-    password: passwordConfig,
-    ssl: poolConfig.ssl ?? mapSslConfig(poolConfig.sslMode),
-    max: poolConfig.max,
-    idleTimeoutMillis: poolConfig.idleTimeoutMillis,
-    connectionTimeoutMillis: poolConfig.connectionTimeoutMillis,
-  };
+  return buildLakebasePgConfig(config, telemetry, logger).poolConfig;
```
pkosiec, line 166: **P2 (correctness, reliability) · confidence 100** — `getLakebasePgConfig` returns `buildLakebasePgConfig(...).poolConfig` and discards `dispose`; `getLakebaseOrmConfig` builds on it and exposes no `dispose` either. With the new `mode: "eager"` default, `createPasswordProvider` starts an `unref`'d background timer that re-fetches an OAuth credential roughly hourly **forever, per call, with no way to stop it** — a regression from the old lazy (timer-free) behaviour for the documented TypeORM/Sequelize path. `createLakebasePool` is unaffected (it wraps `pool.end` → `dispose`).

**Fix:** default these two helpers to `refresh: "lazy"` (matches old behaviour), or return `{ ...config, dispose }` and document calling it on shutdown.

packages/lakebase/src/pool-config.ts
```diff
@@ -44,36 +163,7 @@ export function getLakebasePgConfig(
   telemetry?: DriverTelemetry,
```
pkosiec, line 163: **P3 (maintainability) · confidence 100** — `telemetry?`/`logger?` are vestigial: every caller passes only `config`, yet they're part of the exported `getLakebasePgConfig` signature. Likewise `DriverTelemetry` and `TokenRefreshDeps` (the latter tied to the deprecated `createTokenRefreshCallback`) are re-exported from `packages/lakebase/src/index.ts` with no importer outside the package.

**Fix:** drop the dead params and trim those two exports.

packages/lakebase/src/pool-config.ts
```diff
@@ -1,27 +1,146 @@
...
+              ? { claims: deps.userConfig.claims }
+              : {}),
+          });
+          const expiresAt = new Date(credential.expire_time).getTime();
+          span.setAttribute(
+            "lakebase.token.expires_at",
+            new Date(expiresAt).toISOString(),
+          );
+          span.setStatus({ code: SpanStatusCode.OK });
+          span.end();
```
pkosiec, line 75: **P2 (reliability) · confidence 100** — `span.end()` only runs on the happy path. When `generateDatabaseCredential` throws, the `catch` at L79 rethrows without ending the span (raw OTel `startActiveSpan` does not auto-end on throw). Every failed refresh leaks an unclosed span.

**Fix:** `try { … span.setStatus({code: OK}) } catch (e) { span.recordException(e); span.setStatus({code: ERROR}); throw e } finally { span.end() }`.

packages/lakebase-auth/src/retry.ts
```diff
@@ -0,0 +1,42 @@
...
+export function withRetries<T>(
+  fn: () => Promise<T>,
+  schedule: number[] = DEFAULT_RETRY_SCHEDULE,
+  onLog?: LogFn,
+): () => Promise<T> {
+  return async function retrying(): Promise<T> {
+    for (const retryDelay of schedule) {
+      try {
+        return await fn();
+      } catch (err) {
```
pkosiec, line 28: **P2 (adversarial, reliability) · confidence 100** — `withRetries` retries on *every* thrown error. A non-transient failure (403/permission, `ConfigurationError`, `ValidationError`) is still retried across the full `[50, 500, 5000]ms` schedule (~5.5s) before surfacing — and in eager mode this compounds with the 30s background reschedule, so a misconfigured endpoint fires 4 doomed fetches every 30s for the process lifetime. The docs promise *transient* retries.

**Fix:** add a `shouldRetry?: (err) => boolean` predicate (default-true for back-compat) and fast-fail 4xx/config/validation errors.

packages/lakebase-auth/src/config.ts
```diff
@@ -1,36 +1,72 @@
...
+ *
+ * When `host` is a DNS name (not an IP literal), the `Bun.SQL` SNI server name
+ * is set on the returned object — see {@link DriverSslConfig.serverName}.
+ */
+export function mapSslConfig(sslMode: SslMode, host?: string): DriverSslConfig {
+  switch (sslMode) {
+    case "verify-full": // since JS drivers check root certs, there's an implied sslrootcert=system
+    case "verify-ca": // upgraded to equivalent of verify-full, sslrootcert=system
+    case "require": // upgraded to equivalent of verify-full, sslrootcert=system
+    case "prefer": {
```
pkosiec, line 46: **P2 (api-contract; security: improvement) · confidence 100** — `prefer` (and `require`) now fall through to `{ rejectUnauthorized: true }`, and the default changed `require` → `verify-full`. Old `@databricks/lakebase` mapped `prefer` → `{ rejectUnauthorized: false }`. This is a **security improvement**, but a *silent breaking behaviour change*: a deployment relying on `prefer`/`PGSSLMODE=prefer` against an unverifiable cert will now fail to connect.

**Fix:** add a `BREAKING CHANGE` changelog entry and bump `@databricks/lakebase` minor (0.3.0 → 0.4.0); a `fix:`/`chore:` commit only yields a patch.

packages/lakebase-auth/src/caching.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  function refresh(): Promise<string> {
+    if (inFlight) return inFlight;
+    inFlight = (async () => {
+      try {
+        const { token, expiresAt: exp } = await fetchCredential();
+        cachedToken = token;
+        expiresAt = exp;
+        scheduleNext(exp - Date.now() - earlyRefreshMs);
```
pkosiec, line 62: **P2 (adversarial) · confidence 90** — the next-refresh delay `exp - Date.now() - earlyRefreshMs` is floored by `Math.max(0, …)` in `scheduleNext`. When it's ≤0 — `earlyRefreshMs >= token lifetime` (user-set), clock skew, an already-near/expired token, or a `NaN` expiry from an unparseable `expire_time` — the timer fires at delay 0 and reschedules 0: an unbounded loop hammering the credential API (`unref`'d, so it won't block exit, but runs for the whole process lifetime).

**Fix:** clamp to a `MIN_REFRESH_MS` floor, guard `Number.isFinite(delay)`, and treat expired/near-expired as fetch-now-then-`BACKGROUND_RETRY_MS`-backoff.

pkosiec (commented):
## Code review — `@databricks/lakebase-auth` split

Automated multi-reviewer pass (correctness, security, adversarial, reliability, api-contract, testing, maintainability, project-standards). Load-bearing findings were verified against source. **No P0/P1 code defects.** 11 inline comments below (P2/P3). The DCO sign-off issue is intentionally excluded per request.

### Priority order
1. **Eager-timer leak** — `getLakebasePgConfig`/`getLakebaseOrmConfig` drop `dispose`, leaking an unstoppable background refresh timer (regression). *(pool-config.ts:166)*
2. **SSL `prefer` breaking change** — now verifies certs; document + bump `@databricks/lakebase` minor. *(config.ts:46)*
3. **Retry classification + min-refresh floor** — together they prevent credential-API storms on misconfig/clock-skew. *(retry.ts, caching.ts)*
4. **Release pin/ordering + knip** (below).
5. Span leak, over-wide exports, lazy-mode token discard, and the P3s as follow-ups.

### CI / template (looks good)
- `pack:prerelease` packs all four `@databricks/*` tarballs (lakebase-auth has `tarball:prerelease`), and `prepare-template-artifact.ts` copies the lakebase-auth tarball + sets its `overrides` entry — so the PR template artifact resolves the new package.
- `prepare-release-lakebase-auth.yml` correctly mirrors the lakebase two-stage workflow (build → dist → SBOM → pack → SHA256 → upload; no publish/tag/push).
- New package meets the CLAUDE.md "Creating New Packages" bar (`build:package`/`build:watch`, extends root tsconfig); typecheck wiring already fixed on this branch.

### One CI gap with no diff line to anchor (#9)
**`knip.json` does not exempt `packages/lakebase-auth`**, while it *does* exempt `packages/lakebase` (`ignoreWorkspaces`). So `pnpm run knip` (CI lint job) analyzes the new package, and combined with the over-wide exports (see the `index.ts` comment) it may fail. Run `pnpm knip` locally; either add lakebase-auth to `ignoreWorkspaces` for consistency, or trim the exports.

### Testing gaps (suggestions, not gates)
- `queryWithTelemetry` wrapper (`pool.ts:90-148`) is untested — only `pool.query.name` is asserted; the promise/callback branches, error-span recording, `queryDuration`, and `db.rows_affected` are never exercised.
- Eager-mode in-flight dedup is untested (tests flush the initial fetch before `getToken()`).
- No test that eager-mode `getLakebasePgConfig`/`getLakebaseOrmConfig` avoid leaking a timer; no test that `withRetries` does *not* retry a non-transient error; no negative assertion that tokens never appear in logs/errors; no near/at-expiry or `NaN`-expiry scheduling test.

### Residual risks (non-blocking)
- `sslMode: "disable"` → no TLS, token sent in cleartext (opt-in, default `verify-full`); worth a doc warning.
- No wall-clock timeout on the credential fetch (SDK socket ~5s is the only bound; worst-case retry time can exceed `pg.Pool` `connectionTimeoutMillis`).
- `getUsernameWithApiLookup` swallows all errors → silent `undefined` username.
- `getLakebaseOrmConfig` drops `serverName` when narrowing `ssl` → Bun.SQL via the ORM path loses SNI.

packages/lakebase-auth/src/caching.ts
```diff
@@ -0,0 +1,142 @@
...
+        return token;
+      } catch (err) {
+        onLog?.(
+          "warn",
+          "Background token refresh failed, retrying in %dms: %s",
+          BACKGROUND_RETRY_MS,
+          err instanceof Error ? err.message : String(err),
+        );
+        scheduleNext(BACKGROUND_RETRY_MS);
```
pkosiec, line 71: **P3 (reliability) · confidence 75** — `scheduleNext(BACKGROUND_RETRY_MS)` reschedules a failed background refresh exactly 30s later with no jitter. Multiple pools/processes started together that hit the same outage all retry at the same instants (T+30s, T+60s…) — a synchronized thundering herd.

**Fix:** add jitter, e.g. `BACKGROUND_RETRY_MS + Math.random() * BACKGROUND_RETRY_MS * 0.5`.

packages/lakebase-auth/src/caching.ts
```diff
@@ -0,0 +1,142 @@
...
+      return token;
+    } catch (err) {
+      refreshAfter = PAST_EPOCH; // allow the next call to retry
+      throw err;
+    }
+  }
+
+  return {
+    getToken() {
+      if (Date.now() > refreshAfter) cachedToken = refresh();
```
pkosiec, line 135: **P2 (adversarial) · confidence 80** — `getToken()` does `cachedToken = refresh()`, overwriting the cache with a possibly-rejecting promise while the previous token is still valid for up to `earlyRefreshMs` (default 2 min). A transient outage in that pre-expiry window makes lazy mode fail **100% of new connections** even though a usable token was in hand (eager mode doesn't have this).

**Fix:** keep the last successfully-resolved token+expiry separately and serve it until actual expiry; only surface the rejection once no valid token remains.

packages/lakebase-auth/src/index.ts
```diff
@@ -0,0 +1,42 @@
+export {
+  getUsernameSync,
+  getUsernameWithApiLookup,
+  getWorkspaceClient,
+  mapSslConfig,
```
pkosiec, line 5: **P2 (maintainability, api-contract) · confidence 100** — `parseConfig`, `ParsedAuthConfig`, `mapSslConfig`, `getUsernameSync`, `withRetries`, `DEFAULT_RETRY_SCHEDULE` are exported from the package entry but imported by no consumer outside this package. Once published they become committed public API (breaking to remove later).

**Fix:** trim the surface to `getPgConfig`, `createPasswordProvider`, `generateDatabaseCredential`, `getUsernameWithApiLookup`, errors and types; keep the rest as internal module exports. (Also addresses the knip note in the review summary.)

tools/dist-lakebase.ts
```diff
@@ -13,10 +13,26 @@ const prerelease = values.prerelease;
...
 
+for (const depName of WORKSPACE_PACKAGE_REPLACEMENTS) {
+  if (pkg.dependencies?.[depName] === "workspace:*") {
+    const pkgDirName = depName.split("/").pop() ?? depName;
+    const depPkgPath = path.join(
+      __dirname,
+      `../packages/${pkgDirName}/package.json`,
+    );
+    const depPkg = JSON.parse(fs.readFileSync(depPkgPath, "utf-8"));
+    pkg.dependencies[depName] = `${depPkg.version}`;
```
pkosiec, line 32: **P2/P3 (api-contract, project-standards) · confidence 75/70** — two issues on this version pin:

- **Exact pin + release ordering:** lakebase's `workspace:*` dep on `@databricks/lakebase-auth` is rewritten to the *exact* version. The two `prepare-release-*` workflows are independent and path-triggered, so nothing guarantees `lakebase-auth@0.1.0` publishes before the lakebase release that requires it (first publish breaks otherwise); the exact pin also blocks lakebase-auth patch fixes from reaching lakebase consumers without re-releasing lakebase. **Fix:** emit `` `^${depPkg.version}` `` and coordinate lakebase-auth-first on the initial publish.
- **Prerelease mismatch:** in a prerelease build the lakebase-auth tarball is `0.1.0-pr.<sha>` but this still pins plain `0.1.0`. It only resolves in the template because `prepare-template-artifact.ts` adds an `overrides` entry (matched by name); installing the lakebase prerelease tarball directly would try to resolve `0.1.0` from the registry → stale/missing.

packages/lakebase-auth/src/__tests__/caching.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    const fetch = vi
+      .fn<() => Promise<Credential>>()
+      .mockRejectedValueOnce(new Error("boom"))
+      .mockResolvedValue({ token: "t-ok", expiresAt: Date.now() + HOUR });
+    const onLog = vi.fn();
+
+    const provider = cachedWithTimedRefresh(fetch, EARLY, onLog);
+    await vi.advanceTimersByTimeAsync(0); // initial attempt fails
+    expect(fetch).toHaveBeenCalledTimes(1);
+    expect(onLog).toHaveBeenCalled();
```
pkosiec, line 64: **P3 (testing) · confidence 95** — `expect(onLog).toHaveBeenCalled()` only proves the function ran; it won't catch a wrong level, format string, or args. (Same pattern at `pool-config.test.ts:193` with `logger.error`.) Compare `retry.test.ts`, which asserts all four args.

**Fix:** `expect(onLog).toHaveBeenCalledWith('warn', expect.stringContaining('retrying'), 30000, expect.stringContaining('boom'))`.

PR #462, opened 2026-06-25
feat(cli): add 'appkit add' and 'appkit registry' commands

MarioCadenas:
Adds a component/plugin registry workflow to the AppKit CLI. Items are fetched directly from the [`databricks/appkit-registry`](https://github.com/databricks/appkit-registry) repo and written into the consumer's app — no `shadcn` and no `components.json` required. Registry commands are wired up but hidden from `--help` while the feature is still in development.

## New CLI surface

- **`appkit add <item...>`** (top-level, the primary entry point) — adds UI components and/or server plugins from the registry. Auto-detects each item's type and routes it to the right place:
  - UI components → `<frontend>/src/components/appkit/`
  - Server plugins → `<server>/plugins/<name>/`, then runs `appkit plugin sync --write` and registers the plugin in the server's `createApp` call (`--no-register` to skip the edit).

  Frontend/server roots are detected from common monorepo layouts, so it can be run from the repo root. Transitive `registryDependencies` are resolved breadth-first (deduped, fetched concurrently per level), and npm `dependencies` are installed with the detected package manager (pnpm/yarn/bun/npm).
- **`appkit registry list`** — enumerates items from the registry index, showing each item's kind (component/plugin/hook/…) and a marker for verified items (colorized output).
- **`appkit registry search <terms...>`** — filters items by name / title / description / type / category; all terms must match.
- **`appkit registry info <item>`** — shows an item's resource requirements plus its npm and registry dependencies (`--json` for machine output).

## Resource-aware install

Server plugins declare the Databricks resources they need. On `add`, the CLI:

- reconciles the resource env vars into `.env` (and their names into `.env.example`);
- patches the deploy config — `app.yaml` and `databricks.yml` resource bindings — to match, never clobbering existing entries;
- prompts interactively by default, with a workspace picker to resolve resource IDs; `--yes` runs non-interactively for agents/CI, `--env KEY=VALUE` (repeatable) pre-supplies values, and `--profile` runs `bundle validate` after writing;
- warns when a resource type needs a `user_api_scope` before deploy.

## Security hardening

Registry items are untrusted remote data, so `add` gates and sanitizes them:

- **Verified gate** — refuses items the index doesn't mark verified, evaluated over the *entire resolved graph* so an unverified transitive dep can't ride in on a verified item. Fails closed if the index can't be read; `--allow-unverified` overrides.
- **Identity pinning** — an item is pinned to the fetch key, not the body's self-reported `name`, so it can't claim a verified name to slip past the gate.
- **Name validation** — item refs must be plain slugs before use as a fetch path or on-disk dir (blocks `/`, `..`, control chars).
- **Path containment** — file targets are resolved within the destination dir; absolute or escaping targets are rejected (blocks arbitrary write → RCE).
- **Dependency specs** — only `[@scope/]name[@range]` specs are installed (blocks tarball/git-URL install-script RCE and flag/argument injection; `--` guard as defense in depth).
- **Code interpolation** — a plugin's export name is only spliced into the user's server source when it's a valid JS identifier.

## Tests & supporting changes

- Unit tests + fixtures (`analytics`, `lakebase`) covering `add`, `client`, `config-plan`, `config-writer`, `env-reconcile`, `env-writer`, `requirements`, `server-register`, and `workspace-picker`.
- `schemas/manifest.ts`: single source for field-origin derivation (used by `requirements`).
- `apps/scratch/`: gitignored dir for local scratch apps; pnpm workspace/lock updates.

## Paired change

The registry itself is renamed to `@databricks-appkit` in databricks/appkit-registry#1; this PR's `REGISTRY_NAMESPACE` matches it.


packages/shared/src/cli/commands/registry/server-register.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  const edits = [];
+
+  // toPlugin exports are factories, registered as a call: `hello()`.
+  const newElem = `${exportName}()`;
+
+  // Insert before the first element, matching its indentation so the array
+  // formatting is preserved (or inline for a single-line array).
+  const elementKinds = ["identifier", "call_expression", "spread_element"];
```
atilafassina, line 129: Not a blocker, you nailed 100% of the cases we currently advise and document, I'm just thinking as we get a registry and others are creating plugins, we could perhaps add `new_expression` ...?

```
┌─────────────────────────────┬────────────────────┬───────────────────────────┐
│ element written by the user │      c.kind()      │     in elementKinds?      │
├─────────────────────────────┼────────────────────┼───────────────────────────┤
│ analytics()                 │ call_expression    │ ✅ → safe (insert branch) │
├─────────────────────────────┼────────────────────┼───────────────────────────┤
│ myPlugin                    │ identifier         │ ✅ → safe                 │
├─────────────────────────────┼────────────────────┼───────────────────────────┤
│ ...basePlugins              │ spread_element     │ ✅ → safe                 │
├─────────────────────────────┼────────────────────┼───────────────────────────┤
│ new AnalyticsPlugin()       │ new_expression     │ ❌                        │
└─────────────────────────────┴────────────────────┴───────────────────────────┘
```
Besides that, we may need to document the supported element kinds somewhere for people creating plugins (but that's definitely a follow-up for once we have the registry).

atilafassina (approved):

packages/shared/src/cli/commands/registry/server-register.ts
MarioCadenas: yep we could probably add some linting rules in the registry to avoid exporting plugins without the wrapper

PR #464, opened 2026-06-26
feat(appkit): resolve schema-relative typegen queries via catalog/schema context

QuentinAmbard:
## Problem

The type-generator's DESCRIBE pass runs `DESCRIBE QUERY <sql>` with only a `warehouse_id` and **no session catalog/schema**. Queries written **schema-relative** — unqualified table names like `FROM orders` — therefore fail at typegen time with `TABLE_OR_VIEW_NOT_FOUND`, even though they resolve fine at runtime (the analytics route can pass catalog/schema as statement context). When DESCRIBE fails, the generated types silently degrade to `result: unknown`, so apps lose chart type-safety with no obvious cause.

This is increasingly common: writing queries schema-relative (and supplying catalog/schema as session context) is how an app stays portable across workspaces without hardcoding `catalog.schema.table` in every `.sql`.

### Repro

Same table, same warehouse — only difference is the session context:

```jsonc
// No catalog/schema (what typegen does today):
{ "warehouse_id": "…", "statement": "SELECT * FROM silver_returns LIMIT 1" }
// → FAILED: [TABLE_OR_VIEW_NOT_FOUND] The table or view `silver_returns` cannot be found.

// With catalog/schema (what the runtime route does):
{ "warehouse_id": "…", "catalog": "retail_consumer_goods", "schema": "luxebeauty_demo",
  "statement": "SELECT * FROM silver_returns LIMIT 1" }
// → SUCCEEDED
```

## Fix

Thread an **optional** `catalog`/`schema` through the typegen path so DESCRIBE runs with the same session context the runtime uses:

- `vite-plugin.ts` reads `DATABRICKS_CATALOG` / `DATABRICKS_SCHEMA` from env (right next to the existing `DATABRICKS_WAREHOUSE_ID`) and forwards them.
- `generateFromEntryPoint` → `generateQueriesFromDescribe` → `describeAdaptive` carry the optional context down to `executeStatement`, which honors top-level `catalog`/`schema` as the equivalent of `USE CATALOG` / `USE SCHEMA` for that one statement.

I used the native top-level `catalog`/`schema` fields rather than prepending `USE CATALOG; …` to the statement string, since the Statement Execution API already supports them and it avoids multi-statement handling.

## Backward compatibility

When the env vars are unset, **no** `catalog`/`schema` keys are added to the request and behavior is byte-for-byte unchanged. Existing behavior for fully-qualified queries is unaffected.

## Tests

Added two `describeAdaptive` tests in `statement-result.test.ts`:
- forwards `catalog`/`schema` to `executeStatement` when context is provided;
- omits both keys when no context is passed (back-compat).

Full package: `tsc --noEmit` clean, `biome check` clean, commitlint clean, **524 type-generator + analytics tests pass**.

### DIFF
--- a/packages/appkit/src/type-generator/index.ts
+++ b/packages/appkit/src/type-generator/index.ts
@@ -298,6 +298,9 @@ export async function generateFromEntryPoint(options: {
   mvOutFile?: string;
   mvMetadataOutFile?: string;
   metricFetcher?: DescribeFetcher;
+  /** Session catalog/schema for DESCRIBE — see generateQueriesFromDescribe. */
+  catalog?: string;
+  schema?: string;
 }) {
   const {
     outFile,
@@ -308,6 +311,8 @@ export async function generateFromEntryPoint(options: {
     mvOutFile,
     mvMetadataOutFile,
     metricFetcher,
+    catalog,
+    schema,
   } = options;
   const projectRoot = resolveProjectRoot(outFile);
 
@@ -320,6 +325,8 @@ export async function generateFromEntryPoint(options: {
     const result = await generateQueriesFromDescribe(queryFolder, warehouseId, {
       noCache,
       mode,
+      catalog,
+      schema,
     });
     queryRegistry = result.schemas;
     syntaxErrors = result.syntaxErrors ?? [];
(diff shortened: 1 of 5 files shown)


atilafassina:
Hey @QuentinAmbard , thanks a lot for this (and sorry for the wait!)

I think the logic is sound and (despite our CI not collaborating) tests are passing 🏆 
My main concern with merging this is that right now CATALOG and SCHEMA aren't part of the query cache.

I think we would need to add it there so there aren't stale cache entries when one of those values is added or changed.

Would you be open to making this change in this PR?

PR #469, opened 2026-07-03
chore(database): add internal schema contract and builder foundation

ditadi:
## Summary

This PR adds the first AppKit-side foundation for the database plugin.

The broader database plugin goal is to give AppKit apps a typed, Lakebase-backed database lifecycle: developers declare tables in TypeScript, AppKit derives runtime/query metadata from that schema, and later stack layers use the same contract for DDL, migrations, type generation, and DML-only runtime access.

This PR specifically:

- Adds the database contract types used by the runtime/query layer: column metadata, relation metadata, filter operators, and wire limits.
- Adds `defineSchema(...)`, table/column builders, FK refs, private/owner column markers, and Drizzle-backed engine table handles.
- Resolves FK-derived relations in one pass, including forward `toOne` and inferred reverse `toMany` relations.
- Derives Zod insert/update schemas from table metadata.
- Adds unit coverage for contract serialization, schema building, FK handling, relation inference, engine relation generation, private columns, and validators.

## Scope

This is the schema/contract foundation only. It does not add the database plugin runtime, lifecycle CLI commands, migrations, grants, or deploy orchestration; those are layered in later PRs in the stack.

## Tests

- Unit tests added under `packages/appkit/src/database/**/tests`.


packages/appkit/src/database/schema-builder/private.ts
```diff
@@ -0,0 +1,27 @@
...
+  return Object.values(table.$columns)
+    .filter((c) => !isPrivateColumn(c))
+    .map((c) => c.columnName);
+}
+
+/**
+ * The RLS owner column name (`.owner()`), if one is declared.
+ */
+export function ownerColumnName(table: AppKitTable): string | undefined {
+  return Object.values(table.$columns).find((c) => c.isOwner)?.columnName;
```
atilafassina, line 26: If I understand the flow right, this will return the first owner found. But technically multiple owners can be declared on the schema

I think if we don't want to consider multiple owners, we should throw on schema definition (or here when multiple owners are found). Or simply return all owners and allow joint ownership for RLS

wdyt?

ditadi: nice catch, I added a validation in `defineSchema` to reject multiple `.owner()` columns.

atilafassina (approved):

PR #474, opened 2026-07-10
feat: metric-view runtime

atilafassina:
Metric Views runtime support

**User-facing**
- `POST /api/analytics/metric/:key` route that measures a registered UC Metric View, mirroring the existing `/query` SSE envelope

**Internals**
- Metric-view route skeleton + measures-only SQL (`MEASURE(m) AS m`)
- Dimensions + `GROUP BY ALL`; structured recursive filter engine (12 operators, parameterized `:f_<idx>` binds, AND/OR groups with depth + cardinality caps)
- `timeGrain` bucketing via an explicit `timeDimension` (`date_trunc`)
- Lane dispatch from the entry's `executor` (SP shared-cache vs OBO per-user cache); cache-key composition + per-user identity hashing

**Manual testing**
1. Get a warehouse with valid metric-views
2. add `metric-views.json`
3. run dev-playground connected

the below query runs against a dev-playground on port `:8001` and queries a `arr` MV

```sh
curl -N -X POST http://localhost:8001/api/analytics/metric/revenue \
  -H "Content-Type: application/json" -H "Accept: text/event-stream" \
  -d '{"measures":["arr"]}'
```
<details>
<summary> example response</summary>

```
id: 721b3205-71a1-42ec-930c-f27262971929
event: warehouse_status
data: {"type":"warehouse_status","status":{"state":"RUNNING","elapsedMs":791}}

id: 1842d739-bb41-4327-9f4c-659afd827549
event: result
data: {"type":"result","data":[{"arr":"1.597096272E9"}]}
```

</details>

---

> [!NOTE]
> the headless React hook is a follow-up once the server-side is approved


docs/static/schemas/metric-source.schema.json
```diff
@@ -21,6 +21,7 @@
         "properties": {
           "source": {
             "type": "string",
+            "maxLength": 767,
```
atilafassina, line 24: max character length for FQN, derived from the sum of UC's object names

packages/appkit/src/type-generator/mv-registry/config.ts
```diff
@@ -94,26 +94,13 @@ function isValidMetricKey(key: string): boolean {
...
- * is a separate concern enforced (with its own message) in resolveMetricConfig.
- */
-export function isValidFqn(fqn: string): boolean {
-  const segments = fqn.split(".");
-  if (segments.length !== FQN_SEGMENT_COUNT) {
-    return false;
-  }
-  return segments.every((segment) => UC_FQN_PATTERN.test(segment));
-}
+// `isValidFqn` (the three-part UC FQN predicate) now lives in the shared
```
MarioCadenas: do we need this? method is gone but comment is still here?

packages/appkit/src/plugins/analytics/mv/registry.ts
```diff
@@ -0,0 +1,174 @@
+import fs from "node:fs/promises";
```
MarioCadenas: seems to me that many of this logic for reading and retrieving this data should go through the AppManager class no? that way we can centralize and it can keep working with dev-remote / HMR and also avoid repeating the same definition of QUERIES_DIR?

packages/appkit/src/plugins/analytics/analytics.ts
```diff
@@ -116,10 +125,18 @@ export class AnalyticsPlugin extends Plugin implements ToolProvider {
    */
   private _arrowCapability = new Map<string, ArrowCapability>();
 
+  /**
+   * Directory the metric-view registry is read from (holds
+   * `metric-views.json`).
+   * @default `config/queries/` (under the process cwd)
+   */
+  private readonly _queriesDir: string;
```
MarioCadenas: we already define this in AppManager, we should go through that

packages/appkit/src/plugins/analytics/types.ts
```diff
@@ -23,6 +23,17 @@ export interface IAnalyticsConfig extends BasePluginConfig {
...
+   * Directory to read the metric-view registry (`metric-views.json`) from.
+   *
+   * @internal FOR TESTS ONLY. Production always reads the fixed
+   * `config/queries/` path (the default), the same directory the `.sql` query
+   * path uses — do NOT set this to relocate config in a real app. It exists so
+   * tests can point the loader at a temp directory and exercise the real
+   * stat → read → cache path against a real file, rather than poking private
+   * plugin state or stubbing the loader.
+   */
+  queriesDir?: string;
```
MarioCadenas: I don't think we need this no?

packages/appkit/src/plugins/analytics/analytics.ts
```diff
@@ -426,6 +453,260 @@ export class AnalyticsPlugin extends Plugin implements ToolProvider {
...
+
+    if (!key) {
+      res.status(400).json({ error: "metric key is required" });
+      return;
+    }
+
+    // Resolve the registry from disk (cached + mtime-revalidated by `getMetricRegistry`).
+    let registry: Record<string, MetricRegistration>;
+    try {
+      registry = await getMetricRegistry(this._queriesDir);
```
MarioCadenas: I think this should go through AppManager as we do with queries

packages/appkit/src/plugins/analytics/mv/constants.ts
```diff
@@ -0,0 +1,97 @@
+import path from "node:path";
+import type { MetricFilterOperatorName, MetricLane } from "../types";
+
+/**
+ * Default queries directory. Mirrors `AppManager`'s
+ * `path.resolve(process.cwd(), "config/queries")` so dev mode and production
+ * share a single source of truth for where metric config lives. Exported so
+ * `AnalyticsPlugin` can default `config.queriesDir` to the same path.
+ */
+export const QUERIES_DIR = path.resolve(process.cwd(), "config/queries");
```
MarioCadenas: we should have this only in AppManager

MarioCadenas (approved):

packages/appkit/src/type-generator/mv-registry/config.ts
```diff
@@ -18,7 +18,7 @@ import type {
   ResolvedMetricEntry,
 } from "./types";
 
-const MV_CONFIG_FILE = "metric-views.json";
+const MV_CONFIG_FILE = "definitions.json";
```
MarioCadenas: we have this constant defined in `packages/appkit/src/plugins/analytics/mv/constants.ts` too, should we reuse?

packages/appkit/src/type-generator/mv-registry/config.ts
```diff
@@ -44,18 +44,19 @@ function compareKeys(a: string, b: string): number {
...
- * additive — apps without metric-views.json must not be penalized). There is
- * deliberately no fallback to the legacy `metric.json` filename.
+ * additive — apps without definitions.json must not be penalized). There is
+ * deliberately no fallback to a legacy filename.
  *
  * Throws on JSON parse errors so misconfiguration surfaces loudly.
  */
 export async function readMetricConfig(
-  queryFolder: string,
+  metricViewsFolder: string,
```
MarioCadenas, line 56: should this also use the appmanager to read the file?

packages/appkit/src/type-generator/mv-registry/config.ts
atilafassina: Good call — deduped. Hoisted a single `METRIC_CONFIG_FILE` into the shared zod-free `packages/shared/src/schemas/metric-fqn.ts` (the one module the type-generator can import without dragging zod into its locked dependency graph, and that `config.ts` already imports for `UC_FQN_PATTERN`). Repointed all four value-uses to it: `mv/constants.ts`, this file (`config.ts`), `type-generator/vite-plugin.ts`, and `shared/.../generate-types.ts`. One source of truth now.

packages/appkit/src/type-generator/mv-registry/config.ts
atilafassina: Id keep this on plain `fs` here. `readMetricConfig` runs at **build/CLI time** (the `generate-types` / `mv sync` path via `type-generator/index.ts`) — there is no HTTP request and no `DevFileReader`, so there is nothing for `AppManager` to be dev-tunnel-aware about. `AppManager` is a request-time abstraction whose whole value is the `?dev` WebSocket-tunnel branch; wiring it in here would also pull the app-runtime module into the type-generators deliberately-isolated dependency graph (same ruling that keeps this file off the shared zod schema — see the header comment) for no behavioral gain. The **runtime** read does go through `AppManager` — `plugins/analytics/mv/registry.ts` uses `readMetricViewsConfig`, which is dev-tunnel/HMR-aware. So the split is intentional: runtime → AppManager, build → fs. Happy to reconsider if you want the CLI to be tunnel-aware too, but I dont think it needs to be.

MarioCadenas (approved):

PR #475, opened 2026-07-10
feat(appkit): introduce workspace-client wrapper over the Databricks SDK

MarioCadenas:
## What

Introduces `packages/appkit/src/workspace-client/` — a facade over the legacy `@databricks/sdk-experimental` client that mirrors the modular Databricks SDK's multi-client shape. This **prepares** the library for an incremental, per-service migration to the modular SDK. **Nothing is migrated yet** — every service still delegates to the legacy client.

Supersedes the exploration in #414, which actually migrated services (files/warehouses/vectorSearch), added upstream patches, and introduced dual-error handling. This PR does only the seam.

## Why introduce a wrapper that changes nothing?

Fair question — on its own this PR adds indirection that delegates every call straight back to the same SDK. The value is entirely in what it unblocks.

AppKit is on `@databricks/sdk-experimental` (one monolithic `WorkspaceClient`). The supported SDK is [`databricks/sdk-js`](https://github.com/databricks/sdk-js) — a **modular** architecture: a `@databricks/sdk-core` + `-auth` + `-options` foundation plus one `@databricks/sdk-<service>` package per API, each constructed like `new FilesClient({ host, credentials })`. We want to move onto it, but a big-bang migration is risky (that's what #414 was, and it was too large to land safely).

So the plan is **migrate one service at a time, each in its own PR**, with the app fully working after every step. This wrapper is the seam that makes that possible:

- Every service goes through **one accessor** (`wrapper.files`, `wrapper.genie`, …), and a Biome boundary rule guarantees **nothing else in appkit imports the SDK directly**. So migrating a service is a localized change — one getter in `client.ts`, one accessor type, one connector — with a provably bounded blast radius.
- Unmigrated services keep delegating to legacy, so **main stays green between every migration PR** — no half-broken state.
- `toLegacyWorkspaceClient()` is the deliberate escape hatch for code still on the old SDK (notably `@databricks/lakebase`), so the seam can be complete even while the migration is partial.

That's why the boundary rule and escape hatch exist — they're the mechanism for a safe, incremental migration, not speculative abstraction.

### Where this goes (for context, not in this PR)

Coverage was verified 2026-07-23: **every service appkit uses now has a published modular package** (`sdk-files`, `sdk-warehouses`, `sdk-vectorsearch`, `sdk-statementexecution`, `sdk-scim` for currentUser, `sdk-modelserving` for serving, `sdk-genie`, `sdk-jobs`), so full migration and eventual removal of `sdk-experimental` is a reachable end state. Rough sequence: clean services first (files/warehouses/vectorSearch, PoC-proven), then statementExecution + the raw-`apiClient.request` sites → `sdk-core/http`, then the divergent ones last (genie/jobs need a connector rewrite for the waiter idiom + camelCase fields). The honest caveat: the wrapper makes the *wiring* trivial but does **not** shrink the per-connector rewrite — that's real, case-by-case work per service.

Full plan lives in the AppKit roadmap under Pillar 4 → "SDK migration".

## Design

- **`legacy.ts` is the only module that imports `@databricks/sdk-experimental`.** It constructs the client and re-exports the SDK symbols AppKit uses (`Context`, `ConfigError`, `Time`, `TimeUnits`, type namespaces `files`/`jobs`/`serving`/`sql`, etc.).
- **The `WorkspaceClient` facade** exposes per-service accessors (`files`, `warehouses`, `genie`, `jobs`, `statementExecution`, `servingEndpoints`, `currentUser`, `config`, `apiClient`), each legacy-typed and delegating to the underlying client. Migrating a service later is a localized swap: one getter + its accessor type + one connector — the rest of the codebase never touches the SDK.
- **`createWorkspaceClient()`** replaces every `new WorkspaceClient(...)` site (runtime + build-time type-generator). **`toLegacyWorkspaceClient()`** is the escape hatch for `@databricks/lakebase`, which is still on the old SDK.
- **`createApp({ client })`** and the public package index now expose the wrapper type instead of the raw SDK client. (Breaking at the type level for anyone passing a raw SDK client to `createApp` — no known external users.)
- **A Biome `noRestrictedImports` boundary rule** forbids importing `@databricks/sdk-experimental` outside `workspace-client/`, so the seam stays enforceable. Tests mock the wrapper rather than the SDK.

### Note on `Time`

`Time` is sourced off the SDK namespace (`SDK.Time ?? SDK.default.Time`) rather than a static `export { Time }`. The SDK's CommonJS `Time` export is emitted as a getter calling `__importDefault(...)`, which defeats Node's static ESM named-export detection — a direct re-export throws *"does not provide an export named 'Time'"* at link time. This mirrors the guard the genie connector already used.

## Verification

- `pnpm build` — clean, publint OK
- `pnpm -r typecheck` — 0 errors, all packages
- Biome boundary rule — 0 violations; `@databricks/sdk-experimental` imported only in `workspace-client/`
- `pnpm check` — 0 errors, 63 warnings (identical to `main` baseline)
- Full test suite — 3284 passed / 0 failed
- Built playground boots past agent registration + plugin setup (fails only at Databricks auth, as expected without credentials) — confirms the ESM `Time` fix end-to-end

### DIFF
--- a/packages/appkit/src/workspace-client/legacy.ts
+++ b/packages/appkit/src/workspace-client/legacy.ts
@@ -1,0 +1,94 @@
+/**
+ * The single module in AppKit allowed to import `@databricks/sdk-experimental`
+ * directly (enforced by the Biome `noRestrictedImports` boundary rule). Every
+ * other AppKit module reaches the SDK through the wrapper's re-exports and the
+ * {@link WorkspaceClient} facade.
+ *
+ * Isolating the SDK here is what makes the incremental migration to the modular
+ * Databricks SDK a localized change: to migrate a service, swap its getter in
+ * `client.ts` from the legacy delegate to the modular client — nothing else in
+ * the codebase imports the SDK, so the blast radius is one connector + its test.
+ */
+
+import type {
+  ClientOptions,
+  WorkspaceClient as SdkWorkspaceClient,
+} from "@databricks/sdk-experimental";
+import * as SDK from "@databricks/sdk-experimental";
+
+const { WorkspaceClient: SdkWorkspaceClientCtor } = SDK;
+
+/** The concrete legacy SDK client type. */
+export type LegacyWorkspaceClient = SdkWorkspaceClient;
+
+/**
+ * Options used to construct the wrapper. Mirrors the subset of the old SDK's
+ * `Config` + `ClientOptions` that AppKit relies on today; we deliberately do
+ * NOT re-expose every old-SDK config knob.
+ */
+export interface WorkspaceClientOptions {
+  /** Databricks host, e.g. https://my-workspace.cloud.databricks.com. Defaults to DATABRICKS_HOST / profile resolution. */
+  host?: string;
+  /** Bearer token. When set, `authType` defaults to "pat". */
+  token?: string;
+  /** Authentication strategy passed to the legacy client. */
+  authType?: "pat";
+  /**
+   * SDK client options (product / productVersion / userAgentExtra) used to
+   * stamp the outbound User-Agent. Produced by `getClientOptions()`; omitted
+   * by build-time callers that don't stamp a User-Agent.
+   */
+  clientOptions?: ClientOptions;
+}
+
+/**
+ * Construct a legacy `WorkspaceClient` from wrapper options.
+ *
+ * Centralised so the wrapper facade, the `.toLegacyWorkspaceClient()` escape
+ * hatch, and any per-request OBO client all build it the same way.
+ */
+export function buildLegacyWorkspaceClient(
+  opts: WorkspaceClientOptions,
+): LegacyWorkspaceClient {
+  // Check `token !== undefined`, NOT truthiness: an explicitly-passed token
+  // must stick to the PAT path even when it's an empty string. Falling through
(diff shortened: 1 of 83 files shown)


packages/appkit/src/workspace-client/legacy.ts
```diff
@@ -0,0 +1,88 @@
...
+ * hatch, and any per-request OBO client all build it the same way.
+ */
+export function buildLegacyWorkspaceClient(
+  opts: WorkspaceClientOptions,
+): LegacyWorkspaceClient {
+  const cfg = opts.token
+    ? { host: opts.host, token: opts.token, authType: opts.authType ?? "pat" }
+    : opts.host
+      ? { host: opts.host }
+      : {};
```
atilafassina: isn't this ternary executing a bad OBO request as SP?

(if OBO token is empty string `""` it will evaluate falsy and silently run as SP, while, I think what we would want is for the call to fail in production, no?

So `opts.token !== undefined` sticks with OBO and if the token is invalid, it blows up.
Or maybe we also check for env here and let it escalate in dev?? 🤔

MarioCadenas: fixed

atilafassina (approved):

PR #476, opened 2026-07-16
feat(appkit): /ui visual variant picker

IamGalymzhan:
## /ui variant selector skill

[Document](https://docs.google.com/document/d/1NYpdR0kMaafc0jHR7I4FpJj-HiUt154m8GUVV71t5LY/edit?usp=sharing)

This PR is adding a new feature of /ui-variants skill, that makes agent create several design versions of the UI component and allows user to select it live on the web page, while agent could read it on the next turn and implement selected degisn.

### How to test
0. Wire up the skill `pnpm --filter=dev-playground exec appkit skills install`
1. Use `/ui-variants` command via Isaac/Claude Code and type want you want to build in dev-playground
2. Once agent completes go to the page and select needed variants
   * You can see choices landed in the `node_modules/.databricks/appkit/appkit-ui-choices.jsonl` file
3. Tell agent that you have selected variants that you liked 
4. Confirm everything is done (No <Variants> component, selected the right design)

### How should it work for user
1. Agent skills is installed on setup (via symlink to the project's .claude/ folder) 
2. `/ui-variants` skill is available right away


.claude/commands/ui.md
```diff
@@ -0,0 +1,106 @@
+---
```
calvarjorge: Is this the place where we want to place the skill? Maybe yes, but there is a difference with the rest of the skills in the dir -- the other skills are to be used by appkit developers, while this skilled is meant to be installed and used by appkit users

calvarjorge: Do we want to name it `/ui`? Can we align on this to avoid changing it later? Maybe this is to broad, other option could be `/appkit-ui-variants`

packages/appkit-ui/src/react/ui-variants/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Variant, type VariantProps } from "./variant";
```
calvarjorge, line 1: Should we name the dir `/ui-variants` to be more specific and be consistent with the `appkit-ui` naming?

calvarjorge (commented):
Nice work! Took a first look and left some comments

packages/appkit-ui/src/react/ui-variants/index.ts
IamGalymzhan: Sure, I'll make it

.claude/commands/ui.md
IamGalymzhan: Moved the skill into the appkit itself under the skills folder, with installation via appkit cli command (`appkit skills install`)

IamGalymzhan: I am proposing a name `/ui-variants`

IamGalymzhan: It is `/appkit-ui-variants` now

packages/appkit-ui/src/react/ui-variants/variants.tsx
```diff
@@ -0,0 +1,354 @@
...
+  accent = DEFAULT_ACCENT,
+}: VariantsProps) {
+  const variants = useMemo(
+    () => Children.toArray(children).filter(isVariantElement),
+    [children],
+  );
+  const [active, setActive] = useState(0);
+  const [confirmState, setConfirmState] = useState<ConfirmState>("idle");
+
+  if (variants.length === 0) return null;
```
MarioCadenas, line 128: wdyt about a small prod safety net right after the length check (kept below the hooks):

```suggestion
  if (variants.length === 0) return null;
  if (!import.meta.env.DEV) return <>{variants[0]}</>;
```

so if a block ever ships without being finalized, users see real content instead of the picker chrome — `variants[0]` avoids the blank section that returning `null` would leave. not a security thing (the recorder is already server-gated), just belt-and-suspenders on top of the `no-variants-in-prod` lint. `import.meta.env.DEV` to match what we already do in `use-query-hmr.ts` 😄

non-blocking, just a thought

packages/appkit/src/index.ts
```diff
@@ -83,6 +83,13 @@ export type {
   ServingEndpointRegistry,
   ServingFactory,
 } from "./plugins/serving/types";
+// Hand-exported: the hyphenated `ui-variants` folder can't be an auto-export.
+export {
+  type ChoiceSink,
+  FileChoiceSink,
+  type UiChoiceRecord,
+  uiVariants,
```
MarioCadenas: let's remove the export

IamGalymzhan: done

packages/appkit-ui/src/react/ui-variants/variants.tsx
IamGalymzhan: Yes, thanks for noting that, added it

MarioCadenas (approved):
LGTM! amazing job @IamGalymzhan ! 🚀

PR #477, opened 2026-07-16
feat(appkit): mlflow tracing for agents (stack 1/5)

MarioCadenas:
Stack 1/5 · targets `main`.

Adds MLflow tracing to the agents plugin. Agent turns and tool calls are traced to a bound MLflow experiment via the `mlflow-tracing` SDK (OpenTelemetry under the hood). Tracing is a no-op unless the plugin's optional `experiment` resource is set (`MLFLOW_EXPERIMENT_ID`); auth is resolved from the app's own Databricks credentials, so no tokens or OTLP headers are wired by hand.

- `withAgentSpan` wraps each turn (AGENT span) and tool dispatch (TOOL span, auto-nested).
- Sets trace-table Request/Response previews and the `mlflow.traceName` tag.
- `experiment` optional resource in the agents manifest.

This is the base of a 5-PR stack that builds out an agent evaluation framework. Reviewable on its own — touches only the agents plugin.

### DIFF
--- a/packages/appkit/src/plugins/agents/agents.ts
+++ b/packages/appkit/src/plugins/agents/agents.ts
@@ -1302,10 +1302,8 @@ export class AgentsPlugin extends Plugin implements ToolProvider {
     const tools = Array.from(registered.toolIndex.values()).map((e) => e.def);
     const limits = this.resolvedLimits;
 
-    // Surfaced in the response envelope so an eval runner can attach
-    // assessments to this turn's trace. Set inside the span below (the only
-    // context where the active trace id is resolvable); undefined when tracing
-    // is disabled.
+    // Assigned inside the span below (the only place the active trace id
+    // resolves), read into the response envelope after.
     let mlflowTraceId: string | undefined;
 
     const runState: RunState = {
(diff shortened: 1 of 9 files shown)


packages/appkit/src/plugins/agents/schemas.ts
```diff
@@ -34,6 +34,12 @@ export const chatRequestSchema = z.object({
     ),
   threadId: z.string().optional(),
   agent: z.string().optional(),
+  /**
+   * MLflow run id to link this turn's trace to (via `mlflow.sourceRun`), set by
+   * the eval runner so each case's trace shows under the evaluation run. Capped
+   * to a run-id-shaped length since it reaches trace metadata and logs.
+   */
+  mlflowRunId: z.string().max(64).optional(),
```
atilafassina, line 42: shouldn't this also exist in the `invocationsRequestSchema` ?
If I'm following this through, there's no way to map a tool call to a MLFlow run, right?

packages/appkit/src/plugins/agents/agents.ts
```diff
@@ -276,6 +287,7 @@ export class AgentsPlugin extends Plugin implements ToolProvider {
   }
 
   async setup() {
+    await initAgentTracing();
```
atilafassina, line 290: maybe dumb question: is tracing supposed to be always-on?

MarioCadenas: its only on if an MLFLOW_EXPERIMENT_ID is provided, if not the promise resolves instantly

packages/appkit/src/plugins/agents/schemas.ts
MarioCadenas: good catch, fixed. /invocations + /responses now mirror /chat: _runAgentNonStreaming links the trace to the run via linkTraceToRun and returns the trace id as mlflow_trace_id in the response body (the non-streaming surface has no SSE channel, so it goes in the envelope instead of a metadata event). added mlflowRunId to invocationsRequestSchema + tests. so an eval runner can now group + attach assessments through the invoke surface too.

atilafassina (approved):

PR #479, opened 2026-07-16
feat(appkit): managed eval datasets + turn semantics (stack 3/5)

MarioCadenas:
Stack 3/5 · targets `pr/agent-evals-2-framework` (review after #2).

Managed datasets and explicit conversation semantics.

- **Managed datasets**: `dataset: { table }` sweeps a Databricks managed evaluation dataset (a Unity Catalog `catalog.schema.table` read via the SQL Statement Execution API — the Python datasets API needs Spark). One run per row; `t.input`/`t.expected` bind each row. `--warehouse` + auth required.
- **Turn semantics**: one-shot (single `send`), multi-turn (consecutive `send`s share a thread), and `t.reset()` for an independent turn in one test. Each dataset row gets a fresh driver so rows don't bleed.
- **Judges gate by default**: a judge miss now fails the eval (was soft); `.atLeast(n)` sets the threshold, `.soft()` opts out.

### DIFF
--- a/packages/appkit/src/evals/dataset.ts
+++ b/packages/appkit/src/evals/dataset.ts
@@ -1,0 +1,74 @@
+import type { WorkspaceClient } from "@databricks/sdk-experimental";
+import { SQLWarehouseConnector } from "../connectors";
+
+/**
+ * One row of a managed evaluation dataset. `inputs` are the kwargs passed to the
+ * agent for the turn; `expectations` (when present) is the row's ground truth /
+ * guidelines. Mirrors the `{inputs, expectations}` shape of `mlflow.genai`
+ * datasets and of the Unity Catalog table backing a managed eval dataset.
+ */
+export interface DatasetRow {
+  inputs: Record<string, unknown>;
+  expectations?: Record<string, unknown>;
+}
+
+export interface ReadEvalDatasetOptions {
+  /** Fully-qualified UC table: `catalog.schema.table`. */
+  table: string;
+  /** SQL warehouse id to run the read against. */
+  warehouseId: string;
+  /** Optional row cap. */
+  limit?: number;
+}
+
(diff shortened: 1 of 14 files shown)


packages/appkit/src/evals/dataset.ts
```diff
@@ -0,0 +1,74 @@
...
+      : "";
+  const connector = new SQLWarehouseConnector({});
+  const response = await connector.executeStatement(client, {
+    warehouse_id: options.warehouseId,
+    statement: `SELECT inputs, expectations FROM ${options.table}${limit}`,
+  });
+
+  const rows =
+    (response.result as { data?: Array<Record<string, unknown>> } | undefined)
+      ?.data ?? [];
```
atilafassina, line 68: do we want a run if there's no data?
shouldn't we break out of the evals if there's nothing to do?

e.g.:
```ts
if (rows.length === 0) {
      return {
        rows: [undefined],
        error: `dataset "${def.dataset.table}" returned no rows`,
      };
    }
```

MarioCadenas: fixed!

MarioCadenas: Good catch — fixed. An empty dataset now surfaces a `dataset "<table>" returned no rows` error (one failing result) instead of silently running a single pass with `row=undefined`, matching the other failure branches in `resolveDatasetRows`. Added a unit test in `resolve-dataset-rows.test.ts`.

Also folded in the `LIMIT` nit while here — `readEvalDataset` now ignores a non-positive limit instead of emitting invalid SQL — and fixed a couple of stale `--warehouse` references the flag rename missed (now `--warehouse-id`).

atilafassina (approved):

PR #488, opened 2026-07-23
feat(analytics): add useMetricView client experience

atilafassina:
## TL;DR

Delivers the complete Metric Views client experience on top of #512: applications can query typed Metric Views, format results from catalog metadata, and build interactive cross-filtering charts without duplicating query or presentation logic.

## Delivered

- [x] Add `useMetricView` with inferred selected-row types, metadata, loading/error state, cancellation, HMR refresh, and shared warehouse-readiness reporting.
- [x] Consolidate the Analytics SSE lifecycle used by `useMetricView` and `useAnalyticsQuery`.
- [x] Add framework-agnostic `formatValue`, `formatLabel`, `toD3Format`, and `toMetricFilter` utilities.
- [x] Add chart click and controlled-selection APIs (`onDataClick` and `selected`) for cross-filter interactions.
- [x] Add an end-to-end Metric Views playground covering metadata-driven labels, formatting, filters, charts, and table output.
- [x] Document hook usage, formatting, filters, and Plotly/ECharts integration.
- [x] Cover hook typing, SSE behavior, format/filter utilities, chart interactions, and the playground integration with focused tests.

Stacked on #512.


apps/dev-playground/client/src/routes/metric-views.route.tsx
```diff
@@ -0,0 +1,628 @@
...
+/** Inverse of {@link toItemValue} — maps the sentinels back to a selection. */
+function fromItemValue(value: string): string | null | undefined {
+  if (value === ALL) return undefined;
+  return value === NONE ? null : value;
+}
+
+/** Display label for a selected dimension value, naming the NULL case. */
+function toDisplayLabel(value: string | null): string {
+  return value === null ? NONE_LABEL : value;
+}
```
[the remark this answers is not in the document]
atilafassina: Fixed in 2226fde4. Added an `EMPTY = "__empty__"` sentinel alongside `ALL`/`NONE`, decoded back to `""` in `fromItemValue`, and gave it an `(empty)` label in `toDisplayLabel` — kept distinct from the NULL group's `(none)` since `""` compiles to `equals ''` while `null` compiles to `notSet`/`IS NULL`.

apps/dev-playground/client/src/routes/metric-views.route.tsx
```diff
@@ -0,0 +1,628 @@
...
+  const applied = appliedDimensions(selection, exclude);
+  if (applied.length === 0) return null;
+  return (
+    <Badge variant="secondary" className="gap-1">
+      <FilterIcon className="size-3" aria-hidden />
+      {applied
+        .map(
+          (dimension) => `${formatLabel(dimension)}: ${selection[dimension]}`,
+        )
+        .join(" · ")}
```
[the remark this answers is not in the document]
atilafassina: Fixed in 2226fde4 — the badge now routes the value through `toDisplayLabel`, so it matches the active-filter chip instead of showing "Region: null" next to the chip's "Region: (none)".

packages/appkit-ui/src/react/hooks/use-metric-view.ts
```diff
@@ -0,0 +1,178 @@
...
+    unpublish: unpublishWarehouseStatus,
+  } = useAnalyticsWarehousePublisher(publisherId, key);
+
+  if (!key || key.trim().length === 0) {
+    throw new Error("useMetricView: 'key' must be a non-empty string.");
+  }
+
+  // Serialize the request body from only the defined fields.
+  // enforces referencial equality before payloads
+  const payload = useMemo(() => {
```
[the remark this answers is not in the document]
atilafassina: Fixed in 9eeed3aa. Rewrote rather than just respelling — the old note had it backwards: serializing to a string is what lets `start`'s dependency check compare the request *by value* instead of by object identity, so inline `measures`/`filter` literals don't re-fire the query every render.

pkosiec (commented):
## Review: `useMetricView` client experience

Solid PR — the hook, SSE consolidation, and format/filter utilities are well-built. I verified the feature live in the dev-playground against the dogfood warehouse: labels, currency formatting, all filter paths (dropdowns, table-row, **chart-click**), multi-dimension filters, chip removal, and Clear all all work.

One thing is broken: **the time-series line chart scrambles into spaghetti after any filter** (details inline). Everything else is UX polish or API consistency.

**Findings**
- 🔴 **P1** — trend line chart renders unordered → scrambled (reproduced)
- 🟠 **P2** — every filter blanks all visuals to skeletons (flash)
- 🟠 **P2** — `useMetricView` doesn't return `warehouseStatus` (no cold-start UI)
- `nit` × 6 — charts ignore metadata format (playground), empty-key throw, `autoStart` parity, `selected` scope, leftover `console.log`, dense comments

**Third-party (not this PR, FYI):** filtering spams the console with `TypeError: Cannot read properties of undefined (reading 'disconnect')` (×10 in one session). Traced to `size-sensor` (`resizeObserver.js` `destroy()`), a transitive dep of `echarts-for-react` — not our code. The rapid chart remount-on-filter (see the skeleton-flash finding) is what surfaces it. Fixing that flash likely quiets it; otherwise a separate ticket / dep note.

Non-blocking (COMMENT). Nothing needs to gate merge except your own call on the P1.

apps/dev-playground/client/src/routes/metric-views.route.tsx
```diff
@@ -0,0 +1,652 @@
...
+
+  // Revenue by segment — also supplies the Segment dropdown's options.
+  const segment = useMetricView("revenue", {
+    measures: ARR_MEASURE,
+    dimensions: SEGMENT_DIM,
+    filter: segmentFilter,
+  });
+
+  // ARR + MRR over time — the hero trend.
+  const trend = useMetricView("revenue", {
```
pkosiec, line 322: 🔴 **P1 — trend line chart renders scrambled after any filter.**

Reproduced live: pick a region/segment or click a table row → this line chart tangles into criss-crossing lines. It sometimes looks fine unfiltered — that's luck, not correctness.

Why:
- This query has no `orderBy`, and the metric route only adds `ORDER BY` when `limit` is set. So rows come back in arbitrary order.
- ECharts draws a line in data-array order, so unordered rows zig-zag.
- The client's `sortTimeSeriesAscending` only helps if the rows are fully reversed, not just shuffled.

This is the documented happy path (`useMetricView` by time → `LineChart`), so worth fixing at more than one layer:
- **Here (must-fix):** add `orderBy: [{ field: "created_at", direction: "ASC" }]` — one line, and you already have `TABLE_ORDER_BY` as the pattern.
- **Docs:** note that time-series charts need `orderBy` on the time dimension.
- **Follow-up (pre-existing):** make `sortTimeSeriesAscending` actually sort ascending, not just reverse a descending array — fixes every consumer.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
```diff
@@ -0,0 +1,193 @@
...
+    options.limit,
+  ]);
+
+  const start = useCallback(() => {
+    abortControllerRef.current?.abort();
+
+    setLoading(true);
+    setError(null);
+    setErrorCode(null);
+    setData(null);
```
pkosiec: 🟠 **P2 — every re-query blanks all visuals to skeletons.**

`start()` calls `setData(null)` on every requery, so a cross-filter click tears down and re-skeletons all four visuals at once — a full-page flash over the warehouse round-trip. Cross-filter is the headline feature, so this is exactly where it should feel smooth.

Suggestion: stale-while-revalidate — keep the previous `data` during a refetch, show the skeleton only on first load. One caveat: clear `data` on a **key** change (old metric's rows are meaningless), but keep it on a **filter/payload** change.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    return () => {
+      abortControllerRef.current?.abort();
+      unpublishWarehouseStatus();
+    };
+  }, [start, unpublishWarehouseStatus]);
+
+  useQueryHMR(key, start);
+
+  return { data, loading, error, errorCode, metadata };
```
pkosiec: 🟠 **P2 — hook doesn't return `warehouseStatus`, so there's no cold-start affordance.**

The hook publishes warehouse status to the provider but doesn't return it, unlike `useAnalyticsQuery`. Without a `ResourceStatusProvider` mounted (the playground has none), a cold warehouse is just a long blank skeleton with no "warehouse starting…" signal. Combined with the skeleton-flash above, cold starts feel broken.

Suggestion: return `warehouseStatus` for parity, or document that a provider is required for the cold-start UX.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
```diff
@@ -0,0 +1,193 @@
...
+  >(undefined);
+  const abortControllerRef = useRef<AbortController | null>(null);
+
+  const publisherId = useId();
+  const {
+    publish: publishWarehouseStatus,
+    unpublish: unpublishWarehouseStatus,
+  } = useAnalyticsWarehousePublisher(publisherId, key);
+
+  if (!key || key.trim().length === 0) {
```
pkosiec, line 106: `nit`: empty `key` throws during render. If a metric key comes from a dropdown that's `""` for a frame, this crashes the subtree instead of flowing through the existing `error` state. Consider setting `error` and skipping the request. Matches `useAnalyticsQuery`, so it may be intentional — low priority.

apps/dev-playground/client/src/routes/metric-views.route.tsx
```diff
@@ -0,0 +1,652 @@
...
+              <VisualStatus
+                loading={region.loading}
+                error={region.error}
+                data={region.data}
+              />
+              {!region.loading &&
+                !region.error &&
+                region.data &&
+                region.data.length > 0 && (
+                  <BarChart
```
pkosiec: `nit`: charts don't apply the metadata number format. Seen live: the bar tooltip shows `849,732,624` while the table shows `$849,732,624.00` — same number, same page. The table uses `formatValue(row[col], metadata?.[col]?.format)`, but the chart gets only `data`.

`toD3Format` exists for exactly this. Since this file is the API reference, it'd be nice to show the metadata→chart wiring (thread the d3 specifier through `options`) so users discover it. Playground-only, so low priority.

packages/appkit-ui/src/react/hooks/types.ts
```diff
@@ -297,9 +298,160 @@ export type InferServingRequest<K> =
...
+  | {
+      timeGrain?: undefined;
+      timeDimension?: Extract<D[number], InferTimeDimensionKeys<K>>;
+    }
+  | {
+      timeGrain: GrainsForSelectedTimeDims<K, D>;
+      timeDimension: Extract<D[number], InferTimeDimensionKeys<K>>;
+    };
+
+export type UseMetricViewOptions<
```
pkosiec, line 434: `nit`: `useMetricView` has no `autoStart` option, but `useAnalyticsQuery` does. The two hooks read as twins, so the asymmetry surprises. Consider adding `autoStart` (and maybe a returned `refetch`) for parity, or noting the omission in the docs.

packages/appkit-ui/src/react/charts/base.tsx
```diff
@@ -168,6 +171,23 @@ export interface BaseChartProps {
...
+   * the same action.
+   */
+  onDataClick?: (datum: ChartClickDatum) => void;
+  /**
+   * Controlled selection by category name. Matching data element(s) render at full
+   * prominence while the rest are dimmed. Drive it from your own state to reflect a
+   * cross-filter or selection. Categorical charts (bar, pie/donut) show emphasis;
+   * other chart types ignore it.
+   */
+  selected?: string | string[];
```
pkosiec, line 189: `nit`: `selected` emphasis only applies to bar/pie/donut; line/area/scatter/heatmap/radar silently ignore it. It's documented on the prop — just worth a line in the chart-integration docs so users don't wire `selected` to a line chart and wonder why nothing dims.

apps/dev-playground/client/src/routes/metric-views.route.tsx
```diff
@@ -0,0 +1,652 @@
...
+              trend.data &&
+              trend.data.length > 0 && (
+                <LineChart
+                  data={trend.data}
+                  xKey="created_at"
+                  yKey={[...TREND_MEASURES]}
+                  height={320}
+                  showLegend
+                  onDataClick={(datum) => {
+                    console.log("[Metric Views] Line chart clicked", datum);
```
pkosiec: `nit`: leftover `console.log` in the reference app. Since this file is the copy-paste template, replace it with a real handler or a commented-out `// TODO: cross-filter here` showing the intended wiring (the other charts show the pattern).

packages/appkit/src/plugins/analytics/mv/cache.ts
```diff
@@ -12,6 +12,24 @@ export function composeMetricCacheKey(input: MetricCacheKeyInput): string[] {
   // `timeDimension` only changes the SQL when `timeGrain` is set (see renderDimensionClause)
   const timeDimensionPart =
     input.timeGrain != null ? (input.timeDimension ?? "_") : "_";
+  // `orderBy` is NOT sorted (unlike measures/dimensions) because the sequence is
```
pkosiec: `nit`: comments are dense in a few spots — worth a trim pass.

The intent-comments are good, but some run long:
- `cache.ts` here — 12 lines explaining one `orderByPart` ternary.
- `metric-filter/index.ts` — ~54% comments.
- `use-metric-view.ts:98` and a few chart blocks are essay-length.

These read more like design docs than code comments. Suggestion: keep the "why" (e.g. "orderBy not sorted — sequence is semantic under LIMIT") in 1-2 lines and drop the worked-through reasoning. Leaner diff, same signal. Public-API JSDoc with examples is fine to keep.

pkosiec (commented):
## Review: `useMetricView` client experience

Solid PR — the hook, SSE consolidation, and format/filter utilities are well-built. I verified the feature live in the dev-playground against the dogfood warehouse: labels, currency formatting, all filter paths (dropdowns, table-row, **chart-click**), multi-dimension filters, chip removal, and Clear all all work.

One thing is broken: **the time-series line chart scrambles into spaghetti after any filter** (details inline). Everything else is UX polish or API consistency.

**Findings**
- 🔴 **P1** — trend line chart renders unordered → scrambled (reproduced)
- 🟠 **P2** — every filter blanks all visuals to skeletons (flash)
- 🟠 **P2** — `useMetricView` doesn't return `warehouseStatus` (no cold-start UI)
- `nit` × 6 — charts ignore metadata format (playground), empty-key throw, `autoStart` parity, `selected` scope, leftover `console.log`, dense comments

**Third-party (not this PR, FYI):** filtering spams the console with `TypeError: Cannot read properties of undefined (reading 'disconnect')` (×10 in one session). Traced to `size-sensor` (`resizeObserver.js` `destroy()`), a transitive dep of `echarts-for-react` — not our code. The rapid chart remount-on-filter (see the skeleton-flash finding) is what surfaces it. Fixing that flash likely quiets it; otherwise a separate ticket / dep note.

Non-blocking (COMMENT). Nothing needs to gate merge except your own call on the P1.

apps/dev-playground/client/src/routes/metric-views.route.tsx
pkosiec, line 322: 🔴 **P1 — trend line chart intermittently renders as scrambled spaghetti.**

Reproduced live (screenshots in thread). It's **non-deterministic** — a coin flip per cached result, not reliable on demand. That unpredictability *is* the bug.

**Mechanism (all three confirmed):**
1. This query sets no `orderBy`, and the metric route only emits SQL `ORDER BY` when `limit` is set. So rows come back in **arbitrary order** — verified against the live route: `created_at` is non-monotonic for both filtered and unfiltered queries.
2. ECharts draws a line in data-array order, so unordered rows zig-zag.
3. The client's `sortTimeSeriesAscending` (charts/utils.ts:327) only sorts when `first > last` (first row's timestamp is later than the last row's). If that guard happens to fire, you get a full ascending sort → clean chart. If it doesn't, the shuffled rows render as-is → spaghetti. Whether it fires depends purely on the arbitrary arrival/caching order → ~coin flip.

**Repro note:** you can't force it from the UI reliably — flip a filter and reload a few times, or clear the metric cache, and the same query alternates clean vs scrambled as the cached row order changes.

This is the documented happy path (`useMetricView` by time → `LineChart`), so worth fixing at more than one layer:
- **Here (must-fix):** add `orderBy: [{ field: "created_at", direction: "ASC" }]` — one line, and you already have `TABLE_ORDER_BY` as the pattern.
- **Docs:** note that time-series charts need `orderBy` on the time dimension (the route doesn't order by default without `limit`).
- **Follow-up (pre-existing):** `sortTimeSeriesAscending` should always sort ascending by x, not only when `first > last`. Right now a shuffled series that isn't fully descending is left untouched — fixing this hardens every consumer, not just this demo.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
pkosiec: 🟠 **P2 — every re-query blanks all visuals to skeletons.**

`start()` calls `setData(null)` on every requery, so a cross-filter click tears down and re-skeletons all four visuals at once — a full-page flash over the warehouse round-trip. Cross-filter is the headline feature, so this is exactly where it should feel smooth.

Suggestion: stale-while-revalidate — keep the previous `data` during a refetch, show the skeleton only on first load. One caveat: clear `data` on a **key** change (old metric's rows are meaningless), but keep it on a **filter/payload** change.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
pkosiec: 🟠 **P2 — hook doesn't return `warehouseStatus`, so there's no cold-start affordance.**

The hook publishes warehouse status to the provider but doesn't return it, unlike `useAnalyticsQuery`. Without a `ResourceStatusProvider` mounted (the playground has none), a cold warehouse is just a long blank skeleton with no "warehouse starting…" signal. Combined with the skeleton-flash above, cold starts feel broken.

Suggestion: return `warehouseStatus` for parity, or document that a provider is required for the cold-start UX.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
pkosiec, line 106: `nit`: empty `key` throws during render. If a metric key comes from a dropdown that's `""` for a frame, this crashes the subtree instead of flowing through the existing `error` state. Consider setting `error` and skipping the request. Matches `useAnalyticsQuery`, so it may be intentional — low priority.

apps/dev-playground/client/src/routes/metric-views.route.tsx
pkosiec: `nit`: charts don't apply the metadata number format. Seen live: the bar tooltip shows `849,732,624` while the table shows `$849,732,624.00` — same number, same page. The table uses `formatValue(row[col], metadata?.[col]?.format)`, but the chart gets only `data`.

`toD3Format` exists for exactly this. Since this file is the API reference, it'd be nice to show the metadata→chart wiring (thread the d3 specifier through `options`) so users discover it. Playground-only, so low priority.

packages/appkit-ui/src/react/hooks/types.ts
pkosiec, line 434: `nit`: `useMetricView` has no `autoStart` option, but `useAnalyticsQuery` does. The two hooks read as twins, so the asymmetry surprises. Consider adding `autoStart` (and maybe a returned `refetch`) for parity, or noting the omission in the docs.

packages/appkit-ui/src/react/charts/base.tsx
pkosiec, line 189: `nit`: `selected` emphasis only applies to bar/pie/donut; line/area/scatter/heatmap/radar silently ignore it. It's documented on the prop — just worth a line in the chart-integration docs so users don't wire `selected` to a line chart and wonder why nothing dims.

apps/dev-playground/client/src/routes/metric-views.route.tsx
pkosiec: `nit`: leftover `console.log` in the reference app. Since this file is the copy-paste template, replace it with a real handler or a commented-out `// TODO: cross-filter here` showing the intended wiring (the other charts show the pattern).

packages/appkit/src/plugins/analytics/mv/cache.ts
pkosiec: `nit`: comments are dense in a few spots — worth a trim pass.

The intent-comments are good, but some run long:
- `cache.ts` here — 12 lines explaining one `orderByPart` ternary.
- `metric-filter/index.ts` — ~54% comments.
- `use-metric-view.ts:98` and a few chart blocks are essay-length.

These read more like design docs than code comments. Suggestion: keep the "why" (e.g. "orderBy not sorted — sequence is semantic under LIMIT") in 1-2 lines and drop the worked-through reasoning. Leaner diff, same signal. Public-API JSDoc with examples is fine to keep.

apps/dev-playground/client/src/routes/metric-views.route.tsx
pkosiec: Here's how the graph looks like for me: 
image
image

apps/dev-playground/server/index.ts
```diff
@@ -378,7 +384,7 @@ createApp({
     server(),
     reconnect(),
     telemetryExamples(),
-    analytics({}),
+    analytics({ metricViewsMetadata }),
```
MarioCadenas: could this be internally handled so user doesn't have to pass it?

apps/dev-playground/shared/appkit-types/metric-views.d.ts
```diff
@@ -1,6 +1,6 @@
 // Auto-generated by AppKit - DO NOT EDIT
 // Generated by 'npx @databricks/appkit generate-types' or Vite plugin during build
-import "@databricks/appkit-ui/react";
+import type {} from "@databricks/appkit-ui/react";
```
MarioCadenas: can we not mix types and runtime code in the same file and keep the .d.ts? I'd prefer having them separated

packages/appkit/src/plugins/analytics/mv/constants.ts
```diff
@@ -1,11 +1,24 @@
...
+// type union). Re-exported here so analytics-local callers (`schemas.ts`,
+// `formatters.ts`) keep importing operators + subsets from this barrel.
+export {
+  LIST_VALUE_OPERATORS,
+  METRIC_FILTER_OPERATORS,
+  METRIC_ORDER_DIRECTIONS,
+  NULL_OPERATORS,
+  SINGLE_VALUE_OPERATORS,
+  STRING_OPERATORS,
+} from "shared";
```
MarioCadenas: does this need to be in the `shared` package? I don't see the UI using any of these constants, we can probably just keep it inside the plugin no?

apps/dev-playground/client/src/routes/metric-views.route.tsx
atilafassina: Fixed across all three layers you named.

- **Playground `orderBy`** — `metric-views.route.tsx:53,328`: `TREND_ORDER_BY = [{ field: "created_at", direction: "ASC" }]`, using the `TABLE_ORDER_BY` pattern.
- **Docs** — `docs/docs/plugins/analytics.md` now states that time-series queries should explicitly order their time dimension ascending, and that SQL result order is otherwise unspecified.
- **`sortTimeSeriesAscending`** — `charts/utils.ts:337` rewritten as a decorated index sort: it now always sorts ascending by x instead of only reversing a fully-descending array.

Your diagnosis also turned up a fourth layer you had not seen. The `first > last` guard was not the only gate: `normalize.ts` decided `chartType` via an `isDateString` that accepted only the ISO `T` separator, while the sorter accepted a space. Spark JSON_ARRAY emits `2025-01-01 00:00:00`, so those rows were classified `categorical` and the sort was never reached at all. Both predicates now share one parse-based helper (`toChronologicalValue`), and there is a `normalizeChartData`-level test for the space-separated case — the previous test called the sorter directly and so passed straight over this.

atilafassina: Fixed across all three layers you named.

- **Playground `orderBy`** — `metric-views.route.tsx:53,328`: `TREND_ORDER_BY = [{ field: "created_at", direction: "ASC" }]`, using the `TABLE_ORDER_BY` pattern.
- **Docs** — `docs/docs/plugins/analytics.md` now states that time-series queries should explicitly order their time dimension ascending, and that SQL result order is otherwise unspecified.
- **`sortTimeSeriesAscending`** — `charts/utils.ts:337` rewritten as a decorated index sort: it now always sorts ascending by x instead of only reversing a fully-descending array.

Your diagnosis also turned up a fourth layer you had not seen. The `first > last` guard was not the only gate: `normalize.ts` decided `chartType` via an `isDateString` that accepted only the ISO `T` separator, while the sorter accepted a space. Spark JSON_ARRAY emits `2025-01-01 00:00:00`, so those rows were classified `categorical` and the sort was never reached at all. Both predicates now share one parse-based helper (`toChronologicalValue`), and there is a `normalizeChartData`-level test for the space-separated case — the previous test called the sorter directly and so passed straight over this.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
atilafassina: Fixed with stale-while-revalidate, including the caveat you flagged.

`use-metric-view.ts:142-152`: state is now a single `{ shape, data, metadata }` object where `shape` is the serialized metric key + measures + dimensions. During a refetch that keeps the same shape, the previous `data`/`metadata` stay visible and only `loading` flips — no skeleton teardown. Changing the key or the selected columns clears the result, so rows are never shown under a mismatched result type.

Your caveat is the exact seam: filter, `orderBy`, `limit`, and `timeGrain` change *which rows* come back without changing their *fields*, so they revalidate in place; key and column changes reset.

atilafassina: Fixed with stale-while-revalidate, including the caveat you flagged.

`use-metric-view.ts:142-152`: state is now a single `{ shape, data, metadata }` object where `shape` is the serialized metric key + measures + dimensions. During a refetch that keeps the same shape, the previous `data`/`metadata` stay visible and only `loading` flips — no skeleton teardown. Changing the key or the selected columns clears the result, so rows are never shown under a mismatched result type.

Your caveat is the exact seam: filter, `orderBy`, `limit`, and `timeGrain` change *which rows* come back without changing their *fields*, so they revalidate in place; key and column changes reset.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
atilafassina: Fixed — the hook now returns it, matching `useAnalyticsQuery`.

`use-metric-view.ts:227` returns `warehouseStatus`, and it is declared on `UseMetricViewResult` (`hooks/types.ts:463`) as `WarehouseStatus | null`, `null` until the current request emits a status event. `docs/docs/plugins/analytics.md` documents that both hooks expose it and that it resets to `null` when a request starts.

No `ResourceStatusProvider` is required for the affordance now — a consumer can render the status directly from the hook.

atilafassina: Fixed — the hook now returns it, matching `useAnalyticsQuery`.

`use-metric-view.ts:227` returns `warehouseStatus`, and it is declared on `UseMetricViewResult` (`hooks/types.ts:463`) as `WarehouseStatus | null`, `null` until the current request emits a status event. `docs/docs/plugins/analytics.md` documents that both hooks expose it and that it resets to `null` when a request starts.

No `ResourceStatusProvider` is required for the affordance now — a consumer can render the status directly from the hook.

packages/appkit-ui/src/react/hooks/use-metric-view.ts
atilafassina: Fixed — `use-metric-view.ts:104` now guards `!key || key.trim().length === 0` and routes it through the existing `error` state instead of throwing during render, so a dropdown that is `""` for a frame no longer takes down the subtree.

atilafassina: Fixed — `use-metric-view.ts:104` now guards `!key || key.trim().length === 0` and routes it through the existing `error` state instead of throwing during render, so a dropdown that is `""` for a frame no longer takes down the subtree.

packages/appkit-ui/src/react/hooks/types.ts
atilafassina: Added — `autoStart` is now on `UseMetricViewOptions` (`hooks/types.ts:448`, `/** Whether to automatically start the metric query. Default is true. */`), so the two hooks read as twins.

atilafassina: Added — `autoStart` is now on `UseMetricViewOptions` (`hooks/types.ts:448`, `/** Whether to automatically start the metric query. Default is true. */`), so the two hooks read as twins.

packages/appkit-ui/src/react/charts/base.tsx
atilafassina: Documented — `docs/docs/plugins/analytics.md:679`: "The `selected` prop adds declarative emphasis only to bar, pie, and donut charts. Line, area, scatter, heatmap, and radar charts ignore it because category-selection semantics are not defined for those chart types." Names all five ignoring types explicitly so nobody wires it to a line chart and waits for a dim.

atilafassina: Documented — `docs/docs/plugins/analytics.md:679`: "The `selected` prop adds declarative emphasis only to bar, pie, and donut charts. Line, area, scatter, heatmap, and radar charts ignore it because category-selection semantics are not defined for those chart types." Names all five ignoring types explicitly so nobody wires it to a line chart and waits for a dim.

apps/dev-playground/client/src/routes/metric-views.route.tsx
atilafassina: Fixed in `bc4f27aa` — the `console.log` is gone, replaced by a documented placeholder (`metric-views.route.tsx:567-570`):

```tsx
// Seam for write-back: `datum` carries the clicked points
// measure, x value and series, which is what an edit dialog
// would need to write a correction to the source table.
// TODO: build the write-back flow on top of this handler.
onDataClick={() => {}}
```

Took your second option rather than a real handler: the two bar charts already demonstrate cross-filtering, and this seam is intentionally reserved for write-back.

The empty body is deliberate and load-bearing — `BaseChart` derives `interactive` from handler *presence* (`charts/base.tsx:237`, `const interactive = !!onDataClick`), so dropping the prop entirely would switch off line-stroke click emission and the seam with it.

Tracing this also explained why the log was there: `dbef8f84` added a `notify` toast export plus a Write-back demo, then `b9be4f10` removed "the unused public notify export and the fake Write back demo" and left this handler behind. So it was residue from that removal, not from the original build.

atilafassina: Fixed in `bc4f27aa` — the `console.log` is gone, replaced by a documented placeholder (`metric-views.route.tsx:567-570`):

```tsx
// Seam for write-back: `datum` carries the clicked points
// measure, x value and series, which is what an edit dialog
// would need to write a correction to the source table.
// TODO: build the write-back flow on top of this handler.
onDataClick={() => {}}
```

Took your second option rather than a real handler: the two bar charts already demonstrate cross-filtering, and this seam is intentionally reserved for write-back.

The empty body is deliberate and load-bearing — `BaseChart` derives `interactive` from handler *presence* (`charts/base.tsx:237`, `const interactive = !!onDataClick`), so dropping the prop entirely would switch off line-stroke click emission and the seam with it.

Tracing this also explained why the log was there: `dbef8f84` added a `notify` toast export plus a Write-back demo, then `b9be4f10` removed "the unused public notify export and the fake Write back demo" and left this handler behind. So it was residue from that removal, not from the original build.

apps/dev-playground/server/index.ts
atilafassina: Yes — done in `4fd8f839`. The playground is back to `analytics({})` (`server/index.ts:381`) and the import is gone.

You were right that this had no business being the app's job. The plugin already discovers metric views from `config/metric-views/definitions.json` with zero wiring, so display metadata — same data, same generator pass — was taking a second, redundant channel into the same plugin.

The generator now writes the runtime half to `config/metric-views/metadata.generated.json`, beside the hand-authored `definitions.json` (separate file, because a generator writing into a user-authored one would clobber hand edits). The route reads it through `AppManager.readMetricViewsConfig`, inheriting the dev tunnel and the traversal guard, and caches the parse keyed on raw contents so a regenerated bundle is picked up without a restart.

`metricViewsMetadata` survives as an explicit override that wins over the discovered bundle, so this is not breaking and an app building metadata another way keeps working.

Every failure mode degrades rather than throws — metadata is pure response decoration, so an absent, malformed, or version-mismatched bundle yields unlabeled columns instead of failing a query that would otherwise have succeeded.

apps/dev-playground/shared/appkit-types/metric-views.d.ts
atilafassina: Agreed and done in `4fd8f839` — it is a `.d.ts` again, with zero runtime exports (`grep -c "export const"` → 0).

Worth recording the root cause, because your two comments were the same bug. The file could not be a `.d.ts` while it carried the `metricViewsMetadata` object literal: an object-literal `const` in an ambient declaration is a **TS1254** error. (The code comments claimed TS1039 — I probed it, and the actual error is TS1254, and it fires *only* because of the object literal; an augmentation-only `.d.ts` compiles fine.) That one format choice cascaded into the prop you flagged above, the `.d.ts`→`.ts` rename, the `import type {}` oddity, and a special case in the `--wait` committed-types gate.

Moving the metadata to JSON dissolved all of it: the artifact is declaration-only, typegen no longer has to reject a `.d.ts` `mvOutFile` (`vite-plugin.ts`), the stale-sibling sweep inverted to clean up a leftover `.ts`, and the fallback gate lost its special case.

Verified against the built bundle rather than just source, since unbundle mode has dropped exports before: `METRIC_TYPES_FILE = "metric-views.d.ts"` and the `METRIC_METADATA_FILE` write are both present in `dist/`.

packages/appkit/src/plugins/analytics/mv/constants.ts
atilafassina: Correct — moved back into the plugin in `a99a16b3`. `mv/constants.ts` now defines the runtime tuples and sets locally (`METRIC_FILTER_OPERATORS`, `LIST_VALUE_OPERATORS`, `NULL_OPERATORS`, `STRING_OPERATORS`, `SINGLE_VALUE_OPERATORS`, `METRIC_ORDER_DIRECTIONS`) and imports only the *types* from `shared`.

Your premise checked out: the UI never consumed those runtime values, only the type union.

The local constants are declared `as const satisfies readonly MetricFilterOperatorName[]`, so drift from the shared wire contract is a compile error rather than a silent divergence, and there is a type-parity test (`tests/types.test.ts`) asserting `SharedMetricFilterOperatorName` equals `(typeof METRIC_FILTER_OPERATORS)[number]` in both directions.

apps/dev-playground/client/src/routes/metric-views.route.tsx
atilafassina: Closed — this was fixed after your review, in `0da50eba` ("polish metric view query experience"). All three charts now thread the metadata format spec through `valueFormatter`:

```tsx
valueFormatter={(value, field) =>
  formatValue(value, region.metadata?.[field]?.format)
}
```

at `metric-views.route.tsx:485` (bar), `:525` (donut), `:564` (line) — same `metadata?.[col]?.format` lookup the table uses, so the mismatch you saw live (`849,732,624` in the tooltip vs `$849,732,624.00` in the table) is gone.

I traced it through the core to confirm it actually reaches the surfaces you named, rather than just being passed in:
- bar series tooltip — `charts/options.ts:284` → `tooltipValueFormatter(ctx.valueFormatter, key)`
- value-axis labels — `options.ts:105` `valueAxisLabel`, so axis ticks format too
- pie/donut tooltip — `options.ts:195`; line series tooltip — `options.ts:471`

On the `toD3Format` suggestion: I did not add it here. `valueFormatter` already covers ECharts tooltips and axes, and `toD3Format` exists for consumers that need a d3 `specifier` + currency `prefix` (Plotly, custom d3) rather than a callback. The Plotly example in `docs/docs/plugins/analytics.md` is where that wiring is demonstrated, which seemed the better home for it than duplicating a second formatting path in this route.

atilafassina: Closed — this was fixed after your review, in `0da50eba` ("polish metric view query experience"). All three charts now thread the metadata format spec through `valueFormatter`:

```tsx
valueFormatter={(value, field) =>
  formatValue(value, region.metadata?.[field]?.format)
}
```

at `metric-views.route.tsx:485` (bar), `:525` (donut), `:564` (line) — same `metadata?.[col]?.format` lookup the table uses, so the mismatch you saw live (`849,732,624` in the tooltip vs `$849,732,624.00` in the table) is gone.

I traced it through the core to confirm it actually reaches the surfaces you named, rather than just being passed in:
- bar series tooltip — `charts/options.ts:284` → `tooltipValueFormatter(ctx.valueFormatter, key)`
- value-axis labels — `options.ts:105` `valueAxisLabel`, so axis ticks format too
- pie/donut tooltip — `options.ts:195`; line series tooltip — `options.ts:471`

On the `toD3Format` suggestion: I did not add it here. `valueFormatter` already covers ECharts tooltips and axes, and `toD3Format` exists for consumers that need a d3 `specifier` + currency `prefix` (Plotly, custom d3) rather than a callback. The Plotly example in `docs/docs/plugins/analytics.md` is where that wiring is demonstrated, which seemed the better home for it than duplicating a second formatting path in this route.

packages/appkit/src/plugins/analytics/mv/cache.ts
atilafassina: Done a trim pass on your standard — keep the "why" in 1-2 lines, drop the worked-through reasoning.

**`mv/cache.ts`** — 7 comment lines → 2. This block is now the single line you suggested almost verbatim:
```ts
// Preserve `orderBy` sequence: under LIMIT, `a, b` and `b, a` return different rows.
```
I kept that sentence because it is the load-bearing correctness rule (a collision there serves wrongly-ordered cached rows) and dropped the paragraph explaining why we tolerate extra entries rather than duplicating `renderOrderByClause`. Same for the JSON-encoding comment below it: 2 lines → 1.

**`use-metric-view.ts`** — 8 comment lines → 4, across the payload memo, the `resultShape` block, and the shape-transition line. Kept the semantic distinction (shape change clears rows, filter/order/limit/grain revalidations keep them); dropped the render-timing walkthrough.

**`metric-filter/index.ts`** — left at 54% deliberately. Reading it, the density is public-API JSDoc with `@example` blocks, which your comment explicitly exempted. There is very little prose there that is not the documented contract, so trimming it would cost real API docs. Happy to revisit if you meant something narrower in that file.

Only comment lines changed — verified with `git diff -U0` filtered to non-comment lines (empty). appkit 3146 pass, appkit-ui 529 pass, typecheck clean.

atilafassina: Done a trim pass on your standard — keep the "why" in 1-2 lines, drop the worked-through reasoning.

**`mv/cache.ts`** — 7 comment lines → 2. This block is now the single line you suggested almost verbatim:
```ts
// Preserve `orderBy` sequence: under LIMIT, `a, b` and `b, a` return different rows.
```
I kept that sentence because it is the load-bearing correctness rule (a collision there serves wrongly-ordered cached rows) and dropped the paragraph explaining why we tolerate extra entries rather than duplicating `renderOrderByClause`. Same for the JSON-encoding comment below it: 2 lines → 1.

**`use-metric-view.ts`** — 8 comment lines → 4, across the payload memo, the `resultShape` block, and the shape-transition line. Kept the semantic distinction (shape change clears rows, filter/order/limit/grain revalidations keep them); dropped the render-timing walkthrough.

**`metric-filter/index.ts`** — left at 54% deliberately. Reading it, the density is public-API JSDoc with `@example` blocks, which your comment explicitly exempted. There is very little prose there that is not the documented contract, so trimming it would cost real API docs. Happy to revisit if you meant something narrower in that file.

Only comment lines changed — verified with `git diff -U0` filtered to non-comment lines (empty). appkit 3146 pass, appkit-ui 529 pass, typecheck clean.

packages/appkit-ui/src/react/hooks/index.ts
```diff
@@ -7,19 +7,34 @@ export {
...
+  PickMetricRow,
   PluginRegistry,
   QueryRegistry,
   ServingAlias,
   ServingEndpointRegistry,
   TypedArrowTable,
   UseAnalyticsQueryOptions,
   UseAnalyticsQueryResult,
+  UseMetricViewOptions,
+  UseMetricViewResult,
```
pkosiec, line 38: **`MetricViewColumnDisplay` isn't exported from appkit-ui.**

`useMetricView` returns `metadata: Record<string, MetricViewColumnDisplay>`, but that type only lives in the workspace-only `shared` package (not published). A consumer writing a typed formatter or column renderer over `metadata` has no importable name for it.

`UseMetricViewResult` / `UseMetricViewOptions` are re-exported here — the element type should be too.

Fix: add `MetricViewColumnDisplay` (and `MetricViewsMetadata`) to this `export type { … }` block.

pkosiec (approved):
LGTM, just one comment from an agentic review - PTAL before merge. Thanks!

MarioCadenas (approved):
🚀

packages/appkit-ui/src/react/hooks/index.ts
atilafassina: Fixed in dd95134d.

`MetricViewColumnDisplay` is now re-exported from `js/format/index.ts` (next to the sibling `Metric*` re-exports in `js/metric-filter/index.ts`), so it flows through the `js` barrel, and it's surfaced on the React side via the existing `@/js` block in `hooks/types.ts` plus this export list.

Worth noting the case was stronger than the `metadata` field alone: the public `formatLabel(name, columnMeta?)` in `js/format/index.ts` already took the type as a parameter, so consumers couldn't even wrap a function we ship without a name for it.

One deviation — I left `MetricViewsMetadata` out. It doesn't appear anywhere in `packages/appkit-ui/src/`, and no `appkit-ui` API produces or accepts it (the hook flattens the bundle and returns only the per-column slice), so exporting it from here would advertise a shape this package never hands back. Happy to add it if you'd rather have it for symmetry, but it felt like it belongs on the `appkit` side if anywhere.

Verified: name present in the built `dist/js/index.d.ts` and `dist/react/index.d.ts`, `attw` + `publint` clean, and a simulated consumer import (wrapping `formatLabel`, naming `Record<string, MetricViewColumnDisplay>`, assigning `UseMetricViewResult["metadata"]`) typechecks against the built `.d.ts`.

PR #489, opened 2026-07-23
feat(cli): add appkit doctor command

IamGalymzhan:
## appkit doctor - cli command that checks connectivity to the Databricks resources

The command does four level of checks:
1. auth - validates `DATABRICKS_HOST` and authenticates to the workspace
2. config - validates that each needed environmental variable is present (e.g. warehouse_id, genie_space_id) in .env or specified by --env-file <path>
3. existence - checks that given resource exists and reachable by read requests ( .get() or SELECT 1) to check connection
4. wiring - checks that resources are declared correctly across `appkit.plugings.json` &harr; `apps.yaml` &harr; `databricks.yml` 

Output is a list of resources and problems if any with hints. Also there is options --json for machine readable output. and --detail for more detailed overview with error itself. 

### Examples:

**Failed auth**

image

**Genie Space is not provided, Job is not accessible or doesn't exists**

image

**Files are not added to the app.yaml**

image

**Everything is setup correctly**

image

### How it works:
1. Resolve all resources needed for the app from `appkit.plugins.json`
2. Auth: calls currentUser.me() to check that auth is fine
3. Config: for each resource declared checks that corresponding environmental variable is present
4. Existence: for each resource declared tries to connect and check if resource exists / has permission to access
5. Wiring: checks that all required resources are declared correctly across appkit.plugins.json, databricks.yml and apps.yaml
5. Print a report for user or for the agent with --json option


calvarjorge (commented):
Great work, looks very nice! Some questions about how it works:

- For auth, what error do we print if no profile is set? Do we prioritize profile or host? (I think both are ways to point to a workspace)
- For env vars, we should somehow differentiate where the var is missing and the impact. In your example screenshot, it says: `DATABRICKS_GENIE_SPACE_ID is not set`. We should probably clarify it refers to the env file, as oposed to `app.yaml` or `databricks.yml`. 
- How is the env determined? Do we just expect the user to have `.env`, or we have a flexible way to determine which to use? 
- For the job, where permissions are missing, can we easily add more info about it? For example, adding the name can help identify it quicker than just the numeric id. Also, can we provide some instructions about how to fix it (ie, how to add the permissions)? is in the UI (then maybe we could even add a link) or in the databricks.yml?
- Re. copy in "Files are not added to the app.yaml", I think saying "it won't be set in the environment of the deployed app" would be more clear.

IamGalymzhan:
I have done few fixes

> * For auth, what error do we print if no profile is set? Do we prioritize profile or host? (I think both are ways to point to a workspace)

Appkit doctor makes requests to the host if it is set with token of profile, on a mismatch of profile and .env host gives a warn. 

> * For env vars, we should somehow differentiate where the var is missing and the impact. In your example screenshot, it says: `DATABRICKS_GENIE_SPACE_ID is not set`. We should probably clarify it refers to the env file, as oposed to `app.yaml` or `databricks.yml`.

Done

> * How is the env determined? Do we just expect the user to have `.env`, or we have a flexible way to determine which to use?

There is --env-file <path> option for doctor to override default .env 

> * For the job, where permissions are missing, can we easily add more info about it? For example, adding the name can help identify it quicker than just the numeric id. Also, can we provide some instructions about how to fix it (ie, how to add the permissions)? is in the UI (then maybe we could even add a link) or in the databricks.yml?

There is no way no distinguish between no permission and job doesn't exists errors, so now it just gives a generic error doesn't exist or you don't have permission. 

> * Re. copy in "Files are not added to the app.yaml", I think saying "it won't be set in the environment of the deployed app" would be more clear.

Done

packages/shared/src/cli/commands/doctor/run.ts
```diff
@@ -0,0 +1,212 @@
...
+  configCtx: ConfigCheckContext,
+): Promise<ResourceCheckResult> {
+  const layers: LayerResult[] = [];
+  let rolled: CheckStatus = "ok";
+
+  const configResult = checkConfig(target, configCtx);
+  layers.push(configResult);
+  rolled = worst(rolled, configResult.status);
+  // A hard config failure makes the existence probe meaningless.
+  if (configResult.status === "error") {
```
MarioCadenas, line 75: think theres a bug here for bundle-managed resources. the config check + early return at 48-49 runs before the `bundle-managed` branch at 54, so if the env var isnt set locally (normal — the resource is created on deploy) we return error and never hit the bundle-managed skip. then report.ts printResourceRow (146) collapses it to "⧗ will be created on deploy" and drops the error layer, so you get a green-looking row but summary says "1 error" and exit 1.

repro on a bundle-managed warehouse with DATABRICKS_WAREHOUSE_ID unset:
```
  ✓  Auth — authenticated as ...
  ⧗  SQL Warehouse  will be created on deploy
1 error, 1 ok        <- exit 1, no visible error row
```
this fails CI for a correctly configured app. can we move the bundle-managed short-circuit above checkConfig? and printResourceRow should still print non-ok layers instead of hiding them for bundle-managed

packages/shared/src/cli/commands/doctor/checks.ts
```diff
@@ -0,0 +1,349 @@
...
+ * error message, or null when the host is acceptable or unset.
+ */
+export function validateHost(host: string | undefined): string | null {
+  if (host === undefined || host.trim().length === 0) return null;
+
+  let url: URL;
+  try {
+    url = new URL(host);
+  } catch {
+    return `DATABRICKS_HOST is not a valid URL: "${host}"`;
```
MarioCadenas: the comment on checkAuth (146) says we only ever report the sanitized host so embedded creds dont leak, but validateHost builds all three messages (124/128/135) from the raw host, and that goes into auth.detail (159) which prints unconditionally + only auth.raw gets stripped from --json. so this leaks:
```
DATABRICKS_HOST=ftp://user:s3cr3t@secret.example.com appkit doctor --json
...
"host": "ftp://secret.example.com/",         <- redacted, good
"detail": "DATABRICKS_HOST must be an http(s) URL: \"ftp://user:s3cr3t@secret.example.com\""   <- leaks
```
s3cr3t shows up in both human + json without --detail. can we run validateHost against sanitizeHost(rawHost), or build the messages from the sanitized value

packages/shared/src/cli/commands/doctor/checks.ts
```diff
@@ -0,0 +1,349 @@
...
+ */
+export function sanitizeHost(host: string | undefined): string | undefined {
+  if (host === undefined) return undefined;
+  try {
+    const url = new URL(host);
+    if (!url.username && !url.password) return host;
+    url.username = "";
+    url.password = "";
+    return url.toString();
+  } catch {
```
MarioCadenas, line 107: small follow-up to the host leak — on a URL that doesnt parse this catch returns the raw host as-is, so `ht!tp://user:pass@x` keeps the creds in the host field too (and also trips the "not a valid URL" message above). maybe strip a leading `...@` or return a fixed sentinel on parse failure

packages/shared/src/cli/commands/doctor/checks.ts
```diff
@@ -0,0 +1,349 @@
...
+    return `DATABRICKS_HOST is not a valid URL: "${host}"`;
+  }
+
+  if (url.protocol !== "https:" && url.protocol !== "http:") {
+    return `DATABRICKS_HOST must be an http(s) URL: "${host}"`;
+  }
+
+  // Placeholders like "https://..." parse but have no real dotted label.
+  const hostname = url.hostname;
+  const hasRealLabel = /[a-z0-9]/i.test(hostname) && hostname.includes(".");
```
MarioCadenas: this rejects `https://localhost:8080` as an "unfilled placeholder" since theres no dot in the hostname. breaks local/tunnel dev hosts — can we allow localhost explicitly

packages/shared/src/cli/commands/doctor/checks-existence.ts
```diff
@@ -0,0 +1,358 @@
...
+    return EXISTENCE_OK;
+  } catch (err) {
+    return classifyError(err, target);
+  }
+};
+
+const probeJob: ExistenceProbe = async (client, target) => {
+  const raw = field(target, "id");
+  if (!raw) return missingField("id");
+  const jobId = Number(raw);
```
MarioCadenas: job ids are int64, Number() loses precision past 2^53 and Number.isInteger still passes, so we'd probe the wrong id and report a false not-found. safer to validate with `/^\d+$/` on the raw string

packages/shared/src/cli/commands/doctor/report.ts
```diff
@@ -0,0 +1,265 @@
...
+}
+
+/**
+ * Highlights actionable tokens in a detail/hint line: cyan for a "quoted" id or
+ * `backticked` code span (delimiters stripped), bold for a SCREAMING_SNAKE env
+ * var. Nesting is safe — bold's reset (22) doesn't clear the cyan (39).
+ */
+function highlight(text: string): string {
+  return text
+    .replace(/`([^`]+)`/g, (_, code) => pc.cyan(code))
```
MarioCadenas, line 73: tiny one — the jsdoc + readme say env vars go bold+cyan inside a code span, but they dont: the cyan replaces at 66/67 wrap the token in ansi first, so the `\b` in the bold regex at 68 wont match. either drop that claim from the docs or run the bold pass before the cyan ones

MarioCadenas (commented):
ran this e2e against a scaffolded app on a real workspace — behaviour is solid, auth / config / existence (404, stopped) / wiring both directions / --json / exit codes all check out. two things id want fixed before merge: the bundle-managed false-error (hides the error but still exits 1) and the host cred leak in auth.detail. rest are small.

MarioCadenas (approved):

PR #491, opened 2026-07-24
ci: comment a run-anywhere command to try the PR template artifact

MarioCadenas:
## What

Adds a sticky PR comment to the **PR Template Artifact** job that prints a **self-contained, run-anywhere** command to scaffold an app from this PR's SDK build:

```bash
gh run download <run-id> -R databricks/appkit -n appkit-template-<version> -D appkit-pr-<n> \
  && unzip -o "appkit-pr-<n>/appkit-template-<version>.zip" -d "appkit-pr-<n>" \
  && databricks apps init --template "appkit-pr-<n>"
```

## Why

The job already builds a template with this PR's `@databricks/appkit` + `@databricks/appkit-ui` tarballs pinned via `file:` deps — but it lands in the artifacts UI with no obvious way to consume it. Testing meant hunting for the CI run, downloading the zip by hand, unzipping, and remembering the `apps init` invocation.

The requirement here was that the command be **pasteable into any folder** — not run from inside an appkit checkout. So the run id, repo, artifact name, and output dir are all interpolated from CI context at build time; there's no repo-local script to depend on.

## How

- New `.github/scripts/upsert-template-artifact-comment.cjs` — marker-based sticky-comment upsert, cloned from the existing bundle-size / breaking-change scripts.
- Two steps appended to `pr-template-artifact`: one builds the comment body (`template-artifact-comment.md`), one upserts it via `actions/github-script`.
- Job-level `pull-requests: write` (overrides the top-level `read`), matching the bundle-size job.

Additive only (86 lines, 2 files); no change to the artifact itself.

## Notes / limitations

- The command needs the **GitHub CLI** (`gh auth login`) since artifact download requires auth, and the **Databricks CLI**. Both are stated in the comment.
- Fork PRs won't get `pull-requests: write`, so the comment step is skipped there — same behavior as the repo's other comment jobs.
- Verified: YAML parses, biome clean on the `.cjs`, and the rendered comment markdown was simulated locally.


pkosiec:
@MarioCadenas quick question - is it another comment or are we merging with the existing one? 
image

I'd vote to have just a single automated comment under PR to avoid bloat, especially that it's related - WDYT?

MarioCadenas:
> @MarioCadenas quick question - is it another comment or are we merging with the existing one? image
> 
> I'd vote to have just a single automated comment under PR to avoid bloat, especially that it's related - WDYT?

its not related at all and they run in completely different points in time AFAIK

pkosiec (approved):
Approved regardless of the suggestion to merge Evals comment with this

> its not related at all and they run in completely different points in time AFAIK

we already discussed that with Mario offline, the template artifact is going to be used by the evals pipeline (currently it's not due to the limitations of the evals pipeline). The evals comment is posted right away but it should've wait for the artifact too, so it makes sense to merge them 👍

PR #493, opened 2026-07-28
fix(type-generator): treat empty/un-rendered definitions.json as no metric views

keugenek:
## Summary

`readMetricConfig` threw a fatal `Failed to parse definitions.json` whenever `config/metric-views/definitions.json` was **empty** or still held its **un-rendered Go-template guard**.

The template ships that file wrapped in `{{if .plugins.analytics}}`:
```
{{if .plugins.analytics -}}
{ "$schema": "...", "metricViews": {} }
{{- end}}
```
So two entirely normal situations produced invalid JSON:
- **Scaffolding an app without the analytics plugin** → the guard renders nothing → empty file.
- **Running typegen against the raw template** (e.g. an `npm install` `postinstall` → `appkit generate-types` warmup, before `appkit setup` renders anything) → the literal `{{...}}` is still present.

Both are "no metric views defined" — an additive, non-error state the function already documents (`Returns null if the file does not exist ... apps without definitions.json must not be penalized`). But `JSON.parse("")` / `JSON.parse("{{...")` threw, aborting `appkit generate-types` and thus failing `npm install` for any analytics-off app and any tool that warms the raw template.

## Fix

Treat empty / whitespace-only / un-rendered (`{{`-prefixed) content the same as an absent file — `return null`. A non-empty, rendered file with genuinely malformed JSON **still throws loudly**, so real misconfiguration is unaffected.

## How I found it

Discovered while running the Databricks app-evals harness against the `latest` AppKit template (`template-v0.48.0`): the eval cluster's template npm-warmup and the agent's `appkit setup` both died with `FATAL config/metric-views/definitions.json: Failed to parse definitions.json ... Expected property name or '}' in JSON`.

## Testing

Added `mv-registry` tests for the empty, whitespace-only, and un-rendered cases (each expects `null`); the existing "throws on malformed JSON" test still guards that genuine bad JSON errors. Verified the branch logic against all six cases (absent / empty / whitespace / unrendered / valid / malformed).

### DIFF
--- a/packages/appkit/src/type-generator/mv-registry/config.ts
+++ b/packages/appkit/src/type-generator/mv-registry/config.ts
@@ -46,11 +46,17 @@ function compareKeys(a: string, b: string): number {
  * Read {@link METRIC_CONFIG_FILE} from a metric-views folder
  * (`config/metric-views/`).
  *
- * Returns `null` if the file does not exist (the metric-view path is
- * additive — apps without definitions.json must not be penalized). There is
- * deliberately no fallback to a legacy filename.
+ * Returns `null` if the file does not exist, is empty, or has not been rendered
+ * yet (the metric-view path is additive — apps without a defined
+ * definitions.json must not be penalized). The template ships definitions.json
+ * wrapped in a `{{if .plugins.analytics}}` guard, so scaffolding without the
+ * analytics plugin leaves the file empty, and running typegen against the raw
+ * (un-rendered) template leaves the literal `{{...}}` in place — neither is a
+ * misconfiguration, so both are treated the same as an absent file rather than
+ * a fatal parse error. There is deliberately no fallback to a legacy filename.
  *
- * Throws on JSON parse errors so misconfiguration surfaces loudly.
+ * Throws on JSON parse errors in a *non-empty, rendered* file so genuine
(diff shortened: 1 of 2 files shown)


pkosiec (commented):
@keugenek Are you sure this is correct fix?
The templating syntax should be only used for unrendered template, and the databricks apps init command renders it so that there's no longer such syntax 🤔. t

The type generator shouldn't see `{{ ... }}` go templating at all after the template is rendered. What I mean is that the descrinbed "Running typegen against the raw template" isn't a valid case and we don't support it - we should use the template (PR one, or whatever one) only after running `databricks apps init --template {path}` with proper flags.

You can even see the automatic comment how to use the PR template artifact:

image


Does it make sense?

keugenek:
Closing this — @pkosiec is right that the type generator shouldn't ever see un-rendered `{{ }}` Go-template syntax. That only appears when a template is used **without** `databricks apps init` rendering it first, which isn't a supported flow.

The real bug was on the consumer side: the App Builder eval harness was raw-copying the template (`shutil.copytree`) instead of rendering it via `databricks apps init --template`. Fixing that there (so the app is properly rendered before typegen/setup runs) rather than making the type generator tolerate un-rendered input. Thanks for the catch.

PR #499, opened 2026-07-29
ci: fix devhub-validation (build with pnpm) and add a nightly DevHub build

pkosiec:
## What

- Build DevHub with **pnpm** instead of npm in the `devhub-validation` job (`pnpm install --frozen-lockfile` + `pnpm build`).
- Extract the DevHub build into a **reusable workflow** (`devhub-build.yml`) so the per-PR job and the new nightly share one definition.
- Add a **nightly** scheduled build (`devhub-nightly.yml`) that validates AppKit `main` docs against DevHub daily.

## Why

DevHub migrated from npm to pnpm (dropping `package-lock.json` for `pnpm-lock.yaml`), but this job kept installing it with `npm install`:

- npm ignores a pnpm lockfile, so it did a **floating, non-reproducible resolve on every run**.
- A transitive dependency publishing a new version then silently changed the installed tree with **no commit on either side** — the job passed one run and failed the next.
- That drift pulled `@databricks/appkit-ui/react` into Next's react-server graph, where React 19 omits `createContext`, crashing `next build` on `/_not-found` with `TypeError: b.createContext is not a function`.

`pnpm install --frozen-lockfile` pins the exact tree DevHub itself ships, so the job is reproducible and can only go red on a real change — broken AppKit docs (the signal we want) or a genuine DevHub-side change.

## Nightly build

The per-PR job only runs on docs PRs, so a DevHub-side breakage can go unnoticed until the next docs change. The nightly runs the same build against `main` once a day. Failures surface via GitHub's built-in scheduled-workflow-failure email — no extra secrets or notification code.

## Notes

- Both callers grant `id-token: write` so the JFrog OIDC step keeps working. DevHub's lockfile (v9.0) uses integrity-only resolutions, so `--frozen-lockfile` is registry-agnostic through the JFrog proxy.
- **Latent, out of scope**: appkit-ui's `/react` barrel re-exports modules that top-level-call `createContext` without `"use client"`. It only bites under an npm flat-hoist + RSC combo DevHub never uses, and this job compiles the *published* appkit-ui, not PR code. Tracked separately.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -37,6 +37,3 @@ The `pnpm build` command:
 2. Runs `apply-redirects` — replaces HTML pages with redirect pages pointing to developers.databricks.com
 
 Static files remain served from GitHub Pages: JSON schemas (`/schemas/`), `llms.txt`, and `.md` files (used by `npx @databricks/appkit docs` for npm-bundled documentation).
-
-<!-- ci: temporary touch to run devhub-validation via the reusable workflow; reverted before merge -->
-
--- a/docs/README.md
+++ b/docs/README.md
@@ -37,6 +37,3 @@ The `pnpm build` command:
 2. Runs `apply-redirects` — replaces HTML pages with redirect pages pointing to developers.databricks.com
 
 Static files remain served from GitHub Pages: JSON schemas (`/schemas/`), `llms.txt`, and `.md` files (used by `npx @databricks/appkit docs` for npm-bundled documentation).
-
-<!-- ci: touch to run devhub-validation (docs path filter); safe to drop before merge -->
-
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -37,3 +37,6 @@ The `pnpm build` command:
 2. Runs `apply-redirects` — replaces HTML pages with redirect pages pointing to developers.databricks.com
 
 Static files remain served from GitHub Pages: JSON schemas (`/schemas/`), `llms.txt`, and `.md` files (used by `npx @databricks/appkit docs` for npm-bundled documentation).
+
+<!-- ci: touch to run devhub-validation (docs path filter); safe to drop before merge -->
+
```
pkosiec:
```suggestion
```

atilafassina (approved):
🏆  amazing work! Thank you

PR #502, opened 2026-07-31
feat(appkit): relocate typegen cache to committed .appkit/ dir

atilafassina:
## What & why

AppKit's analytics typegen depends on a per-project cache of `DESCRIBE` results. Today it lives under `node_modules/.databricks/appkit/`, so every `pnpm install` destroys it and git can't see it. An upstream authorization change is coming that will leave **CI/CD without warehouse access** — at which point a cache wiped by `install` means every deploy is an all-MISS build with no warehouse to satisfy it, and every deploy crashes.

This relocates the two build-input caches to a **committed, per-app `.appkit/` directory** so a fresh `git checkout` + `pnpm install` on CI arrives with a warm cache. The committed cache acts as a types-lockfile: an all-hit blocking build needs no warehouse, and a genuine miss that can't be resolved **fails loud** instead of shipping degraded (`result: unknown`) types.

The blocking/non-blocking typegen architecture is **unchanged** — blocking mode already provides HIT-skip / MISS-attempt / unresolved-MISS-fatal. The only thing defeating it was the cache location.

## Changes

- Shared path resolver for cache PATH. Single-source-of-truth for all tasks and runtime
- Sort cache entries so output shape is deterministic (prevents noise on `git diff`)
- Discern connectivity errors from warehouse Auth access with a tailored error message on the latter
- move cache from `node_modules/.databricks` to `.appkit/`
- ignore `.appkit/**` on Biome linter
- Add a one-liner to template Readme.md
- Add a few relevant tests

> [!NOTE]
> Also check https://github.com/databricks/appkit/pull/503 for the docs

### DIFF
--- a/packages/appkit/src/type-generator/index.ts
+++ b/packages/appkit/src/type-generator/index.ts
@@ -178,17 +178,25 @@ export class TypegenFatalError extends Error {
   constructor(
     queries: QueryFatalError[],
     warehouseId?: string,
-    cacheInitialized: boolean = true,
+    cacheInitialized?: boolean,
   ) {
-    // Distinguish bootstrap (no committed cache yet) from drift (cache exists but is stale/missing key).
-    // Bootstrap → operator needs to initialize; drift → operator needs to regenerate the affected key.
-    const nextStep = cacheInitialized
-      ? // Drift: cache file exists but is missing/stale for the failing query/metric.
-        warehouseId
-        ? `The committed ${pc.bold(".appkit/")} cache is missing or stale for the failed ${plural(queries.length, "query", "queries")}. Regenerate with ${pc.bold("generate-types --wait")} against warehouse ${pc.bold(warehouseId)} and commit ${pc.bold(".appkit/")}.`
-        : `The committed ${pc.bold(".appkit/")} cache is missing or stale for the failed ${plural(queries.length, "query", "queries")}. Regenerate with ${pc.bold("generate-types --wait")} against a warehouse and commit ${pc.bold(".appkit/")}.`
-      : // Bootstrap: no committed cache found; operator needs to initialize from scratch.
-        `No committed type cache found (${pc.bold(".appkit/")}). Run ${pc.bold("generate-types --wait")} against a warehouse and commit ${pc.bold(".appkit/")} before building without warehouse access.`;
+    // The cache snapshot is optional: callers whose failure relates to the type
+    // cache pass it to get bootstrap/drift remediation; callers with an unrelated
+    // fatal (e.g. a malformed definitions.json) omit it and get generic wording.
+    const nextStep =
+      cacheInitialized === undefined
+        ? // No cache snapshot: the failure is unrelated to the cache — don't
+          // misdirect the operator toward regenerating/committing `.appkit/`.
+          `Fix the ${pc.bold("FATAL")} error(s) above, then re-run ${pc.bold("generate-types")}.`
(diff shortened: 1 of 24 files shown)


packages/appkit/src/type-generator/index.ts
pkosiec, line 1: I'm wondering if we should a small migration for the existing app users after appkit upgrade to force typegen 🤔 I think it would be beneficial - it would run one-time only typegen for a redeployed app

what do you think? 

```
Add helper in migration.ts next to removeOldGeneratedTypes:

  import { getEphemeralStateDir } from "shared/cli/commands/cache-paths";

  /**
   * Remove pre-`.appkit/` typegen caches from their old ephemeral location.
   * Best-effort: silently ignores missing files. Deletes only the two cache
   * JSONs — the spawn lock and UI-choices file still live in this dir.
   */
  export async function removeOldTypegenCaches(projectRoot: string): Promise<void> {
    const oldDir = getEphemeralStateDir(projectRoot);
    for (const f of [".appkit-types-cache.json", ".appkit-serving-types-cache.json"]) {
   * JSONs — the spawn lock and UI-choices file still live in this dir.
   */
  export async function removeOldTypegenCaches(projectRoot: string): Promise<void> {
    const oldDir = getEphemeralStateDir(projectRoot);
    for (const f of [".appkit-types-cache.json", ".appkit-serving-types-cache.json"]) {
      try {
        await fs.unlink(path.join(oldDir, f));
        logger.debug("Removed old typegen cache at %s", path.join(oldDir, f));
      } catch {
        // Not present — nothing to clean up.
      }
    }
  }

  Call it from index.ts:406, right beside the existing cleanup:

  await removeOldGeneratedTypes(projectRoot, "appKitTypes.d.ts");
  await removeOldTypegenCaches(projectRoot);          // <- add
  await migrateProjectConfig(projectRoot);
```

pkosiec, line 1: General comment for the whole PR: can we run deslop to reduce the comments?
https://github.com/pffigueiredo/claude-code-settings/blob/main/commands/clean-ai-slop.md

pkosiec (approved):
LGTM - but please see the comments before merge. Thanks!

packages/appkit/src/type-generator/index.ts
atilafassina: Not a strong opinion, but i initially decided against it because the current cache lives in `node_modules` and once they upgrade, on the first run it will recreate the cache. On build, it will be the exact same behavior if they don't generate the types locally, if they do - it will have a cache MISS locally and then the next build will simply be faster.

So I think this is an incremental migration and I figured it wasn't worth it to write code for it.

pkosiec: Are you sure the cache will be gone from node_modules on upgrade? We saved it in `.databricks` which probably won't be touched on upgrade?

If you tested that and it works - great, so the migration is already handled without any custom code needed, (that's exactly what my intention was) 👍

atilafassina: cache will be gone once they clean `node_modules`, on upgrade the cache will simply be ignored because it will look for at another path :)

pkosiec: right, forgot that we'll always look at the new path 😅 ... so basically we'll just keep the unused files in the node_modules. Yeah, maybe it is not worth to do the "migration" (or rather, a cleanup). Leaving that up to you to decide as it's rather a nit then 👍

PR #512, opened 2026-08-04
feat(analytics): add Metric View metadata pipeline

atilafassina:
## TL;DR

Carries Unity Catalog Metric View display metadata from type generation into Analytics SSE results, giving clients a typed, cache-safe metadata source without runtime `DESCRIBE` calls or duplicated labels and formats.

## Delivered

- [x] Generate `metric-views.ts` with `MetricRegistry` augmentation and a build-inlined `metricViewsMetadata` constant.
- [x] Accept generated metadata in the Analytics plugin and attach only the requested measures and dimensions to each result.
- [x] Keep metadata outside SQL generation and cache identity, applying the latest metadata after cached query execution.
- [x] Migrate the generated artifact from `.d.ts` to `.ts` and remove stale declarations safely.
- [x] Preserve committed generated artifacts during warehouse outages and require every configured artifact before using the fallback.
- [x] Cover generator output, payload scoping, cache behavior, migration, and outage handling with focused tests.

Base for #488.


packages/shared/src/metric-metadata.ts
pkosiec, line 1: nit - the Meta vs Metadata is quite confusing. I thought it's a distinction between build and wire metadata but looks like it's not. What do you think about names like `MetricColumnDisplay` and `MetricViewsDisplay`? 

Even the comment already says "Per-column display metadata".

pkosiec (approved):
As we discussed, before the merge, please run deslop to reduce/get rid of some of the overly descriptive / unnecessary comments. Thanks!

packages/shared/src/metric-metadata.ts
atilafassina: done!

PR #513, opened 2026-08-05
feat(shared): lint rule flags aiSearch indexes missing columns

MarioCadenas:
## What

Adds an `appkit lint` rule, `ai-search-index-requires-columns`, that flags an `aiSearch` plugin config whose indexes have no usable `columns`.

## Why

In production the plugin does **not** auto-discover an index's columns — that's dev-only (`ai-search.ts` `_autoDiscoverColumns`, gated on `NODE_ENV === "development"`). So a config that omits `columns` works in dev but fails at query time in prod (`Field 'columns' must be specified`). `appkit lint` is the pre-deploy static catch, complementing the existing dev nudges (auto-discovery banner + manifest `scaffolding.rules`).

## How

`aiSearch({ indexes: { a: {...}, b: {...} } })` — "every index has non-empty `columns`" can't be expressed as a single ast-grep pattern, so this adds an optional `find?: (root) => SgNode[]` escape hatch to the `Rule` interface. Flat pattern rules and this structural rule share the same reporting path. `lintFile` now delegates to an exported `lintSource()` so rules can be tested against in-memory source.

**Flagged** (fail in prod):
- bare `aiSearch()`, `aiSearch({})`, `aiSearch({ indexes: {} })` — all fall back to the env-seeded default index, which has no columns
- any index object missing `columns`, or with `columns: []`

**Passed** (can't statically prove bad — no false positives):
- `columns: SOME_CONST`, `indexes: buildIndexes()`, `aiSearch(dynamicConfig)`, spread index objects (`{ ...base }`)

The violation points at the offending index alias, not the whole call.

## Test plan

- `packages/shared/src/cli/commands/lint.test.ts` — 14 vitest cases covering every pass/flag branch above
- `pnpm exec vitest run lint` — 14/14 pass
- `pnpm -r typecheck` — clean
- `pnpm exec biome check` on changed files — clean
- Manual: `appkit lint` over a temp `server.ts` with a bad `aiSearch()` flags it and exits 1

### No columns added to the plugin:

```ts
createApp({
  plugins: [
    aiSearch(),
    server(),
  ],
}).catch(console.error);
```

or this 

```ts
createApp({
  plugins: [
 aiSearch({ indexes: { default: { columns: [] } } }),
    server(),
  ],
}).catch(console.error);
```
will produce this output

```bash
➜ npm run lint:ast-grep

> appkit-pr-513-app@1.0.0 lint:ast-grep
> appkit lint

Scanning 11 TypeScript files...

Found 1 violation(s):

server/server.ts:6:5
  ai-search-index-requires-columns: AI Search index has no `columns`; it will fail in production (columns are auto-discovered only in dev). Set `columns` explicitly on each index.
  > aiSearch()
```

#### Columns added

```ts
createApp({
  plugins: [
    aiSearch({ indexes: { default: { columns: ['my_column', 'column'] } } }),
    server(),
  ],
}).catch(console.error);
```

```bash
➜ npm run lint:ast-grep

> appkit-pr-513-app@1.0.0 lint:ast-grep
> appkit lint

Scanning 11 TypeScript files...

No ast-grep lint violations found.
```

### DIFF
--- a/packages/shared/src/cli/commands/lint.test.ts
+++ b/packages/shared/src/cli/commands/lint.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, test } from "vitest";
+import { lintSource } from "./lint";
+
+const RULE = "ai-search-index-requires-columns";
+
+/** Lint TS source, keeping only this rule's violations. */
+function columnsViolations(code: string) {
+  return lintSource(code, "server.ts").filter((v) => v.rule === RULE);
+}
+
+describe("ai-search-index-requires-columns", () => {
+  test("passes an index with columns", () => {
+    const code = `aiSearch({ indexes: { demo: { columns: ["id", "text"] } } });`;
+    expect(columnsViolations(code)).toHaveLength(0);
+  });
+
+  test("flags an index missing columns", () => {
+    const code = `aiSearch({ indexes: { demo: { queryType: "hybrid" } } });`;
+    const v = columnsViolations(code);
+    expect(v).toHaveLength(1);
+    expect(v[0].code).toContain("demo");
+  });
+
+  test("flags an index with an empty columns array", () => {
+    const code = `aiSearch({ indexes: { demo: { columns: [] } } });`;
+    expect(columnsViolations(code)).toHaveLength(1);
+  });
+
+  test("flags an empty array with whitespace", () => {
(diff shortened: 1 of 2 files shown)


atilafassina (approved):
i think i found one edge-case with a spread operator, but let's not defer this plugin on that. It delivers well despite a small false-positive that arguably would be against best practice anyway 🐑 it

PR #525, opened 2026-08-10
feat(appkit): add the database runtime and harden its schema builder

ditadi:
## Stack

- [**#525** — the database runtime](https://github.com/databricks/appkit/pull/525) ← **you are here**
  - [**#526** — the typed plugin API](https://github.com/databricks/appkit/pull/526)
    - [**#527** — generated reads](https://github.com/databricks/appkit/pull/527)

Each PR targets the one above it. This one is the base of the stack, so its diff is against `main`.

## What

Rebuilds `packages/appkit/src/database/` into a foundation the upcoming `DatabasePlugin` can be built on: the schema builder now rejects the declarations it used to accept silently, and a new runtime layer gives every future database API a single, bounded path to PostgreSQL.

Nothing here is exported. Neither `index.ts` nor `beta.ts` reaches into `database/` today, so this PR changes no public API and no generated types or docs — it is groundwork, reviewable on its own, and the plugin that consumes it lands in the following PRs.

## Changes

**A backend-neutral execution boundary (`database/runtime/`)**

`DataPath` is the only interface the layers above will talk to — `select`, `findOne`, `count`, `insert`, `update`, `upsert`, `delete`, `raw`, and `transaction`. Its field names are schema keys that an adapter resolves through column metadata, never caller-supplied SQL identifiers. The Drizzle adapter and the `translate` module behind it hold the guarantees:

- Identifiers come from metadata and every operand is parameterized. Tagged `raw` SQL interpolates values only; a structural interpolation is rejected.
- Reads are bounded by default. A collection read without an explicit limit takes `DEFAULT_LIMIT`, explicit bounds are validated against `MAX_LIMIT`, and to-many includes are bounded the same way.
- The default projection is private-safe, so a column marked `.private()` never leaves the database unless it is asked for by name.
- Filters go through one operator matrix keyed by column value kind. `in` lists are bounded, an empty `in` has deterministic semantics rather than matching everything, null matching is only ever `is: null`, and an empty `and`/`or` group is rejected instead of silently widening the query.
- Row counts are invariants, not hopes: `insert` and `upsert` must produce exactly one row, `update` and `delete` accept zero or one and reject many.
- A table handle from a different schema is rejected, and driver errors are raised without retaining their details.

**Schema-builder invariants**

The builder previously accepted declarations that could not hold at runtime. It now refuses them at declaration time:

- Foreign keys inherit the target's full storage contract, resolve through chains independently of declaration order, and reject cycles that never reach a concrete type. A target must be a primary key or unique; unknown, cross-schema, and omitted targets are refused. `SET NULL` requires a nullable column and `SET DEFAULT` a compatible local default. A generated identity mirrors to non-generated integer storage on the referencing side.
- Relations reject forward and reverse collisions with column names, reject a second foreign key from one table to the same target, keep reverse relations to-many even when the foreign key is unique, and expose only the forward edge of a self-reference. The resulting objects are frozen.
- Columns allow one explicit default mode, accept referential actions only where a foreign key exists, validate `varchar` lengths and enum declarations, record literal defaults rather than synthesizing them, and restrict the `defaultNow`/`defaultRandom` helpers to the timestamp and UUID kinds.
- `defineSchema` requires each declared table exactly once, rejects aliases and duplicate handles, validates literal defaults against both storage and enum values, refuses a table that would overwrite Drizzle relation metadata, and no longer leaves handles half-finalized when validation fails partway through.

**Removals**

- `.owner()` and the `isOwner` metadata it set are gone. They described row ownership for an RLS story this runtime does not implement, and nothing in the repository called them.
- `contract/column-info.ts`, `contract/relation.ts`, and `schema-builder/private.ts` were folded into the modules that own their data, which narrows the internal barrel accordingly.

## Verification

- `pnpm vitest run` — 3966 passing, 1 skipped
- `pnpm run generate:types`, `pnpm run sync:template`, and `pnpm run docs:build` produce no drift, as expected for a change with no exported surface
- Rebased onto `main` at v0.57.0


packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
```diff
@@ -0,0 +1,270 @@
...
+      return runDatabaseOperation(() =>
+        db.$count(pgTable(table), whereSql(table, where)),
+      );
+    },
+
+    async insert(table, values) {
+      const engineTable = pgTable(table);
+      const parameters = mutationValues(table, values);
+      const rows = await runDatabaseOperation(() =>
+        db.insert(engineTable).values(parameters).returning(),
```
atilafassina: I think here (and in other operations) we're using `.returning()` but it doesn't take into account if a column is `private()`, I think we'll need to abstract this away and avoid passing all columns to `expect*` methods. Wdyt?

packages/appkit/src/database/runtime/data-path.ts
```diff
@@ -0,0 +1,126 @@
...
+  return limit;
+}
+
+/** Apply the conservative collection default when no limit is supplied. */
+export function limitOrDefault(limit?: number): number {
+  return limit === undefined ? DEFAULT_LIMIT : validateLimit(limit);
+}
+
+/** Reject offsets that PostgreSQL cannot represent safely as JS integers. */
+export function validateOffset(offset: number): number {
```
atilafassina, line 101: thoughts about setting an offset max boundary as well here?
Maybe this is worth it as belt-and-suspenders protection?

atilafassina (commented):
Awesome work. just 2 question/suggestions before approving :)

packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
ditadi: Nice catch. `returningColumns()` now shares the same public-column set as default reads.

packages/appkit/src/database/runtime/data-path.ts
ditadi: Nice, yes. I added a `MAX_OFFSET` cap.

packages/appkit/src/database/runtime/engine/translate.ts
```diff
@@ -0,0 +1,307 @@
...
+  return Object.entries(order).map(([key, direction]) => {
+    if (direction !== "asc" && direction !== "desc") {
+      throw new DataPathError(`Unknown order direction "${direction}"`);
+    }
+    const column = columnOf(table, key);
+    return direction === "desc" ? desc(column) : asc(column);
+  });
+}
+
+export function selectToColumns(
```
MarioCadenas, line 301: **[High] Private-column protection is projection-default-only.** `isPrivate` is consulted only in `defaultColumns`; `selectToColumns` (here) has no `isPrivate` gate, so an explicit `select: ["secret"]` projects a `.private()` column. Separately, `translateWhere`/`translateOrder` resolve private columns too, so `where: {secret: {like: "a%"}}` / `order: {secret: "asc"}` act as a prefix/binary-search exfiltration oracle even without projecting the value. This is a trusted-port design (the HTTP layer in #527 filters via `selectable`/`queryable`), but the runtime offers no defense-in-depth on `select`/`where`/`order`.

_Automated review finding._

packages/appkit/src/database/runtime/engine/translate.ts
```diff
@@ -0,0 +1,307 @@
...
+export function translateWhere(
+  table: AppKitTable,
+  clause: WhereClause,
+): SQL | undefined {
+  if (clause === null || typeof clause !== "object" || Array.isArray(clause)) {
+    throw new DataPathError("where must be an object");
+  }
+  const conditions: SQL[] = [];
+  for (const [key, value] of Object.entries(clause)) {
+    if (key === "and" || key === "or") {
```
MarioCadenas, line 211: **[Medium] The `where` predicate tree has no depth or breadth cap.** The `and`/`or` branch recurses with no depth limit and maps the group array with no size limit — unlike `translateInclude` (bounded by `MAX_INCLUDES`). A deeply nested `{and:[{and:[…]}]}` can overflow the stack (caught as a generic 500), and a wide `{or:[…50k…]}` defeats `IN_CAP` and forces an arbitrarily expensive predicate. #527's HTTP decoder bounds this, but the runtime itself does not, so direct/SP callers remain unbounded.

_Automated review finding._

packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
```diff
@@ -0,0 +1,288 @@
...
+      const engineTable = pgTable(table);
+      const parameters = mutationValues(table, values);
+      const target = conflictTargetMeta(table, onConflict);
+      const rows = await runDatabaseOperation(() =>
+        db
+          .insert(engineTable)
+          .values(parameters)
+          .onConflictDoUpdate({
+            target: columnOf(table, target.columnName),
+            set: parameters,
```
MarioCadenas: **[Medium] `upsert` can rewrite an existing row's primary key.** `onConflictDoUpdate({ set: parameters })` reuses the full insert payload — including the PK — as the DO UPDATE SET. With a natural (non-`serverGenerated`) PK plus a separate unique column, `upsert({pk:"new", uniqueCol:"existing"}, {onConflict:"uniqueCol"})` flips the existing row's PK → FK orphaning / identity reassignment. `deriveUpdateSchema` excludes the PK precisely to prevent this; the upsert `set` should do the same (exclude the PK and ideally the conflict target).

_Automated review finding._

packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
```diff
@@ -0,0 +1,288 @@
...
+  table: AppKitTable,
+  select?: readonly string[],
+): Record<string, true> {
+  return select === undefined
+    ? defaultColumns(table)
+    : selectToColumns(table, select);
+}
+
+/** Keep mutation identifiers schema-owned and every supplied value parameterized. */
+function mutationValues(table: AppKitTable, values: Row): Row {
```
MarioCadenas: **[Medium] The mutation port applies none of the schema's own write invariants.** `mutationValues` only checks key membership and rejects SQL wrappers; it does not run `columnValueSchema` (type check) or exclude `serverGenerated`/`primaryKey`. A caller reaching `DataPath` directly can forge a server-generated `id` (identity is `generatedByDefaultAsIdentity`, which accepts explicit values) or rewrite a PK via `update`. This is by-contract (the typed client in #526 applies `$insertSchema`/`$updateSchema` first), but any future caller that reaches `DataPath` without that layer bypasses every write guard.

_Automated review finding._

atilafassina (approved):

packages/appkit/src/database/runtime/engine/translate.ts
ditadi: nice catch, fixed by enforcing private-columns checks across select, where, order and include with explicit trusted access for internal callers.

packages/appkit/src/database/runtime/engine/translate.ts
ditadi: nice catch, added runtime limits for predicate depth, group size and total conditions

packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
ditadi: nice catch, upsert updates now exclude pk, server-generated fields and the conflict target

packages/appkit/src/database/runtime/engine/drizzle-data-path.ts
ditadi: nice catch, mutations now revalidates values with the finalized insert/update schemas before reaching the database.

PR #527, opened 2026-08-10
feat(appkit): add opt-in generated database reads

ditadi:
## Stack

- [**#525** — the database runtime](https://github.com/databricks/appkit/pull/525)
  - [**#526** — the typed plugin API](https://github.com/databricks/appkit/pull/526)
    - [**#527** — generated reads](https://github.com/databricks/appkit/pull/527) ← **you are here**

Each PR targets the one above it, so the diff shown here is only the delta on top of #526. Review in order.

## What

Adds the first HTTP surface: `crudRoutes` projects the typed read API from #526 onto generated `GET /:table` and `GET /:table/:id` routes. It is off by default and opt-in per table, because a generated route is reachable by anyone the app admits.

```ts
database({
  schema,
  crudRoutes: { tables: ["notes"] },
  hooks: {
    notes: {
      serialize: (row) => ({ ...row, excerpt: String(row.body).slice(0, 80) }),
    },
  },
});
```

```
GET /api/database/notes?where={"author":{"eq":"ada"}}&order={"createdAt":"desc"}&limit=20
→ { "items": [...], "limit": 20, "offset": 0 }

GET /api/database/notes/42
→ { "id": 42, "body": "...", "createdAt": "..." }
```

Only writes are missing after this PR; they arrive in the next one.

## Changes

**Exposure is a decision per table (`crud/exposure.ts`)**

`crudRoutes` accepts `false` (the default), `true`, or `{ tables: [...] }`, and the table names are checked against the schema type, so a typo does not silently expose nothing. An enabled table is also what makes it includable from its neighbours: a relation whose target is not enabled cannot be included, so one table's data sits behind exactly one decision rather than leaking through a join.

**The query grammar is bounded before any SQL runs (`crud/query.ts`)**

`where`, `order`, `select`, `include`, `limit`, and `offset` are decoded from the raw query string — not from Express's normalized `req.query`, which would accept repeated and array-shaped parameters. Every decoded piece is checked against the table's compiled columns: an unknown column, an operator the column's kind does not support, or a value that fails its codec is a 400 before the plugin is asked for anything.

The budgets are explicit constants in `defaults.ts` and are all enforced at decode time: query string size, `where` nesting depth and condition count, `order` field count, `offset` ceiling, and the number of rows the include tree may materialize. `limit` defaults to a page and is capped by the same wire cap a typed caller sees, so HTTP cannot ask for more than server code can.

Rejections name a fixed parameter and a fixed sentence. The decoder never echoes caller-supplied text back into the response.

**Rows are shaped, not forwarded (`crud/contract.ts`, `crud/codecs.ts`)**

Each enabled table compiles once into a `CrudTable`: its public columns, their codecs, its primary key decoder, and a projection that drops private columns. A row is projected before it reaches the optional `serialize` hook, so a serializer cannot re-expose a column the schema marked private, and the hook's output is re-sanitized against depth and node budgets afterwards. `serialize` is typed to return synchronously — a `Promise` does not compile — because it runs inside the response path.

**Responses are bounded and never cached**

The encoded body is measured before it is sent, so a request that would exceed the byte budget fails as `413` instead of streaming a partial answer. Every generated read sends `Cache-Control: no-store`: the same URL answers differently once the table changes.

**Pagination is stable**

The list handler appends the primary key to whatever `order` the caller asked for, so rows with equal sort keys cannot reshuffle between pages. A table without a primary key has no tie-breaker to append, so it must name its own `order` and is told so.

**Spans**

Each generated read runs inside a span named for its route template, not its URL, and a failure is recorded as `not_found`, `rejected`, or `failed` — derived from the safe status code, so cardinality stays bounded and no caller input reaches the span.

## Known limitation

Text filters accept caller-supplied `like`/`ilike` patterns, and this beta adds no statement cancellation below the connector, so an expensive pattern runs to completion while holding its pooled connection. This is documented on `CrudRoutesConfig` alongside the note that generated routes carry no per-user filter.

## Verification

- `pnpm vitest run` — 4109 passing, 1 skipped; new suites cover the query decoder and its budgets, the codecs, the row contract, the route handlers, and the read spans

### DIFF
--- a/packages/appkit/src/plugins/database/crud/contract.ts
+++ b/packages/appkit/src/plugins/database/crud/contract.ts
@@ -182,7 +182,10 @@ function compileTable(table: AppKitTable): MutableCrudTable {
   for (const meta of Object.values(table.$columns)) {
     const column = compileColumn(meta);
     columns.set(meta.columnName, column);
-    if (meta.primaryKey) primaryKey = column;
+    // A private key must not power `GET /:table/:id`: per-id probing would
+    // answer 200 or 404 on an identifier the schema hides, so over HTTP the
+    // table is keyless — no detail route, and lists must name their own order.
+    if (meta.primaryKey && !meta.isPrivate) primaryKey = column;
     if (meta.isPrivate) continue;
     selectable.add(meta.columnName);
     if (filterOperatorsForKind(meta.kind).length > 0) {
--- a/packages/appkit/src/plugins/database/crud/contract.ts
+++ b/packages/appkit/src/plugins/database/crud/contract.ts
@@ -1,0 +1,243 @@
+import {
+  DatabasePluginError,
+  invalidDatabaseInput,
+} from "../../../database/errors";
+import type { IdValue, Row } from "../../../database/runtime";
+import type { AppKitTable } from "../../../database/schema-builder";
+import { filterOperatorsForKind } from "../../../database/schema-builder/types";
+import { MAX_SERIALIZED_DEPTH, MAX_SERIALIZED_NODES } from "../defaults";
+import { type CompiledColumn, compileColumn, type JsonValue } from "./codecs";
+
+/** One relation edge wired to the contract of its target table. */
+export interface CrudRelation {
+  readonly cardinality: "toOne" | "toMany";
+  readonly target: CrudTable;
+}
+
+/** Private HTTP contract compiled once for one explicitly exposed table. */
+export interface CrudTable {
+  readonly name: string;
+  readonly primaryKey?: CompiledColumn;
+  readonly columns: ReadonlyMap<string, CompiledColumn>;
+  /** Public columns a request may project. */
+  readonly selectable: ReadonlySet<string>;
+  /** Public columns a request may filter or order by. */
+  readonly queryable: ReadonlySet<string>;
+  readonly relations: ReadonlyMap<string, CrudRelation>;
+  decodeId(raw: string): IdValue;
+  projectPublicRow(row: Row): JsonValue;
+  sanitizeSerializedRow(row: unknown): JsonValue;
+}
+
+type MutableCrudTable = Omit<CrudTable, "relations"> & {
+  readonly relations: Map<string, CrudRelation>;
+};
+
+interface SanitizeState {
+  nodes: number;
+  readonly ancestors: Set<object>;
+}
+
+/** A bare object literal; a `Date`, class instance, or `Map` is not JSON. */
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  if (value === null || typeof value !== "object" || Array.isArray(value)) {
+    return false;
+  }
+  const prototype = Object.getPrototypeOf(value);
+  return prototype === Object.prototype || prototype === null;
+}
+
+/** A serializer that breaks its contract is trusted code failing, not input. */
+function serializerFault(): never {
+  throw new DatabasePluginError("INTERNAL", "read");
+}
+
+/** Charge one value against the output budget before descending into it. */
+function countNode(depth: number, state: SanitizeState): void {
+  state.nodes += 1;
+  if (state.nodes > MAX_SERIALIZED_NODES || depth > MAX_SERIALIZED_DEPTH) {
+    serializerFault();
+  }
+}
+
+/** Walk a container while its ancestors are tracked, so a cycle cannot pass. */
+function enterObject<T>(
+  value: object,
+  state: SanitizeState,
+  visit: () => T,
+): T {
+  if (state.ancestors.has(value)) serializerFault();
+  state.ancestors.add(value);
+  try {
+    return visit();
+  } finally {
+    state.ancestors.delete(value);
+  }
+}
+
+/** Accept a serializer's own added value only where it is already JSON. */
+function sanitizeJson(
+  value: unknown,
+  depth: number,
+  state: SanitizeState,
+): JsonValue {
+  countNode(depth, state);
+  if (
+    value === null ||
+    typeof value === "string" ||
+    typeof value === "boolean"
+  ) {
+    return value;
+  }
+  if (typeof value === "number") {
+    if (!Number.isFinite(value)) serializerFault();
+    return value;
+  }
+  if (Array.isArray(value)) {
+    return enterObject(value, state, () =>
+      value.map((item) => sanitizeJson(item, depth + 1, state)),
+    );
+  }
+  if (!isPlainObject(value)) serializerFault();
+  return enterObject(value, state, () => {
+    // A null prototype keeps `__proto__` an ordinary key instead of a setter.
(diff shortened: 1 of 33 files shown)


packages/appkit/src/plugins/database/database.ts
```diff
@@ -48,6 +69,49 @@ export class DatabasePlugin<TSchema extends Schema> extends Plugin<
...
+      | Record<string, { serialize?: ReadSerializer } | undefined>
+      | undefined;
+    // Every exposed name is a declared table, so its export is an entity client.
+    const entities = () =>
+      this.exports() as unknown as Record<string, CrudReadEntity>;
+
+    for (const table of tables.values()) {
+      const deps: ReadRouteDeps = {
+        table,
+        entity: () => entities()[table.name],
```
MarioCadenas, line 94: **[High] Generated routes have no per-user / per-row authorization.** The handler resolves its entity via `this.exports()` (service principal) with no `.asUser(req)` and no owner/row-scope predicate — the `.owner()` concept was dropped in #525. So enabling a table (especially `crudRoutes: true`, which exposes every table) makes every non-private column of every row readable by anyone the app admits, ignoring their own DB grants. It is documented in the `types.ts` docstring, but there is no code-level guardrail or owner-scope seam. Consider requiring an explicit per-table authorization/owner callback (and/or OBO execution).

_Automated review finding._

packages/appkit/src/plugins/database/crud/routes.ts
```diff
@@ -0,0 +1,181 @@
...
+ */
+function writeJson(res: Response, status: number, payload: string): void {
+  res.status(status);
+  res.type("application/json");
+  res.setHeader("Cache-Control", "no-store");
+  res.send(payload);
+}
+
+/** Measure the encoded body before sending so no partial response escapes. */
+function sendJson(res: Response, body: JsonValue): void {
```
MarioCadenas, line 111: **[Medium] The response-size cap is enforced only after full in-memory materialization.** `sendJson` runs `JSON.stringify(body)` then checks `Buffer.byteLength > MAX_RESPONSE_BYTES`, so up to `limit` rows (× 2–3 projected copies) and the entire response string are built in memory before the 413. `MAX_MATERIALIZED_NODES` bounds row *count*, not bytes, so a list over fat text/jsonb columns spikes heap regardless of the cap; N concurrent such requests → GC thrash / OOM even though each is ultimately rejected. Cap row count, or measure incrementally before building the string.

_Automated review finding._

packages/appkit/src/plugins/database/crud/query.ts
```diff
@@ -0,0 +1,472 @@
...
+  if (value === undefined)
+    reject(parameter, "Operand does not match its column type");
+  return value;
+}
+
+/**
+ * Decode the predicate on one column against the operator matrix its kind
+ * allows, counting conditions so a filter cannot grow without bound.
+ */
+function decodeCondition(
```
MarioCadenas, line 177: **[Medium] Reads have no timeout / statement cancellation → connection-pool exhaustion.** `databaseReadDefaults` sets no `timeout`, so there is no `TimeoutInterceptor` and no AbortSignal deadline; a caller-supplied leading-wildcard `like`/`ilike` (accepted here for string columns) forces a sequential scan that holds its pooled connection to completion. N concurrent such requests (≥ pool size) block all reads *and* writes. Add a `statement_timeout` on the connection and/or a read deadline.

_Automated review finding._

packages/appkit/src/plugins/database/crud/contract.ts
```diff
@@ -0,0 +1,243 @@
...
+function compileTable(table: AppKitTable): MutableCrudTable {
+  const columns = new Map<string, CompiledColumn>();
+  const selectable = new Set<string>();
+  const queryable = new Set<string>();
+  let primaryKey: CompiledColumn | undefined;
+
+  for (const meta of Object.values(table.$columns)) {
+    const column = compileColumn(meta);
+    columns.set(meta.columnName, column);
+    if (meta.primaryKey) primaryKey = column;
```
MarioCadenas: **[Low-Medium] A private primary key still powers the detail route.** `primaryKey` is assigned here *before* the `if (meta.isPrivate) continue`, so a `.private()` PK is excluded from `selectable`/`queryable` (good) but still registers `GET /:table/:id` and `decodeId`. The result is an enumeration/existence oracle keyed on the id the author marked private (the value is never echoed and list rows omit it, so clients cannot discover ids — but per-id probing returns 200 vs 404). Consider treating a private-PK table as keyless for HTTP, or requiring a public id.

_Automated review finding._

atilafassina (approved):

atilafassina:
2 possible PII-leak sites in stream logs: drizzle-data-path.ts:134 (primary — carries params) and errors.ts:110 (unclassified path, can also receive a DrizzleQueryError). The two lifecycle.ts sites are lower-risk — schema-validation and select 1 health-check errors carry no row params (setup can carry connection/infra detail, worth trimming eventually, but not the row-PII concern).

packages/appkit/src/plugins/database/database.ts
ditadi: This is on purpose, the first version is SP only. OBO and Lakebase have a weird relationship, and we're not sure we want to encourage it. 

Authentication to the database is via SP, not each user.

packages/appkit/src/plugins/database/crud/routes.ts
ditadi: nice, fixed.

List responses are now encoded row by row against the byte budget, so the 413 fires before further rows are projected or serialized; memory stays bounded by the budget.

packages/appkit/src/plugins/database/crud/query.ts
ditadi: Fixed.

every pooled connection now carries a server-side statement_timeout (30s), so PostgreSQL cancels an expensive pattern and frees the connection; query_canceled (57014) is classified as transient (503).

packages/appkit/src/plugins/database/crud/contract.ts
ditadi: Fixed.

A private primary key no longer registers GET /:table/:id. The table is keyless over HTTP, so there is no per-id existence probe.

ditadi:
> 2 possible PII-leak sites in stream logs: drizzle-data-path.ts:134 (primary — carries params) and errors.ts:110 (unclassified path, can also receive a DrizzleQueryError). The two lifecycle.ts sites are lower-risk — schema-validation and select 1 health-check errors carry no row params (setup can carry connection/infra detail, worth trimming eventually, but not the row-PII concern).

Fixed.

The errors.ts unclassified path now logs only the error's constructor-name chain (never %O), so a DrizzleQueryError can't leak query text or params; the drizzle-data-path site was already fixed in an earlier push.

PR #528, opened 2026-08-10
feat(appkit): add database mutations and transactional hooks

ditadi:
## Follow-up status

The `api` rename, CRUD-on default, configuration diagnostics, transaction predicate fix, and pool timeout forwarding described below are implemented and validated locally, but have not been pushed yet. The current GitHub head may still show `crudRoutes` and opt-in writes.

## Stack

- [#525: database runtime](https://github.com/databricks/appkit/pull/525)
  - [#526: typed plugin API](https://github.com/databricks/appkit/pull/526)
    - [#527: generated reads](https://github.com/databricks/appkit/pull/527)
      - [#528: mutations and hooks](https://github.com/databricks/appkit/pull/528), this PR
        - [#529: playground example](https://github.com/databricks/appkit/pull/529)

This PR now targets `main`.

## What

Adds `create`, `update`, `upsert`, and `delete` to the typed entity API, plus transactional before/after hooks. `database({ schema })` generates full HTTP CRUD by default. No separate write opt-in is required.

The public configuration option is `api`, replacing `crudRoutes`:

```ts
// Full generated CRUD for every declared table.
database({ schema });

// Disable generated routes, preserving the server-side client.
database({ schema, api: false });

// Read-only routes.
database({ schema, api: { writes: false } });

// Full CRUD for selected tables.
database({ schema, api: { tables: ["notes"] } });

// Reads, create, and update without delete.
database({ schema, api: { writes: { operations: ["create", "update"] } } });
```

### Access model

The database plugin executes as the app's service principal in deployed Databricks Apps. OBO is not supported and is not planned for this plugin. Generated routes do not implement per-user or per-row authorization. Every caller who can reach them receives every enabled operation on every exposed row.

CRUD-on is the intended AppKit default, consistent with its opinionated, plug-and-play application model. It does not imply row-level isolation. Applications with narrower access requirements must restrict or disable generated routes and implement authorization in custom server routes. App admission and least-privilege database grants remain application/deployment responsibilities.

Restrictions reject invalid configuration rather than silently enabling more routes. Unknown keys, invalid values, undeclared tables, duplicate names, and unsupported operations fail setup before pool creation. The removed `crudRoutes` option is rejected with migration guidance so an old opt-out cannot silently turn into full CRUD.

### Hooks share the mutation's transaction

A hooked mutation opens its transaction before running `before*`, the write, and `after*`. Writes through `ctx.app.database` join the same transaction and commit or roll back together. This does not extend transactions to other plugins or external services.

An instance-owned `AsyncLocalStorage` keeps concurrent requests and plugin instances isolated. A shared budget caps each transaction at 100 database operations, including direct SQL. Nested mutations reject repeated entity/operation pairs and depth beyond 8 frames.

Transactions have a 30-second callback deadline. PostgreSQL also receives a 30-second `statement_timeout` and a 30-second `idle_in_transaction_session_timeout` through the Lakebase pool configuration. The JavaScript deadline is not cancellation of arbitrary hook code or external side effects.

Before-hook replacement values are revalidated against the trusted schema before persistence. An update or delete that matches no row throws inside the mutation transaction so hook writes roll back before the generated route returns 404.

### Deliberate validation failures

`DatabaseValidationError` produces HTTP 422. Only issues naming public columns are returned, with at most 50 issues. Other hook failures remain opaque server errors.

### HTTP writes and row integrity

Generated write allowlists exclude private columns and server-owned fields. Generated primary keys, including UUID primary keys with `defaultRandom`, cannot be supplied by HTTP callers. Natural primary keys may be supplied on create, but no primary key can be updated over HTTP.

Programmatic upsert excludes primary keys, server-generated fields, and the conflict target from the update values. There is no generated upsert route.

Unknown or read-only write fields are rejected rather than silently discarded. Rejections identify a field only when it is a public column; other fields use the generic `["body"]` path.

### Query state survives transaction redirection

Redirecting an entity mutation to a transaction client now clones the caller's query state. `where()` predicates remain attached to update/delete through hooks and explicit transactions without mutating the shared transaction client. Tests cover all five transaction paths identified in review.

`create` and `upsert` reject accumulated predicates rather than silently ignoring them.

### Diagnostics and documentation

Invalid route names and configuration restrictions now produce actionable server-side setup errors. Errors name the problematic key or table and explain how to fix it. Non-routable tables are not silently omitted: rename them, select a routable subset with `api.tables`, or disable generated routes with `api: false`.

Client messages remain generic, and request boundaries strip setup diagnostics. The new database plugin guide documents the access model, defaults, restrictions, private columns, hooks, and timeout limitations. API references include the exported API configuration types.

### Bounded responses

Success and error responses share the encoded byte budget. Error details are dropped if they exceed it. The row sanitizer uses null-prototype objects so JSON keys such as `__proto__` remain data.

## Local verification

- `pnpm exec vitest run --maxWorkers=4`: 4,778 passed, 1 skipped.
- Focused database and Lakebase suites: 255 passed.
- `pnpm build && pnpm docs:build`: passed.
- `pnpm check:fix && pnpm -r typecheck`: passed; lint reports four existing unused-import/variable warnings outside these changes. Untracked local scratch directories were excluded from formatting.
- The initial test run with default concurrency hit a five-second timeout in an existing type-generator test. The full suite passed with four workers.

No commit or push has been made for the local follow-up yet.


packages/appkit/src/plugins/database/crud/routes.ts
```diff
@@ -145,37 +127,99 @@ export function createListHandler(deps: ReadRouteDeps): ReadHandler {
...
+        sendJson(res, 200, deps.table.projectPublicRow(row));
+      });
+    } catch (error) {
+      sendError(res, deps.table, "write", error);
+    }
+  };
+}
+
+/** `DELETE /:table/:id` — `204` with no body, or 404 when nothing matched. */
+export function createDeleteHandler(deps: CrudRouteDeps): RouteHandler {
```
MarioCadenas, line 212: **[High] Writes are not opt-in separately from reads.** A single `crudRoutes` value registers `GET` **and** `POST`/`PATCH`/`DELETE` for each exposed table; there is no read-only mode and no per-operation gate. Combined with service-principal execution and no per-user/row authorization (see #527), any caller the app admits can create/update/**delete** any row of any exposed table — and a table opted into generated *reads* is silently upgraded to public destructive writes on adopting this PR. Consider a separate write opt-in plus an authorization/owner seam.

_Automated review finding._

packages/appkit/src/plugins/database/lifecycle.ts
```diff
@@ -59,45 +65,71 @@ function buildTransactionClient(context: ExportContext): TransactionClient {
...
+  const active = context.scope.activeTransaction();
+  if (active) return callback(active);
+  return context.getDataPath().transaction(async (txDataPath) => {
+    let open = true;
+    const assertTransactionActive = () => {
+      context.assertActive();
+      if (!open) throw new DatabasePluginError("INTERNAL", "transaction");
+    };
+    // The outer transaction owns execution; inner clients use its connection.
+    const directExecute: EntityExecute = async (operation) => ({
```
MarioCadenas, line 126: **[Medium-High] Hooks run inside a transaction with no per-hook / per-transaction deadline.** The transaction is opened via `getDataPath().transaction()` (not through `ctx.execute`), and `directExecute` runs each op with no interceptors — so there is no AppKit timeout, and there is no PG `statement_timeout` / `idle_in_transaction_session_timeout`. A slow or hanging `before`/`after` hook holds the pooled connection + row/index locks for its full duration (the pool's `connectionTimeoutMillis` bounds *waiters*, not the *holder*); enough concurrent slow-hook writes exhaust the pool. Bound hook/transaction wall-clock.

_Automated review finding._

packages/appkit/src/plugins/database/scope.ts
```diff
@@ -0,0 +1,68 @@
...
+    runWithTransaction<T>(
+      transaction: TransactionClient,
+      run: () => Promise<T>,
+    ): Promise<T> {
+      const frames = storage.getStore()?.frames ?? [];
+      return storage.run({ transaction, frames }, run);
+    },
+
+    /** Open one frame, refusing a repeated entity/operation pair or deep nesting. */
+    async runMutation<T>(
```
MarioCadenas, line 80: **[Medium] The re-entrancy guard bounds nesting depth but not sibling fan-out.** `runMutation` caps nesting (`MAX_MUTATION_DEPTH=8`) and refuses repeated `(entity, operation)` pairs, but sibling writes at one level do not accumulate frames — so a hook can issue an unbounded number of writes in a loop (e.g. `beforeCreate` iterating an attacker-controlled array field), all inside the single no-deadline transaction → an arbitrarily long-held transaction / lock set.

_Automated review finding._

packages/appkit/src/plugins/database/entity-client.ts
```diff
@@ -164,41 +182,208 @@ export class EntityClient {
...
+        "write",
+        (dataPath) => dataPath.delete(this.ctx.table, id, this.state.where),
+        databaseWriteDefaults,
+      );
+    if (this.isDirect("delete")) return remove();
+    return this.mutate(
+      "delete",
+      (entity) => entity.delete(id),
+      async (context) => {
+        await this.runHook(() => hooks?.beforeDelete?.(id, context));
```
MarioCadenas: **[Medium] Before-hook writes commit even when the update/delete matches no row and the route returns 404.** `beforeUpdate`/`beforeDelete` run unconditionally inside the transaction and may write via `ctx.app.database`; when the primary op matches 0 rows it returns `null`/`false` *without throwing*, so the transaction commits the hook's side effects and the 404 is thrown afterward in the route handler. E.g. a `beforeDelete` that writes an audit row on `DELETE /notes/9999` (nonexistent) commits the audit row, then the caller gets 404 — persisted state is inconsistent with the reported outcome. Throw `NOT_FOUND` inside the transaction on a no-match.

_Automated review finding._

packages/appkit/src/plugins/database/crud/contract.ts
```diff
@@ -188,6 +197,13 @@ function compileTable(table: AppKitTable): MutableCrudTable {
     if (filterOperatorsForKind(meta.kind).length > 0) {
       queryable.add(meta.columnName);
     }
+    // A generated identity belongs to the server, never the caller.
+    if (meta.serverGenerated) continue;
```
MarioCadenas: **[Medium] `creatable` lets the client choose the primary key for natural / `defaultRandom` PKs.** `creatable` excludes only `serverGenerated` columns, so a `uuid().primaryKey().defaultRandom()` (or any natural PK) is settable via the create body: `POST /:table {id:"chosen-uuid", …}` overrides the generated default and lets a caller pick primary keys (id-squatting / prediction) — which, with no per-user authz, any caller can do. `id()`/`bigid()` tables are safe. Consider excluding all primary keys from `creatable`.

_Automated review finding._

packages/appkit/src/plugins/database/entity-client.ts
```diff
@@ -164,41 +182,208 @@ export class EntityClient {
...
+        const row = await apply(payload, payload !== values);
+        // Matching no row is an ordinary outcome, not something to react to.
+        if (row) await this.runHook(() => hooks?.afterUpdate?.(row, context));
+        return row;
+      },
     );
   }
 
-  upsert(values: Row, options: { onConflict: string }): Promise<Row> {
+  async upsert(values: Row, options: { onConflict: string }): Promise<Row> {
```
MarioCadenas, line 269: **[Medium] upsert PK-rewrite is reachable through the typed `upsert` (carry-forward from #526).** `beforeUpsert` replacements are re-validated by `$insertSchema`, which still includes natural PKs, and the underlying `DataPath.upsert` reuses the full payload as `ON CONFLICT DO UPDATE SET` — so a natural PK + cross-column `onConflict` rewrites the existing row's PK. Not reachable over HTTP (no upsert route), but a real integrity hole in the programmatic `upsert`. Fix at the runtime (`set` should exclude the PK).

_Automated review finding._

packages/appkit/src/plugins/database/entity-client.ts
```diff
@@ -165,41 +190,267 @@ export class EntityClient {
...
+          dataPath.update(this.ctx.table, id, parsed, this.state.where),
+        byHook,
+      );
+    if (this.isDirect("update")) {
+      return this.directMutation("update", () => apply(values));
+    }
+    try {
+      return await this.mutate(
+        "update",
+        (entity) => entity.updateWithNoMatch(id, values, noMatch),
```
MarioCadenas, line 245: **[High] `where()` predicate is silently dropped on `update`/`delete` when the mutation is redirected through a transaction.** On a hooked table (or when an outer explicit `db.transaction(...)` is active), `isDirect()` returns false, so `mutate()` restarts via `runInTransaction` against a **fresh** `EntityClient` built with default `state = {}` (`lifecycle.ts` `entityOf`). The restarted client's `this.state.where` is `undefined`, so `dataPath.update`/`delete` runs keyed by primary key alone and the accumulated predicate is lost.

This contradicts the PR's headline claim that `where()` now binds `update(id)`/`delete(id)`: `db.notes.where({ tenantId }).update(id, vals)` on a hooked table performs a **cross-scope write/delete** instead of the intended zero-row no-op — a silent authorization-scoping bypass in trusted code. (HTTP write routes always pass an empty `where`, so the generated surface is unaffected; this is the typed API only.)

Untested: the only predicate test (`entity-client.test.ts`, "narrows keyed mutations by the accumulated predicate") uses an *unhooked* harness, so only the direct path is covered — the redirected path slips through.

Fix: thread the predicate onto the restarted client, e.g. `restart = (entity) => (this.state.where ? entity.where(this.state.where) : entity).updateWithNoMatch(id, values, noMatch)` (same for `deleteWithNoMatch`), and add a hooked-table `where().update()`/`.delete()` test asserting the predicate reaches `dataPath`. _Automated review finding._

---

packages/appkit/src/plugins/database/crud/routes.ts
ditadi: Product decision: full CRUD remains enabled by default for this plugin. OBO is not supported and is not planned. Generated routes use the app's service principal and do not provide per-user or per-row authorization; app admission alone is not row-level isolation.

The local follow-up renames `crudRoutes` to `api` and keeps explicit restrictions:

- `api: false` disables generated routes without disabling the server-side client.
- `api: { writes: false }` keeps reads only.
- `api.tables` limits exposed tables; `api.writes.tables` and `api.writes.operations` limit writes.
- Invalid restriction keys/values fail setup rather than silently enabling CRUD. The removed `crudRoutes` option also fails with migration guidance so an old opt-out cannot be ignored.

Apps that need user-specific permissions must restrict/disable these generated routes and implement authorization in custom server routes. The new database plugin guide and config reference state this access model explicitly.

This addresses the missing read-only/per-operation controls while deliberately retaining CRUD-on as the product default, rather than adopting a separate write opt-in. The follow-up is implemented and tested locally but has not been pushed yet; the PR description marks that status.

ditadi: CRUD is enabled by default, with options to disable writes, expose selected tables, or keep selected tables read-only; 

this plugin runs exclusively as the app's service principal, with OBO outside its scope.

packages/appkit/src/plugins/database/lifecycle.ts
ditadi: done, Hook transactions now have a 30-second callback deadline, with PostgreSQL enforcing 30-second statement and idle-in-transaction timeouts.

packages/appkit/src/plugins/database/scope.ts
ditadi: done, A shared 100-operation budget now bounds the entire transaction, including sibling mutations, reads, and direct SQL.

packages/appkit/src/plugins/database/entity-client.ts
ditadi: done, Update/delete misses now trigger rollback inside the mutation transaction, preventing before-hook writes from committing when the generated route returns  404

packages/appkit/src/plugins/database/crud/contract.ts
ditadi: done, Generated primary keys, including defaultRandom() UUIDs, are now excluded from HTTP create payloads, while natural keys remain intentionally writable on create

packages/appkit/src/plugins/database/entity-client.ts
ditadi: done, Upsert now preserves the existing primary key by excluding caller-supplied primary-key values from the conflict update.

packages/appkit/src/plugins/database/entity-client.ts
ditadi: done, Redirected mutations now clone the caller’s query state, preserving where() across hooks and explicit transactions without mutating the shared transaction client.

packages/appkit/src/plugins/database/lifecycle.ts
```diff
@@ -60,45 +73,88 @@ function buildTransactionClient(context: ExportContext): TransactionClient {
...
+  const deadline = new Promise<never>((_, reject) => {
+    timer = setTimeout(
+      () => reject(new DatabasePluginError("TRANSIENT", "transaction")),
+      TRANSACTION_TIMEOUT_MS,
+    );
+    timer.unref?.();
+  });
+  return Promise.race([Promise.resolve().then(run), deadline]).finally(() =>
+    clearTimeout(timer),
+  );
```
atilafassina, line 104: I'm wondering what will happen if a deadline hits and then a `run` crashes. It will be an uncaught error during the transaction still.

if so, should we `catch()` this one?

ditadi: Promise.race already handles rejections from every input promise even after the deadline wins, so a late rejection from run does not become unhandled and no additional .catch() is needed

atilafassina (approved):

PR #529, opened 2026-08-10
feat(playground): add a DatabasePlugin example

ditadi:
## Stack

- [**#525** — the database runtime](https://github.com/databricks/appkit/pull/525) (merged)
  - [**#526** — the typed plugin API](https://github.com/databricks/appkit/pull/526) (merged)
    - [**#527** — generated reads](https://github.com/databricks/appkit/pull/527) (merged)
      - [**#528** — mutations and hooks](https://github.com/databricks/appkit/pull/528) (merged)
        - [**#529** — playground example](https://github.com/databricks/appkit/pull/529) ← **you are here**

The four PRs above have landed, so this one targets `main` and the diff shown here is the whole remaining delta.

## What

The last PR in the series. No SDK behaviour changes here: this is the dev-playground page that runs the whole plugin against a live Postgres, plus one integration test that asserts the assembled stack from the outside.

The page is a small board/annotation app — three tables, a two-edge include, all three hook points, and a private column — chosen so every claim the earlier PRs make is visible rather than described.

## Changes

**The schema (`config/database/schema.ts`)**

`boards → notes → note_events`, which is enough to exercise a nested include. `notes.author_email` is `.private()`, so it is writable from server code and refused on the wire — the page shows both halves of that.

The plugin does not create tables. The app expects them to exist, and the page says so when they do not.

The registration passes no `schema`. #582 made `database()` load the named `schema` export from `config/database/schema.ts`, which is where this file already sits, so the example shows that path rather than importing the schema by hand.

**The registration (`server/index.ts`)**

Registered conditionally on `LAKEBASE_ENDPOINT`, the same way `lakebase()` already is in this file, so the playground still boots without a database configured.

`api: { writes: { tables: ["boards", "notes"] } }` generates reads for all three tables but accepts HTTP writes only on `boards` and `notes`. The audit trail is written by a hook and never through the API, which is the exposure decision from #527 shown rather than explained.

The `notes` hooks use all three points:

- `beforeCreate` runs an **agent** to redact personal names from the body before the insert, so the unredacted text never reaches the table, and stamps the private `author_email`. This is the intended shape of the hook API: a hook is server code and can call any other plugin. The model call is bounded by `AbortSignal.timeout(10_000)`, below the 30-second transaction deadline, and empty or errored output rolls the mutation back rather than falling through to the original text. The agent itself lives in `server/agents/redactor/agent.ts`, the same folder convention the playground's other code agents use.
- `serialize` ships a 120-character preview on the list route and the whole body on the detail route.
- `afterCreate` writes the `note_events` row **inside the insert's own transaction**, so the note and the event describing it commit together or not at all.

**The page (`client/src/routes/database.route.tsx`)**

A board explorer over the generated routes, a panel that traces a write through its hooks, and a probe that fires the requests the plugin is supposed to refuse — writing a private column, an unknown field, an over-large `limit` — and shows the status and body that come back.

Every request the page makes goes through a generated route, including the nested include, which it reaches with `?include=`. An earlier revision carried two hand-written routes alongside them; they are gone, so nothing here competes with the generated API.

**One integration test (`plugins/database/tests/mvp.integration.test.ts`)**

Drives the plugin through its real assembly with only the driver mocked: config in, routes registered, request in, response out. It is the test that would catch a regression the unit suites each miss individually.

## Review feedback

The three automated findings on the first revision are addressed in the code:

- The redaction hook no longer falls back to the unredacted body. Empty or whitespace-only model output throws, so the mutation rolls back.
- The model call inside the transaction is bounded by a 10-second abort signal, and the hook re-checks the signal after the await because a cancelled stream can return accumulated text instead of throwing.
- The two hand-written routes that returned raw driver error text are gone; the page uses the generated routes, which sanitize their errors.

## Verification

- `pnpm test` — 5024 passing, 1 skipped
- Exercised manually against a live Lakebase instance: reads, writes, the hook chain, the redaction agent, the transactional rollback, and each refusal in the probe


apps/dev-playground/server/index.ts
```diff
@@ -384,6 +395,54 @@ createApp({
...
+                //
+                // author_email is a private column: refused on the wire but
+                // writable here. A real app takes it from the session.
+                beforeCreate: async (values) => {
+                  const answer = await runAgent(redactor, {
+                    messages: String(values.body),
+                  });
+                  return {
+                    ...values,
+                    body: answer.text || values.body,
```
MarioCadenas, line 423: **[Medium] The redaction hook silently persists the unredacted body on an empty model response.** `body: answer.text || values.body` falls back to the *original* body whenever `answer.text` is falsy (safety refusal, empty completion, whitespace) — breaking the guarantee stated in this file and the route prose ("the unredacted body must never reach the table"), with no signal. As example code this teaches a redaction control with a silent bypass. Throw (roll back) when redaction yields empty, rather than falling back to the raw input.

_Automated review finding._

apps/dev-playground/server/index.ts
```diff
@@ -384,6 +395,54 @@ createApp({
...
+                // An agent is a plain definition, so a hook runs one by
+                // importing it. This has to happen before the insert: the
+                // unredacted body must never reach the table. It also holds
+                // the transaction open while the model answers, which is the
+                // trade being made here.
+                //
+                // author_email is a private column: refused on the wire but
+                // writable here. A real app takes it from the session.
+                beforeCreate: async (values) => {
+                  const answer = await runAgent(redactor, {
```
MarioCadenas, line 418: **[Medium] The example runs an LLM call inside the mutation transaction with no timeout.** `runAgent(redactor, …)` is awaited inside `beforeCreate`, so it holds a pooled DB connection + the row's locks for the full (unbounded) LLM latency — the exact "slow call inside a transaction" the plugin's own docs warn against (`hook-lifecycle.tsx`: "bound anything that leaves the process"; the route prose: "give it a timeout"). `runAgent` accepts a `signal`, so `AbortSignal.timeout(ms)` is one argument away. As the flagship example this teaches the un-bounded pattern.

_Automated review finding._

apps/dev-playground/server/index.ts
```diff
@@ -485,6 +544,46 @@ createApp({
...
+        app.get("/api/boards/stats", async (_req, res) => {
+          try {
+            const rows = await db.sql<{ slug: string; notes: string }>`
+              select b.slug, count(n.id)::text as notes
+              from boards b left join notes n on n.board_id = b.id
+              group by b.slug order by b.slug
+            `;
+            res.json(rows);
+          } catch (error: unknown) {
+            res.status(500).json({ error: (error as Error).message });
```
MarioCadenas, line 562: **[Medium] Custom playground routes leak raw DB error messages to the client.** Both new routes do `catch (error) { res.status(500).json({ error: (error as Error).message }) }` (here and at the `/timeline` route below), returning raw driver/query error text — constraint names, column names, pg `detail` — to any caller, bypassing the error sanitization the generated routes apply. As de-facto documentation this steers developers to leak internal error detail from hand-written routes; return a generic message and log the detail server-side.

_Automated review finding._

apps/dev-playground/server/index.ts
```diff
@@ -58,6 +65,14 @@ const usersOnly: FilePolicy = (_action, _resource, user) => {
   return user.isServicePrincipal !== true;
 };
 
+// Best-effort redaction demonstrates calling an agent from a hook; it is not
+// a guarantee that all personal data will be removed.
+const redactor = createAgent({
```
MarioCadenas, line 70: this should be in its own file not here 😄

MarioCadenas: under server/agents/redactor/agent.ts

ditadi: solved.

MarioCadenas (approved):

PR #532, opened 2026-08-13
feat(appkit): Agent Skills (v1) — SKILL.md progressive disclosure for agents

MarioCadenas:
## Agent Skills (v1)

Runtime **Agent Skills** for the `agents` plugin — the `SKILL.md` format Claude Code / Cursor use, brought to AppKit agents. Only each skill's `name` + `description` sit in the system prompt (always-on, cheap); the full body loads on demand. Works on **any** Databricks-served model — AppKit implements the progressive disclosure itself, so it doesn't depend on a provider-native skills feature. Fills the seam the loader already reserved (`RESERVED_DIRS = new Set(["skills"])`).

Not to be confused with the dev-time "Databricks Agent Skills" product (Claude Code skills for *building* apps) — this is a runtime capability of deployed agents.

### What a skill is
A directory with a `SKILL.md` (frontmatter `name` + `description`, Markdown body) plus optional bundled reference files. Frontmatter is an Anthropic-format **superset** (also tolerates `license`, `allowed-tools`, `metadata`); unknown keys warn, not error — so skills authored elsewhere drop in.

### How it works
- Every visible skill's `name` + `description` is injected into the agent's system prompt.
- Two read-only built-in tools are added to any agent with a catalog: `load_skill(skill)` returns the body + a manifest of bundled files; `read_skill_file(skill, path)` reads one of those files (through a directory-containment guard).
- The model auto-loads a skill when a task matches; a **user** can force one for a turn with `/skill-name` in chat (or `useAgentChat`'s `send(msg, { skill })`).

### Sources & visibility
- **Global bundle** `config/agents/skills/`, **per-agent** `config/agents/<id>/skills/`, and a **catalog** UC Volume (`skillsVolume` / `DATABRICKS_VOLUME_AGENT_SKILLS`), read as the service principal.
- Per-agent skills are always visible; global skills are **opt-in** via `skills: [...]` frontmatter (or `autoInheritSkills`).
- Name collisions resolve to qualified `<scope>:name`; the bare name errors as ambiguous.

### Commits (phased, independently reviewable)
1. `feat(appkit): load and resolve agent skills from bundle sources` — model, parser, loader, per-agent catalog resolution
2. `feat(appkit): expose skills to agents via prompt catalog and load_skill` — disclosure + built-in tools (bundle e2e)
3. `feat(appkit): source agent skills from a Unity Catalog volume` — catalog source, SP-read
4. `feat(appkit): let users load skills from chat (/skill-name + picker)` — client UX + forced load
[...]

PR #533, opened 2026-08-13
feat(appkit): auto-discover code agents from server/agents/

MarioCadenas:
## What

Every agent is now a **folder** under `server/agents/<id>/`, discovered at startup — `agent.md` for a markdown agent, `agent.ts` (`export default createAgent(...)`) for a code agent. The **folder name is the id**, so the plugin call collapses to `agents()` with no map and no imports, and neither form restates its id.

```diff
- // server/agents/helper.ts
- export const helper = createAgent({ name: 'helper', instructions, tools });
- // server/server.ts
- import { helper } from './agents/helper';
- agents({ agents: { helper } })            // id restated; hand-built map
+ // server/agents/helper/agent.ts          ← id IS the folder name
+ export default createAgent({ instructions, tools });
+ // server/server.ts
+ agents()                                  // no import, no map
```

Markdown and code agents live side by side under the same root; a markdown `agent.md` and code `agent.ts` are just two entry-file kinds.

## How

- **`server/agents` is a fixed convention, not a config knob.** There is no `dir` option to relocate the root — the build glob and the runtime scan reference the *same constant*, so they can't disagree about where agents live. (Discovery has no disable toggle either; a map-only app simply has no `server/agents/` folder.)
- **Code: built-first resolution (NODE_ENV-independent).** Compiled output wins over source, so a bundled server never `import()`s a `.ts` (plain Node can't). Dev (`tsx`) loads `server/agents/<id>/agent.ts`; a build loads the compiled `dist/agents/<id>/agent.js` (or `build/agents`).
- **Markdown: read from source** in both dev and prod (it's data, not compiled).
- **Branded detection.** `createAgent` stamps a non-enumerable `Symbol.for("appkit.agent")`; the loader keeps branded exports and skips helpers/chunks. A folder is an agent iff it holds `agent.md` or `agent.ts`; one holding **both** errors at boot (one kind per id).
- **Prod bundling (the trap this avoids).** A runtime scan of a dynamic path is dropped by the bundler, so the build lists `server/agents/*/agent.ts` as entries and the compiled `dist/agents/*/agent.js` exist for the scan. That wiring now ships in the package (below), not hand-copied into every scaffold.

## New: `@databricks/appkit/tsdown` build preset

A scaffolded app's `tsdown.server.config.ts` is a one-liner:

```ts
import { appkitServerConfig } from '@databricks/appkit/tsdown';
export default appkitServerConfig();
```

`appkitServerConfig()` auto-detects `server/agents/` and adds the entry glob + `clean` only when code agents exist; object overrides merge with intent (`entry` unioned so the glob can't be dropped, `external` composed), and a function form gives full control. The build wiring lives in the package, so it reaches existing apps on upgrade instead of being maintained per-scaffold.

## Backward compatibility

- **Markdown still under `config/agents/`** is read as a **deprecated fallback** — merged per-agent with `server/agents` (new location wins on an id clash), with a one-time deprecation warning. Cross-location sub-agent refs resolve one direction (a parent moved to `server/agents` can reference a child still in `config/agents`).
- `agents({ agents: { ... } })` map still works, one-time deprecation warning; discovery wins on an id clash.
- `createAgent({ name })` still honored (the registry key wins).
- **Cross-kind sub-agents:** a markdown parent can delegate to a code sibling folder and vice versa, resolved by folder id.

## Also in this PR

- Migrates the **dev-playground** (8 agents → `server/agents/<id>/` folders) and the **template** (`helper/agent.ts` + `planner/agent.md`, tsdown one-liner).
- Docs: rewrites the `agents` plugin page for the folder model and adds an **"Adding agents to an existing app"** section (markdown = no build change; code = adopt the preset).
- Removes the `dir` config option entirely — `server/agents` is a fixed convention (this also retires the never-shipped `codeAgentsDir` knob). `config/agents` remains a read-only deprecated fallback.

## Verification

- `pnpm -r typecheck` clean · `oxlint` + `oxfmt` clean · appkit **3246 tests pass** · `pnpm build` + `pnpm docs:build` succeed.
- **E2E against the packed tarball (`databricks-appkit-0.62.0.tgz`)** in a fresh app installed from the `.tgz`:
  - preset build emits `dist/agents/<id>/agent.js` exactly where the runtime probes;
  - **dev** (`tsx`, source `.ts`) and **prod** (`node`, compiled `.js` — no TS loader) both discover the code agent, the markdown agent, and the `config/agents` fallback (deprecation warning fires);
  - `dir` is absent from the published `types.d.ts`; `./tsdown` + `./beta` exports resolve.

## Notes

- `agents` is a **beta** plugin (`@databricks/appkit/beta`) — the config-surface changes ride the beta stability tier.
- The Agent Skills PR (#532) is **stacked on this branch** and merges after it.


packages/appkit/src/core/agent/agent-dirs.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Agent-folder names from a directory listing, sorted. An agent lives in a
+ * subfolder, so directories count — and symlinks-to-directories too (a shared
+ * agent folder linked into the agents dir). Single-sourced so the markdown and
+ * code loaders keep the same folder-selection policy.
+ */
+export function agentDirNames(entries: Dirent[]): string[] {
+  return entries
+    .filter((e) => e.isDirectory() || e.isSymbolicLink())
```
ditadi: this allows symlinks without validation. If a symlink points to `/etc/passwd`, he will try to load the agents from it. 

add a simple base directory validation, and it should be enough

packages/appkit/src/plugins/agents/agents.ts
```diff
@@ -382,44 +430,165 @@ export class AgentsPlugin extends Plugin implements ToolProvider {
...
+    };
+
+    const legacyDir = path.resolve(process.cwd(), LEGACY_MARKDOWN_DIR);
+
+    // Deprecated fallback: scan config/agents first, then server/agents with the
+    // legacy defs added as resolvable sub-agent targets — so a parent already
+    // moved to server/agents can still reference a child left in config/agents
+    // mid-migration. Code agents keep precedence in resolution and in the final
+    // merge (server/agents wins on an id clash).
+    const legacy = await loadAgentsFromDir(legacyDir, {
```
ditadi, line 580: why two loadAgentsFromDir one after another?

packages/appkit/src/core/agent/load-code-agents.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  // findEntryFile's readdir follows a symlinked folder and returns null for
+  // anything that isn't a real directory.
+  const folders = agentDirNames(entries);
+
+  const agents: Record<string, AgentDefinition> = {};
+
+  for (const id of folders) {
+    const entryFile = await findEntryFile(path.join(dir, id), opts.extensions);
```
ditadi: each agent load is a I/O, so if you have 50 agents will be fifth sequential I/O -> parsing -> execution. 

You can parallelize the imports here. Probably will reduce the startup time for apps with too many agents.

packages/appkit/src/plugins/agents/agents.ts
MarioCadenas: this is the legacy one (config/agents) to keep that path working until we remove it

packages/appkit/src/core/agent/agent-dirs.ts
MarioCadenas: Good catch. Rather than add validation, I dropped symlink following entirely — `agentDirNames` now keeps only real directories (symlink support here was speculative and unused in-tree), so a link planted in `server/agents` can't pull agent code from outside the tree. Added a unit test. Fixed in bc75ed3d.

packages/appkit/src/plugins/agents/agents.ts
MarioCadenas: It's the deprecated `config/agents/` fallback: scan legacy first, then `server/agents` with the legacy defs threaded in so a parent already moved to `server/agents` can still resolve a sub-agent left behind in `config/agents` mid-migration (they're sequential because the second scan depends on the first). I've now guarded it behind `existsSync(config/agents)`, so the common case is a single scan and the two-scan only runs when the deprecated dir actually exists. bc75ed3d.

packages/appkit/src/core/agent/load-code-agents.ts
MarioCadenas: Done — imports now run concurrently via `Promise.all`, preserving sorted-folder order (the default-agent fallback registers the first) and the no-loader / import-error semantics. bc75ed3d.

ditadi (approved):

PR #536, opened 2026-08-14
refactor(appkit): let plugins contribute otel span processors

MarioCadenas:
## What

Splits `TelemetryManager` into two phases so a **single** global OpenTelemetry tracer provider can carry both the OTLP trace exporter and span processors contributed by plugins — instead of each SDK (AppKit's and, e.g., MLflow's) racing to register the process-global provider, where the second registration is silently dropped.

- `initialize()` — registers the **meter and logger** providers eagerly (unchanged timing, at app bootstrap). This is deliberate: OTel's metrics API has no lazy proxy, so a counter/histogram bound against the NoOp meter (as connectors and the cache do in their constructors) stays NoOp for the process lifetime. It no longer registers a tracer provider.
- `registerSpanProcessor(sp)` — new; lets a plugin contribute a span processor during `setup()`. Ignored with a warning after `start()`, since a started provider's processors are immutable in OTel JS 2.x.
- `start()` — new; called by core after all plugin `setup()` completes. Builds the global `NodeTracerProvider` with the OTLP processor (when configured) plus every contributed processor, and no-ops when nothing needs tracing. `NodeTracerProvider.register()` reinstates the async-hooks context manager and W3C propagators that `NodeSDK.start()` used to provide, so span nesting across `await`s is preserved.
- `shutdown()` now flushes all three providers.

Also swaps the `@opentelemetry/sdk-node` dependency for `@opentelemetry/sdk-trace-node`, since the tracer provider is now built directly rather than via `NodeSDK`.

## Why

When both an OTLP endpoint and a plugin tracer (MLflow) are active, they each build a `NodeSDK` and call `.start()`, which registers the global tracer provider. OTel's `registerGlobal` uses `allowOverride=false`, so the second registration is rejected — and because AppKit initializes first, MLflow's spans silently route to a provider that has no MLflow processor. OTel 2.x removed `addSpanProcessor()`, so the only way to have both coexist is one provider built after every processor is known — which requires deferring the tracer provider past plugin setup.

Deferring is safe for traces (verified against the OTel source): `ProxyTracer` rebinds tracers obtained before registration, and no span is emitted during `setup()`. It is **not** safe for metrics, hence the eager meter/logger split above.

## Scope

This is the **enabling infrastructure only**. No plugin calls `registerSpanProcessor()` yet — wiring the MLflow plugin to contribute its processor is a follow-up, gated on `mlflow-tracing` exposing its `MlflowSpanProcessor` (it is not exported in 0.1.3).

## Verification

Re-run after merging `main` up to date (past the Biome → oxlint/oxfmt migration in #538):

- `pnpm --filter=@databricks/appkit typecheck` — clean
- Telemetry suite — 44 pass, including new two-phase tests (contributed-only-no-OTLP routing, idempotent `start()`, post-`start()` register ignored, no-op when unconfigured, and a guard that a metric instrument obtained before `start()` still records)
- `knip` — clean; `oxlint` + `oxfmt` — clean

---


MarioCadenas:
Collapsed into #545 for a simpler review/merge handoff during my vacation: #545 now contains both the span-processor refactor (this PR) and the mlflow single-provider fix, based directly on `main`. Closing to avoid a stacked-PR + squash-merge rebase dance. Branch `telemetry-manager-refactor` kept intact; commits are preserved in #545's history.

PR #540, opened 2026-08-19
feat(appkit): add createTestApp, a never-crash mock client

IamGalymzhan:
## Stack

  - [**#540** — createTestApp and the mock client](https://github.com/databricks/appkit/pull/540) ← **you are here**
    - [**#555** — withEnv, context options, and createApiError](https://github.com/databricks/appkit/pull/555)
      - [**#542** — the playground smoke tests](https://github.com/databricks/appkit/pull/542)

Each PR targets the one above it, so the diff shown here is only the delta on top of #530. Review in order.

---

## @databricks/appkit/testing

1. createTestApp() — boots a real app on an ephemeral port
   - app.get/post/put/patch/delete() → native Response
   - app.close() - new method
   - options: plugins, responses, strict, client, env, server: false, nodeEnv, cache
   - per-request: body, headers, obo, signal
2. createMockWorkspaceClient() — never-crash fake; every service path resolves instead of throwing
   - 7 services proxied; config + apiClient seeded as real values, not mocks
   - canned defaults: statementExecution.executeStatement, warehouses.get, warehouses.start, currentUser.me
3. getMock(client, "jobs.getRun") — the typed handle for call assertions
4. createTestPlugin(factory, config) — instantiates a plugin through AppKit's real config merge
5. getListeningPort(server)
6. types: TestApp, CreateTestAppOptions, TestRequestOptions, MockWorkspaceClient, CreateMockWorkspaceClientOptions
7. mockServiceContext() now fakes both the service-principal and user clients
8. removed createConfigurableMockWorkspaceClient


packages/appkit/src/core/lifecycle-manager.ts
```diff
@@ -110,13 +118,89 @@ export class LifecycleManager {
...
+      logger.error(
+        "close() did not complete within the %dms budget (phase in flight: %s): %O",
+        timeoutMs,
+        this.shutdownPhase,
+        err,
+      );
+      // Returning early leaves the singletons in place: the phases are still
+      // running and still own these instances, so dropping the pointers now
+      // would hand the next boot a half-released app.
+      return;
```
pkosiec, line 170: **P2** — On a `close()` timeout we skip `releaseCoreSingletons()` on purpose (phases still run) — but nothing releases it afterward either.

**Why it matters:** the owner count never returns to 0, so the next `createTestApp()` skips its reset and reuses the old app's half-closed singletons. Later tests in the file then break silently.

**Fix:** on timeout, attach `.finally(() => releaseCoreSingletons())` to the still-running teardown so release runs once it finishes.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,404 @@
...
+      delete: (path, o) => request("DELETE", path, o),
+    };
+  } catch (err) {
+    // Teardown must run from the failure path too, or the boot leaks env
+    // mutations and singletons into every later test in the file.
+    if (app) {
+      // No release alongside this: close() drops the claim itself, and a second
+      // release would pull the singletons out from under a live sibling app.
+      try {
+        await app.close();
```
pkosiec: **P2** — Same leak on the boot-failure path: if this `close()` times out, the singleton count is never released.

**Why it matters:** a later boot reuses stale singletons, so tests pass or fail for the wrong reason.

**Fix:** release the count once teardown settles (same as the timeout path above).

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,404 @@
...
+      get: (path, o) => request("GET", path, o),
+      post: (path, o) => request("POST", path, o),
+      put: (path, o) => request("PUT", path, o),
+      patch: (path, o) => request("PATCH", path, o),
+      delete: (path, o) => request("DELETE", path, o),
+    };
+  } catch (err) {
+    // Teardown must run from the failure path too, or the boot leaks env
+    // mutations and singletons into every later test in the file.
+    if (app) {
```
pkosiec, line 421: **P3** — Only the "app was never built" branch is tested. This `if (app)` branch — app built, then a later step throws — is not.

**Why it matters:** the cleanup after a successful `createApp` is untested, so a leak here would go unnoticed.

**Fix:** add a test where a step after `createApp` throws, then assert env + singletons still restore.

packages/appkit/src/testing/index.ts
```diff
@@ -41,6 +43,13 @@
 // through this entry point — the class is otherwise reachable only via a deep
 // path (../core/plugin-context) that is not part of the package's exports map.
 export type { PluginContext } from "../core/plugin-context";
+export {
+  createTestApp,
+  type CreateTestAppOptions,
+  getListeningPort,
```
pkosiec, line 49: **P1** — `getListeningPort` is marked `@internal` (see its definition) but exported here, so it ships as public API in the `.d.ts`.

**Why it matters:** the tag says "don't use me" while the export invites it. Someone depends on it, then a cleanup that trusts the tag breaks them. It hasn't shipped yet, so now is the free window.

**Fix:** drop it from the barrel (deep-import it in the `@tools/test-helpers` shim), or make it real public API and remove the `@internal` tag.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,404 @@
...
+  delete(path: string, options?: TestRequestOptions): Promise<Response>;
+}
+
+/**
+ * Point `ServiceContext.createUserContext` at the harness's mock so an `obo`
+ * request does not construct a real SDK client from `DATABRICKS_HOST`.
+ *
+ * Mirrors the `createUserContextSpy` in `fixtures.ts`; returns its restore.
+ */
+function stubUserContext(client: WorkspaceClient): () => void {
```
pkosiec, line 150: **P2** — This `stubUserContext` and fixtures' `createUserContextSpy` fake the same thing but disagree — and neither matches the real `ServiceContext.createUserContext`. Real + this stub throw on a missing token and set `tokenFingerprint`/`userEmail`; the fixtures spy does neither.

**Why it matters:** tests using `mockServiceContext` miss missing-token bugs, and code that reads `tokenFingerprint` (e.g. Lakebase pool rotation) takes a different path than under `createTestApp` or in production. Same test, different result depending on the helper.

**Fix:** make both fakes agree with each other and the real shape, or share one builder. (Also: this stub's `test-${userId}` isn't the real sha256 fingerprint.)

packages/appkit/src/core/appkit.ts
```diff
@@ -27,10 +29,24 @@ import { isToolProvider, PluginContext } from "./plugin-context";
...
 const logger = createLogger("appkit");
 
+/**
+ * Names a plugin manifest may not use: `createAndRegisterPlugin` installs
+ * exports as **own** properties, and an own property shadows a prototype
+ * method. Only *silently* broken names belong here — a plugin named `close`
+ * would no-op teardown, whereas shadowing an internal method throws
+ * `TypeError` on the next registration and needs no guard.
+ */
+const RESERVED_PLUGIN_NAMES = new Set(["close"]);
```
pkosiec: **P3** — This list is hand-kept and only holds `close`, but it has to match `AppHandle`'s method names — nothing links them.

**Why it matters:** add a future named method to `AppHandle` (e.g. `restart`) and a plugin with that name silently shadows it — the exact bug this list exists to stop.

**Fix:** build the list from the handle's method names, or add a drift test. Also note the `close` reservation in the changelog — it can break an app with a plugin named `close`.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,262 @@
...
+ * Symbols and denied names short-circuit before minting; anything already on the
+ * target (seeded members, `Object.prototype`) wins.
+ *
+ * `ownKeys`/`getOwnPropertyDescriptor` stay at their defaults on purpose —
+ * reporting keys makes `util.inspect` probe each one, minting a mock per probe.
+ */
+function neverCrashGet(namespace: string, mint: (path: string) => Mock) {
+  return (target: Any, prop: Any): Any => {
+    if (typeof prop === "symbol") return Reflect.get(target, prop);
+    if (PASSTHROUGH_DENY.has(prop)) return undefined;
```
pkosiec: **P3** — A method with no declared response resolves `undefined` instead of throwing.

**Why it matters:** a test can pass for the wrong reason — you forget a response, the call returns `undefined`, and the assertion still holds. (TypeScript catches typos, not a missing response.)

**Fix:** optional: add a `strict: true` that throws on undeclared calls (cheap runtime win). See the typed-keys experiment comment for the compile-time version.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  /** Apply the canned defaults (SQL succeeds, warehouse RUNNING). Default true. */
+  defaults?: boolean;
+}
+
+export type MockWorkspaceClient = WorkspaceClient;
+
+/**
+ * Applied beneath caller-supplied `responses`.
+ *
```
pkosiec, line 45: **P3** — This comment will go stale: "13 suites" no longer matches (the real number depends on what you count), and "the first three" breaks if the object is reordered.

**Why it matters:** a count and a position in a comment mislead the next reader as soon as the code moves.

**Fix:** state the rule without the number or the order.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,262 @@
...
+  const configProxy = new Proxy(configTarget, {
+    get: neverCrashGet("config", mint),
+  });
+  const apiClientProxy = new Proxy(apiClientTarget, {
+    get: neverCrashGet("apiClient", mint),
+  });
+
+  /** Memoized; routes facade names onto the same objects, others onto the floor. */
+  let legacy: LegacyClient | undefined;
+  function toLegacyWorkspaceClient(): LegacyClient {
```
pkosiec, line 251: **P3** — `toLegacyWorkspaceClient`'s proxy hand-copies the symbol + `PASSTHROUGH_DENY` guard from `neverCrashGet` instead of reusing it.

**Why it matters:** add a denied key in one place and miss the other — then `await client` hangs again, the exact bug the deny-set prevents.

**Fix:** share one guard between both traps.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,404 @@
...
+  env?: Record<string, string>;
+
+  /** No socket; setup, validation, and teardown still run, request methods throw. */
+  server?: false;
+
+  /**
+   * Defaults to `"test"`. `"development"` is refused — it throws a `RangeError`
+   * in `get-port` on `port: 0`, boots Vite, and relaxes validation.
+   */
+  nodeEnv?: string;
```
pkosiec, line 96: **P3** — `nodeEnv` really only refuses `development`. For what the tests check (resource validation), `test` and `production` behave the same — the harness pins strict validation either way.

**Why it matters:** a public option whose only real effect is a refusal is extra surface to learn and keep.

**Fix:** optional (taste): drop the option and keep just the internal pin + the `development` refusal, unless "boot as production" is a real use case.

docs/docs/plugins/testing.md
```diff
@@ -10,14 +10,129 @@ AppKit ships a testing kit at `@databricks/appkit/testing` so you can test a plu
...
+`app.get/post/put/patch/delete(path, options?)` return a native `Response`, so `expectStream` composes directly with no bridge.
+
+- `body` — a non-string value is JSON-encoded with `content-type: application/json`. A string is sent as-is.
+- `headers` — merged last, so they win over anything the harness set.
+- `obo` — `true` for the default test user, or `{ userId, token, email }`. Same shorthand as `createMockRequest({ obo })`, so a handler using `asUser(req)` resolves that identity.
+- `signal` — forwarded to `fetch`.
+
+### Teardown
+
+The harness binds a socket and installs signal handlers, so **every boot needs a `close()`**. `close()` releases the socket, runs your plugin's `shutdown()` hooks, drops AppKit's singletons, and restores `process.env` to its pre-boot state. It's idempotent.
```
pkosiec: **P3** — The Teardown section leads with `try/finally` + `close()`. But the handle supports `await using` — it implements `Symbol.asyncDispose`, so the app closes at scope exit, even on a thrown error.

**Why it matters:** the top footgun here is forgetting `close()` and leaking a listener. `await using` leaves nothing to forget.

**Fix:** lead with `await using app = await createTestApp(...)`; keep try/finally as the fallback for TS < 5.2 (when `using` landed) or older Node. Optional: a `useTestApp()` afterEach helper.

pkosiec (commented):
Nice job! A few comments:

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/**
+ * The typed assertion path onto a mocked method — facade accessors are SDK-typed,
+ * so `expect(client.jobs.getRun).toHaveBeenCalled()` does not typecheck.
+ *
+ * Minting is idempotent, so this can be called before the code under test runs.
+ * Throws for a non-function member such as `"config.host"`.
+ */
+export function getMockFn(client: MockWorkspaceClient, path: string): Mock {
```
pkosiec: **P1** — Rename `getMockFn` -> `getMock` before it ships.

**Why it matters:** it returns vitest's `Mock` type, so the name should match — and it lines up with the kit's other `*Mock*` names. "Fn" is redundant (a `Mock` is already a function). This symbol is new in this PR (not in 0.62.0), so now is the free window; after it ships a rename breaks callers.

**Fix:** `getMock(client, path): Mock`. Not `getSpy` — that implies a `Spy` return type, but the type is vitest's `Mock`. The rename must also update the docs + JSDoc that reference `getMockFn`.

docs/docs/plugins/testing.md
```diff
@@ -160,10 +279,61 @@ The kit re-exports the request/response/context fixtures AppKit uses internally:
...
+`createTestApp` installs one of these for you, so reach for it directly only when you're driving a plugin through `createTestPluginContext` or `mockServiceContext`.
+
+How it works, and what to expect:
+
+- The **facade is typed**, so `client.jbos` is a compile error. AppKit owns that 9-member interface, so it's a closed set, not an open-ended chase of the SDK.
+- Each **service** is a proxy that mints a memoized mock per method. `client.jobs.getRun === client.jobs.getRun`, so call assertions are stable, and `toLegacyWorkspaceClient()` shares the same functions — one `responses` entry covers both views.
+- `config.host` is a real **string** (not a mock), because AppKit builds URLs from it. `apiClient.userAgent()` is synchronous for the same reason, and `apiClient.request` resolves `{}` so destructuring its result doesn't throw.
+- Sensible defaults are built in: SQL statements succeed, warehouses report `RUNNING`, and `currentUser.me()` returns a service user. Pass `defaults: false` to script everything yourself.
+
+:::caution The honest catch
```
pkosiec: **P1** — The admonition title `:::caution The honest catch` reads as AI-generated — a cute label, not a descriptive one. Prefer a plain title, e.g. `:::caution Undeclared methods return undefined`.

**Why it matters:** the docs are the kit's UX; a florid label undercuts trust and makes the page harder to scan.

**Fix:** retitle it. Also, while here: the frontmatter `sidebar_position: 8` collides with jobs.md and manifest.md (both 8), so the sidebar order is unstable — give it a unique position.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,404 @@
...
+    setupDatabricksEnv({
+      DATABRICKS_WORKSPACE_ID: "test-workspace-id",
+      ...env,
+    });
+
+    claimAppKitSingletons();
+
+    // Boot runs ServiceContext.createContext for real, which reads
+    // currentUser.id — the mock's built-in default is what lets it through.
+    const client = suppliedClient ?? createMockWorkspaceClient({ responses });
```
pkosiec: **P3** — `responses` only feeds the built-in mock (`suppliedClient ?? createMockWorkspaceClient({ responses })`). Passing both `client` and `responses` silently drops the `responses` (your passed client is used as-is — its own responses are untouched, but the top-level ones do nothing).

**Why it matters:** a silent no-op invites "why isn't my response taking effect?" The kit already throws on a similar conflict (`server: false` + a server plugin), so the same guard fits.

**Fix:** throw when both `client` and `responses` are passed, telling the caller to configure responses on their own client.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,262 @@
...
+
+type LegacyClient = ReturnType<WorkspaceClient["toLegacyWorkspaceClient"]>;
+
+/** Options for {@link createMockWorkspaceClient}. */
+export interface CreateMockWorkspaceClientOptions {
+  /**
+   * Responses keyed by dotted path (`"jobs.getRun"`). A function value is called
+   * with the arguments, so a test can script behaviour or reject.
+   */
+  responses?: Record<string, Any>;
```
pkosiec, line 23: **P3** — Experiment worth a spike: type the `responses` keys instead of `Record<string, any>`. A template-literal union over the facade (`` `${service}.${method}` ``) would give autocomplete on the dotted paths and turn a typo'd key into a compile error.

**Why it matters:** it kills the whole stringly-typed seam (a typo'd key silently does nothing today), and the same idea helps the `getMock` path arg. Biggest DX win available on the mock.

**Fix:** prototype and measure — mapped/conditional types can hurt IDE perf and produce ugly errors, and it partly fights the "undeclared -> undefined" design, so gate it behind the experiment before committing.

docs/docs/plugins/testing.md
```diff
@@ -10,14 +10,129 @@ AppKit ships a testing kit at `@databricks/appkit/testing` so you can test a plu
 
 Exercise a plugin's real code paths — route registration, cross-plugin tool dispatch, user-scoped (on-behalf-of) execution, and per-call timeouts — against a real `PluginContext` with only its outer edges faked. Nothing about the context is reimplemented, so a test can't drift from production behavior.
```
pkosiec: **P1** — Do a deslop pass over this guide (and the kit's JSDoc). "The honest catch" is the clearest tell, but the AI-generated register shows up elsewhere too.

**Why it matters:** the docs are the kit's first impression; the florid voice reads as machine-written and undercuts trust.

**Fix:** run the repo's `deslop` skill over `testing.md` and the testing-kit JSDoc, then a human read for plain, direct phrasing.

packages/appkit/src/testing/reset.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Drop the process-wide singletons `createApp()` initializes, ignoring how many
+ * apps are live.
+ *
+ * A pointer drop, not teardown — close first or the old app's pools and
+ * exporters leak. `app.close()` does both, so this is only for tests that
+ * hand-roll `createApp`.
+ */
+export function resetAppKitSingletons(): void {
```
pkosiec: **P1** — Rename `resetAppKitSingletons` — "AppKit" is redundant inside `@databricks/appkit/testing`, and "Singletons" leaks the implementation.

**Why it matters:** the name should say what it does, not how it's built. vitest's own global resets are `resetAllMocks`/`resetModules`/`unstubAllGlobals` — always `reset<Scope>` with a plain scope noun, never product-prefixed. This symbol is new in this PR (not in 0.62.0), so a rename is still free.

**Fix:** rename to `resetGlobalState()`. Avoid `dispose*` — it drops pointers, it doesn't release resources. Update the doc reference too.

packages/appkit/src/core/lifecycle-manager.ts
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

packages/appkit/src/testing/index.ts
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

packages/appkit/src/core/appkit.ts
IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

docs/docs/plugins/testing.md
IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

docs/docs/plugins/testing.md
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

docs/docs/plugins/testing.md
IamGalymzhan: Done.

packages/appkit/src/testing/reset.ts
IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done — `strict: true` on the mock, and plumbed through `createTestApp`. Throws on call rather than mint so `getMock` still hands back a handle; canned defaults count as declared so a boot works unchanged.

pkosiec (commented):
Thanks for fixing previous comments! Here's another pass with a couple more comments:

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+  delete(path: string, options?: TestRequestOptions): Promise<Response>;
+}
+
+/**
+ * Point `ServiceContext.createUserContext` at the harness's mock so an `obo`
+ * request does not construct a real SDK client from `DATABRICKS_HOST`.
+ *
+ * Mirrors the `createUserContextSpy` in `fixtures.ts`; returns its restore.
+ */
+function stubUserContext(client: WorkspaceClient): () => void {
```
pkosiec, line 150: **P1** — If two test apps are open at the same time, closing one breaks the fake login for the other. Every `createTestApp` installs one shared spy on `ServiceContext.createUserContext`. `close()` calls `spy.mockRestore()`, which puts back the **real** function — not the other app's spy. So after one app closes, the other app's OBO calls run the real Databricks SDK and try to reach the network.

**Why it matters:** the kit promises "no workspace, no network." This quietly breaks it. The two tests that open two apps at once (create-test-app.test.ts:223 and :540) only check ports and env, so they never catch it.

**Fix:** make the harness one app at a time — throw if a second app boots while one is still open (see the related comment on `app.client`). That makes this impossible.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+      throw new Error(
+        "createTestApp: `responses` configures the built-in mock client, so it " +
+          "does nothing when you also pass `client`. Drop `responses` and " +
+          "configure them on your own client instead.",
+      );
+    }
+
+    // Boot runs ServiceContext.createContext for real, which reads
+    // currentUser.id — the mock's built-in default is what lets it through.
+    const client = suppliedClient ?? createMockWorkspaceClient({ responses });
```
pkosiec: **P1** — If a second app opens while the first is still open, the second app's `client` and `responses` are silently ignored. When one app is already live the shared state isn't rebuilt, so `ServiceContext.initialize` hands back the first app's instance (service-context.ts:56) and `CacheManager.getInstance` does the same (cache/index.ts:124). The second app's handlers end up using the **first** app's client. But `app.client` still returns the second mock, so `getMock(app.client, "jobs.getRun")` shows 0 calls even though the handler did call it — on the other client.

**Why it matters:** the docs say `app.client` is exactly what the handler uses (testing.md:78). That isn't true after the first app, so a test can pass or fail for the wrong reason.

**Fix:** allow only one app at a time. Add a module-level `harnessAppLive` flag (one per test file), throw a clear error at the top of `createTestApp` if it's already set, and clear it when the app closes and on the boot-failure paths. Change the two concurrent-app tests (create-test-app.test.ts:223 and :540) to open apps one after another, and add one that checks a second open throws. Also document that the data plane is shared (see the docs comment).

packages/appkit/src/core/lifecycle-manager.ts
```diff
@@ -110,13 +118,95 @@ export class LifecycleManager {
...
+        timeoutMs,
+        this.shutdownPhase,
+        err,
+      );
+      // Not now — the phases are still running and still own these instances,
+      // so dropping the pointers here would hand the next boot a half-released
+      // app. But release once they settle, or the refcount never returns to
+      // zero and every later boot skips its reset.
+      void this.runOnce()
+        .finally(() => releaseCoreSingletons())
```
pkosiec: **P2** — On a close timeout, the cleanup is put off: `void this.runOnce().finally(() => releaseCoreSingletons())`. So `close()` returns while the app still holds its slot. If the next app opens before the old cleanup finishes, it reuses the closing app's shared state — then the old cleanup reaches phase 5, closes the shared `InMemoryStorage`, and wipes the new app's cache mid-test.

**Why it matters:** one slow close quietly corrupts a later test in the same file — the exact bug the refcount is meant to stop.

**Fix:** release right away on timeout, like the normal path (`releaseCoreSingletons(); return;`). It's safe: `runPhases` already saved its own copies of the cache and telemetry (`capturedCache`/`capturedTelemetry`) and never reads the shared state again, so the background cleanup doesn't need it. It's also less code. Update the comment and the test that checks the delayed release.

packages/appkit/src/testing/tests/published-surface.integration.test.ts
```diff
@@ -0,0 +1,90 @@
+import {
```
pkosiec, line 2: **P2** — This is the only CI test that imports from `@databricks/appkit/testing`, but it only pulls in `createTestApp`, `expectStream`, and `getMock`. Everything else — the main `createTestPluginContext`, all the fixtures, all the types — could be deleted from `index.ts` and this test would still pass.

**Why it matters:** the name says it guards the public surface, but a headline export could quietly disappear from the package with CI still green. That's false confidence.

**Fix:** do `import * as testing from "@databricks/appkit/testing"` and check each expected name is there, or at least call `createTestPluginContext` through the entry.

template/server/example.test.ts
```diff
@@ -1,5 +1,5 @@
-import { Plugin, type PluginManifest } from '@databricks/appkit';
-import { expectStream, createTestPluginContext } from '@databricks/appkit/testing';
+import { Plugin, type PluginManifest, toPlugin } from '@databricks/appkit';
```
pkosiec, line 1: **P3** — No CI runs this file. `template/` isn't in `pnpm-workspace.yaml`, the root vitest config skips `**/template/**`, and the CI template job only checks that manifests and deps are in sync — it never runs the test.

**Why it matters:** the example can quietly rot. If the kit's API changes, nothing fails, and someone who scaffolds a new app gets a broken example. Together with the published-surface test above, nothing checks that `createTestPluginContext` stays exported.

**Fix:** either run the scaffolded template's `vitest run` in CI after `npm install`, or move a `createTestPluginContext`-through-the-entry check into the published-surface test.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
+/**
+ * Boot a real AppKit app with no workspace, credentials, or network, then call it
+ * over real HTTP.
+ */
+
+import { createHash } from "node:crypto";
```
pkosiec: **P3** — `createHash` (line 6) and `AuthenticationError` (line 20) are imported but not used here anymore — that logic moved into `fakeUserContext` in fixtures.ts.

**Why it matters:** dead imports; oxlint isn't flagging them.

**Fix:** remove both.

docs/docs/plugins/testing.md
```diff
@@ -1,26 +1,152 @@
...
+    "apiClient.request": { results: [] },
+  },
+});
+```
+
+A function value receives the call arguments, so you can script per-argument behavior or reject to test an error path. `responses` configures the built-in mock, so passing it alongside your own `client` is rejected rather than silently ignored — configure the responses on that client instead. Any path you **don't** declare resolves `undefined` rather than crashing — see [Mocking Databricks services](#mocking-databricks-services) for the trade-off it makes.
+
+For the response *shapes*, follow the service types on the Databricks SDK. The kit doesn't validate them, so a wrong shape fails in your plugin, not in the fake.
+
+`app.client` is the very object your handler resolves at runtime — reached inside a plugin via `getExecutionContext().client` — so you can assert calls on it:
```
pkosiec: **P3** — This says `app.client` "is the very object your handler resolves at runtime," and only later (line 187) says the cache is shared across the process. The client and the OBO `createUserContext` spy are shared the same way, but that isn't said.

**Why it matters:** that missing half is exactly what makes the two concurrency issues above surprising.

**Fix:** extend the "shared across the process" note to include the client and the OBO spy, and limit the `app.client` line to the one open app.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,280 @@
...
+  },
+  "warehouses.get": { state: "RUNNING" },
+  "warehouses.start": undefined,
+  "currentUser.me": {
+    id: "test-service-user",
+    userName: "test-service-user",
+  },
+};
+
+/** The seven generically-proxied services; `config`/`apiClient` are seeded below. */
```
pkosiec: **P3** — "The seven generically-proxied services" writes a count that goes stale the moment the list changes. Same in testing.md:317 — "AppKit owns that 9-member interface." The exact number is easy to get wrong (9 or 10, depending on whether you count `toLegacyWorkspaceClient`), and it drifts if the interface changes.

**Why it matters:** a number in a comment is wrong as soon as the code changes, and misleads the next reader. The point ("it's a closed set, so a typo is a compile error") holds without it.

**Fix:** describe the rule without the count.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+import { createApp } from "../core/appkit";
+import { AuthenticationError } from "../errors";
+import type { WorkspaceClient } from "../workspace-client";
+import type { OboOption } from "./fixtures";
+import { fakeUserContext, oboHeaders, setupDatabricksEnv } from "./fixtures";
+import type { CreateMockWorkspaceClientOptions } from "./mock-workspace-client";
+import { createMockWorkspaceClient } from "./mock-workspace-client";
+import { claimAppKitSingletons, releaseAppKitSingletons } from "./reset";
+
+type Any = any;
```
pkosiec, line 28: **P3** — `type Any = any` is copied into three kit files, but only fixtures.ts explains why (repo-wide `noExplicitAny` is off, so the alias flags the intent). The other two are bare.

**Why it matters:** someone reading the bare `Any` has no idea what it means or why it isn't just `any`.

**Fix:** put the alias (with its comment) in one shared kit file and import it, or just use `any`.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,280 @@
...
+    const out: Record<string, Any> = {};
+    for (const [key, value] of Object.entries(merged)) {
+      if (key.startsWith(prefix)) out[key.slice(prefix.length)] = value;
+    }
+    return out;
+  }
+
+  // `host` is read as a string and throws if falsy, so it cannot be a mock.
+  const configTarget: Record<string, Any> = {
+    host: "https://test.databricks.com",
```
pkosiec, line 203: **P3** — Seeding an `apiClient.*` value through `responses` wraps it in `vi.fn().mockResolvedValue(value)`, so `responses: { "apiClient.userAgent": "x" }` returns a Promise. That's the very thing the sync default avoids — a Promise turns into `"[object Promise]"` inside a Header. The `config.*` path sets values directly and doesn't have this problem, so the two paths behave differently.

**Why it matters:** rare input, but a wrong header value is hard to track down.

**Fix:** for `apiClient.*` seeds, keep a plain value synchronous like the `config.*` path does, or note that only async members can be seeded this way.

packages/appkit/src/testing/index.ts
```diff
@@ -68,6 +75,14 @@ export {
   type TestContextOptions,
   useServiceContextMock,
 } from "./fixtures";
+export {
+  createMockWorkspaceClient,
+  type CreateMockWorkspaceClientOptions,
+  getMock,
```
pkosiec, line 81: **P3** — The kit mixes two words for test doubles: helpers say "Mock" (`createMockWorkspaceClient`, `mockServiceContext`) but these types say "Fake" (`FakeProvider`, `FakeProviders`, `FakeToolResponse`). `Fake*` shipped in 0.64.0, so renaming would break people.

**Why it matters:** two words for the same idea makes a user wonder if they differ, and searching for "mock" misses the Fake types.

**Fix:** add one doc line on the difference (a fake stands in and works; a mock records calls), or export `Mock*` aliases alongside. Don't rename.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+      try {
+        await app.close();
+      } catch {
+        // The boot error is the interesting one; don't let teardown mask it.
+      }
+    } else {
+      // Nothing was booted, so nothing else will drop the claim taken above.
+      releaseAppKitSingletons();
+    }
+    restoreUserContext?.();
```
pkosiec, line 434: **P3** — The failure-path cleanup calls plain `await app.close()`, which ignores the `closeTimeoutMs` the caller passed — the normal `close` path does pass it through.

**Why it matters:** a slow shutdown on the failure path can run past the timeout the caller set.

**Fix:** pass the same `closeTimeoutMs` here.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+          "remove the plugin to boot without a socket.",
+      );
+    }
+    const bootPlugins = [...plugins] as Plugins;
+    if (serverOption !== false && !hasServer) {
+      const { server: serverPlugin } = await import("../plugins/server");
+      bootPlugins.push(serverPlugin({ port: 0, host: "127.0.0.1" }));
+    }
+
+    // Both extras are load-bearing: without explicit storage the cache builds its
```
pkosiec: **P3** — The new kit code is densely commented (about a quarter to a third of the lines), and many comments are written like a story. A few examples: "load-bearing" (here and mock-workspace-client.ts:66), "pull the singletons out from under a still-live sibling app" (create-test-app.ts:314 and :398), "broke the second app" (lifecycle-manager.ts:144), "hand the next boot a half-released app" (lifecycle-manager.ts:166).

**Why it matters:** round 1 desloped the docs for the same reason. Much of this "why" is worth keeping — it explains real, non-obvious singleton and proxy behavior — but the story voice and the length make the code hard to skim.

**Fix:** do a deslop pass on the kit's code comments, like the docs pass. Keep the load-bearing why, cut it to one plain line, and drop any comment that just restates the code.

packages/appkit/src/testing/create-test-app.ts
pkosiec, line 150: **P1** — If two test apps are open at the same time, closing one breaks the fake login for the other. Every `createTestApp` installs one shared spy on `ServiceContext.createUserContext`. `close()` calls `spy.mockRestore()`, which puts back the **real** function — not the other app's spy. So after one app closes, the other app's OBO calls run the real Databricks SDK and try to reach the network.

**Why it matters:** the kit promises "no workspace, no network." This quietly breaks it. The two tests that open two apps at once (create-test-app.test.ts:223 and :540) only check ports and env, so they never catch it.

**Fix:** make the harness one app at a time — throw if a second app boots while one is still open (see the related comment on `app.client`). That makes this impossible.

packages/appkit/src/testing/create-test-app.ts
pkosiec: **P1** — If a second app opens while the first is still open, the second app's `client` and `responses` are silently ignored. When one app is already live the shared state isn't rebuilt, so `ServiceContext.initialize` hands back the first app's instance (service-context.ts:56) and `CacheManager.getInstance` does the same (cache/index.ts:124). The second app's handlers end up using the **first** app's client. But `app.client` still returns the second mock, so `getMock(app.client, "jobs.getRun")` shows 0 calls even though the handler did call it — on the other client.

**Why it matters:** the docs say `app.client` is exactly what the handler uses (testing.md:78). That isn't true after the first app, so a test can pass or fail for the wrong reason.

**Fix:** allow only one app at a time. Add a module-level `harnessAppLive` flag (one per test file), throw a clear error at the top of `createTestApp` if it's already set, and clear it when the app closes and on the boot-failure paths. Change the two concurrent-app tests (create-test-app.test.ts:223 and :540) to open apps one after another, and add one that checks a second open throws. Also document that the data plane is shared (see the docs comment).

packages/appkit/src/core/lifecycle-manager.ts
pkosiec: **P2** — On a close timeout, the cleanup is put off: `void this.runOnce().finally(() => releaseCoreSingletons())`. So `close()` returns while the app still holds its slot. If the next app opens before the old cleanup finishes, it reuses the closing app's shared state — then the old cleanup reaches phase 5, closes the shared `InMemoryStorage`, and wipes the new app's cache mid-test.

**Why it matters:** one slow close quietly corrupts a later test in the same file — the exact bug the refcount is meant to stop.

**Fix:** release right away on timeout, like the normal path (`releaseCoreSingletons(); return;`). It's safe: `runPhases` already saved its own copies of the cache and telemetry (`capturedCache`/`capturedTelemetry`) and never reads the shared state again, so the background cleanup doesn't need it. It's also less code. Update the comment and the test that checks the delayed release.

packages/appkit/src/testing/tests/published-surface.integration.test.ts
pkosiec, line 2: **P2** — This is the only CI test that imports from `@databricks/appkit/testing`, but it only pulls in `createTestApp`, `expectStream`, and `getMock`. Everything else — the main `createTestPluginContext`, all the fixtures, all the types — could be deleted from `index.ts` and this test would still pass.

**Why it matters:** the name says it guards the public surface, but a headline export could quietly disappear from the package with CI still green. That's false confidence.

**Fix:** do `import * as testing from "@databricks/appkit/testing"` and check each expected name is there, or at least call `createTestPluginContext` through the entry.

template/server/example.test.ts
pkosiec, line 1: **P3** — No CI runs this file. `template/` isn't in `pnpm-workspace.yaml`, the root vitest config skips `**/template/**`, and the CI template job only checks that manifests and deps are in sync — it never runs the test.

**Why it matters:** the example can quietly rot. If the kit's API changes, nothing fails, and someone who scaffolds a new app gets a broken example. Together with the published-surface test above, nothing checks that `createTestPluginContext` stays exported.

**Fix:** either run the scaffolded template's `vitest run` in CI after `npm install`, or move a `createTestPluginContext`-through-the-entry check into the published-surface test.

packages/appkit/src/testing/create-test-app.ts
pkosiec: **P3** — `createHash` (line 6) and `AuthenticationError` (line 20) are imported but not used here anymore — that logic moved into `fakeUserContext` in fixtures.ts.

**Why it matters:** dead imports; oxlint isn't flagging them.

**Fix:** remove both.

docs/docs/plugins/testing.md
pkosiec: **P3** — This says `app.client` "is the very object your handler resolves at runtime," and only later (line 187) says the cache is shared across the process. The client and the OBO `createUserContext` spy are shared the same way, but that isn't said.

**Why it matters:** that missing half is exactly what makes the two concurrency issues above surprising.

**Fix:** extend the "shared across the process" note to include the client and the OBO spy, and limit the `app.client` line to the one open app.

packages/appkit/src/testing/mock-workspace-client.ts
pkosiec: **P3** — "The seven generically-proxied services" writes a count that goes stale the moment the list changes. Same in testing.md:317 — "AppKit owns that 9-member interface." The exact number is easy to get wrong (9 or 10, depending on whether you count `toLegacyWorkspaceClient`), and it drifts if the interface changes.

**Why it matters:** a number in a comment is wrong as soon as the code changes, and misleads the next reader. The point ("it's a closed set, so a typo is a compile error") holds without it.

**Fix:** describe the rule without the count.

packages/appkit/src/testing/create-test-app.ts
pkosiec, line 28: **P3** — `type Any = any` is copied into three kit files, but only fixtures.ts explains why (repo-wide `noExplicitAny` is off, so the alias flags the intent). The other two are bare.

**Why it matters:** someone reading the bare `Any` has no idea what it means or why it isn't just `any`.

**Fix:** put the alias (with its comment) in one shared kit file and import it, or just use `any`.

packages/appkit/src/testing/index.ts
pkosiec, line 81: **P3** — The kit mixes two words for test doubles: helpers say "Mock" (`createMockWorkspaceClient`, `mockServiceContext`) but these types say "Fake" (`FakeProvider`, `FakeProviders`, `FakeToolResponse`). `Fake*` shipped in 0.64.0, so renaming would break people.

**Why it matters:** two words for the same idea makes a user wonder if they differ, and searching for "mock" misses the Fake types.

**Fix:** add one doc line on the difference (a fake stands in and works; a mock records calls), or export `Mock*` aliases alongside. Don't rename.

pkosiec (commented):
Thanks for fixing previous comments! Here's another pass with a couple more comments:

packages/appkit/src/testing/create-test-app.ts
pkosiec: **P3** — The new kit code is densely commented (about a quarter to a third of the lines), and many comments are written like a story. A few examples: "load-bearing" (here and mock-workspace-client.ts:66), "pull the singletons out from under a still-live sibling app" (create-test-app.ts:314 and :398), "broke the second app" (lifecycle-manager.ts:144), "hand the next boot a half-released app" (lifecycle-manager.ts:166).

**Why it matters:** round 1 desloped the docs for the same reason. Much of this "why" is worth keeping — it explains real, non-obvious singleton and proxy behavior — but the story voice and the length make the code hard to skim.

**Fix:** do a deslop pass on the kit's code comments, like the docs pass. Keep the load-bearing why, cut it to one plain line, and drop any comment that just restates the code.

packages/appkit/src/testing/mock-workspace-client.ts
```diff
@@ -0,0 +1,280 @@
...
+    ...seededOverrides("config"),
+  };
+
+  // userAgent() must be synchronous (a Promise stringifies to "[object Promise]"
+  // inside a Headers value); request resolves {} so destructuring works.
+  const apiClientTarget: Record<string, Any> = {
+    userAgent: vi.fn().mockReturnValue("appkit-test/1.0"),
+    request: vi.fn().mockResolvedValue({}),
+  };
+  for (const [key, value] of Object.entries(seededOverrides("apiClient"))) {
```
pkosiec, line 218: **P3** — Seeding an `apiClient.*` value through `responses` wraps it in `vi.fn().mockResolvedValue(value)`, so `responses: { "apiClient.userAgent": "x" }` returns a Promise. That's the very thing the sync default avoids — a Promise turns into `"[object Promise]"` inside a Header. The `config.*` path sets values directly and doesn't have this problem, so the two paths behave differently.

**Why it matters:** rare input, but a wrong header value is hard to track down.

**Fix:** for `apiClient.*` seeds, keep a plain value synchronous like the `config.*` path does, or note that only async members can be seeded this way.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,412 @@
...
+      delete: (path, o) => request("DELETE", path, o),
+    };
+  } catch (err) {
+    // Teardown must run from the failure path too, or the boot leaks env
+    // mutations and singletons into every later test in the file.
+    if (app) {
+      // No release alongside this: close() drops the claim itself, and a second
+      // release would pull the singletons out from under a live sibling app.
+      try {
+        await app.close();
```
pkosiec: **P3** — The failure-path cleanup calls plain `await app.close()`, which ignores the `closeTimeoutMs` the caller passed — the normal `close` path does pass it through.

**Why it matters:** a slow shutdown on the failure path can run past the timeout the caller set.

**Fix:** pass the same `closeTimeoutMs` here.

packages/appkit/src/core/lifecycle-manager.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/tests/published-surface.integration.test.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

docs/docs/plugins/testing.md
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

packages/appkit/src/testing/mock-workspace-client.ts
IamGalymzhan: Done.

packages/appkit/src/testing/index.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

packages/appkit/src/testing/create-test-app.ts
```diff
@@ -0,0 +1,442 @@
...
+  client: WorkspaceClient;
+  /** e.g. `http://127.0.0.1:54321`. Throws when `server: false`. */
+  baseUrl: string;
+  /** The bound ephemeral port. Throws when `server: false`. */
+  port: number;
+  /** The underlying HTTP server, or `undefined` with `server: false`. */
+  server?: Server;
+
+  /** Release the app and restore env. Idempotent. */
+  close(): Promise<void>;
```
pkosiec, line 134: **Suggestion (non-blocking)** — The kit already ships `useServiceContextMock` (fixtures.ts), which registers `beforeEach` and `afterEach` for you so a suite gets a fresh mock per test with no manual teardown. `createTestApp` has no matching helper. So a suite that boots an app in every test has to wire `beforeEach` and `afterEach` by hand, and remember to call `close()` in the `afterEach`.

**Why it matters:** it's a consistency gap. One helper in the kit does the hook wiring for you, the other makes you write it yourself and risk a forgotten `close()`. (`await using` handles a single test well, but it can't carry an app from a `beforeEach` into the test body.)

**Fix (non-blocking):** add a `useTestApp(options)` that mirrors `useServiceContextMock`. It boots before each test, closes after each test, and returns a live accessor like `useServiceContextMock`'s `{ current }`. That way both helpers work the same way.

pkosiec (approved):
🚀 LGTM - final minor comment:

template/server/example.test.ts
IamGalymzhan: Done.

IamGalymzhan: Done.

tools/test-helpers.ts
```diff
@@ -22,9 +25,17 @@ export {
   createTestPluginContext,
   expectStream,
   mockServiceContext,
+  createTestApp,
```
calvarjorge: Why are we adding new exports to this file if we're deprecating it?
Also, can we make sure we're no longer using this package within appkit?

packages/appkit/src/cache/index.ts
```diff
@@ -58,6 +58,8 @@ export class CacheManager {
   private readonly name: string = "cache-manager";
   private static instance: CacheManager | null = null;
   private static initPromise: Promise<CacheManager> | null = null;
+  /** Bumped by {@link reset} so an in-flight init cannot publish over it. */
+  private static generation = 0;
```
calvarjorge, line 62: I wonder whether `CacheManager` should be a singleton. I know this is not something introduced in this PR but we're increasing its complexity because it's a singleton instead of an instance class.

Maybe it's worth considering changing singleton classes that appkit plugins rely on (mainly `ServiceContext` and this one) into fields of the PluginContext instance class. 

Instead of accessing the cache using `CacheManager.getInstanceSync()` we would do `context.cache`.  The benefit would be that the code is clearer, and multiple appkit instances can be created in the same process.

This would be a breaking change, so we have to discuss with the team first (ie, probably not something to implement in this PR, but worth looking into)

calvarjorge (commented):
Can we add tests to the `dev-playground` (or modify the existing ones) as an example of how customers are meant to use the testing package?

tools/test-helpers.ts
IamGalymzhan: Fixed

packages/appkit/src/cache/index.ts
IamGalymzhan: It's out of the scope of this PR. I am looking into it and will raise this question on next AppKit sync

packages/appkit/src/testing/create-test-app.ts
IamGalymzhan: Fixed in #555

packages/appkit/src/cache/index.ts
```diff
@@ -557,6 +564,21 @@ export class CacheManager {
...
+   * Both fields must clear — `getInstance()` falls back to `initPromise` when
+   * `instance` is null. A pointer drop, not teardown: call {@link close} first
+   * or the old storage leaks (a `pg.Pool` under `PersistentStorage`).
+   *
+   * @internal
+   */
+  static reset(): void {
+    CacheManager.instance = null;
+    CacheManager.initPromise = null;
+    CacheManager.generation += 1;
```
MarioCadenas, line 579: what do we need this for?

packages/appkit/src/cache/index.ts
MarioCadenas, line 62: what do we need this for?

packages/appkit/src/cache/index.ts
```diff
@@ -126,9 +128,14 @@ export class CacheManager {
     }
 
     if (!CacheManager.initPromise) {
+      const generation = CacheManager.generation;
```
MarioCadenas, line 131: same for this?

packages/appkit/src/core/reset-singletons.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Drop the process-wide singletons `AppKit._createApp` initializes — called on
+ * boot to clear a previous test's leakage, and on close.
+ *
+ * A pointer drop, not teardown — close first or the old app's storage and
+ * exporters leak. A host that closes then calls `ServiceContext.get()` gets an
+ * `InitializationError`.
+ * @internal
+ */
+export function dropCoreSingletons(): void {
```
MarioCadenas, line 18: do we actually need this in the core? or could we just put this inside the testing kit?

MarioCadenas:
One bigger-picture question before the line notes. This PR is the testing kit, but it also introduces a programmatic `close()` / embeddable-app surface — `AppHandle`, `Symbol.asyncDispose`, `await using app = await createApp()`. I want to make sure we're adding that deliberately and not just because the harness reached for it.

My read is that the kit only needs to tear down the app it booted between tests, and doesn't need a public `close()`. My instinct is to keep the surface small: drop `AppKit.close()`, `[Symbol.asyncDispose]`, the `AppHandle` return type, `resetGlobalState`, and the `shutdown()`/`close()` split, and keep just `dropCoreSingletons()` + the `reset()`s.

**What I'd suggest instead of `close()`:** have the harness tear down what it booted directly — run the booted plugins' `shutdown()` hooks, close the server it started (it already holds the reference from `getServer()`), then `dropCoreSingletons()` and restore env. The one genuinely multi-boot-specific need is signal-handler hygiene: `_createApp` installs SIGTERM/SIGINT handlers on every boot and they pile up across boots in one process. I'd rather handle that by *not installing them under the harness* (a small internal boot flag) than by adding a public `close()` + `removeSignalHandlers()`. That keeps the core change to one opt-out flag rather than a whole teardown API.

That said — I might be missing a real use case. Is embeddable `close()` something we want as a product feature in its own right (hosts embedding AppKit, `await using`)? If so I'd genuinely like to hear the motivation, and I think it deserves its own PR reviewed as product surface. If it's really just serving the tests, I'd lean toward the lighter path above. What's the thinking?

---

packages/appkit/src/core/appkit.ts
```diff
@@ -246,11 +282,36 @@ export class AppKit<TPlugins extends InputPluginMap> {
...
+  async close(options: { timeoutMs?: number } = {}): Promise<void> {
+    await this.#lifecycle?.close(options);
+  }
+
+  /**
+   * Enables `await using app = await createApp(...)`, releasing the app at
+   * scope exit. Unshadowable, unlike {@link close} — a manifest name can
+   * never be a symbol.
+   */
+  async [Symbol.asyncDispose](): Promise<void> {
```
MarioCadenas, line 311: As far as I can tell this isn't used by the kit itself — `createTestApp` returns its own handle with its own `asyncDispose`, and the only thing exercising this is the test for it. Unless we're intentionally offering `await using app = await createApp()` as a feature, could we drop it? Happy to be corrected if there's a consumer I'm not seeing.

packages/appkit/src/core/appkit.ts
```diff
@@ -382,9 +443,16 @@ export async function createApp<
...
+     *
+     * Typed `PluginMap<T>`, not the `AppHandle<T>` this function returns: the
+     * runtime value is the same object, but teardown is not wired up yet, so
+     * `close()` here would silently no-op. The narrower type is deliberate.
+     */
     onPluginsReady?: (appkit: PluginMap<T>) => void | Promise<void>;
     disableInternalTelemetry?: boolean;
   } = {},
-): Promise<PluginMap<T>> {
+): Promise<AppHandle<T>> {
```
MarioCadenas, line 456: This return type changed to `AppHandle` mainly to expose `close()`, right? If we drop `close()`, I think `createApp` can go back to `PluginMap<T>` and `AppHandle` can come out (here and in `shared/plugin.ts`). The harness holds the booted instance already, so it can tear down without a public teardown method — or is there a reason to prefer the public type?

packages/appkit/src/core/lifecycle-manager.ts
```diff
@@ -110,13 +118,91 @@ export class LifecycleManager {
...
+  }
+
+  /** No `await` between read and assign — that gap is the re-entrancy window. */
+  private runOnce(): Promise<number> {
+    this.teardown ??= this.runPhases();
+    return this.teardown;
+  }
+
+  /** Run the phases and report an exit code; no process-termination concerns. */
+  private async runPhases(): Promise<number> {
```
MarioCadenas, line 185: This split — `runPhases` / `runOnce` / `close` / `removeSignalHandlers` — seems to exist so `close()` can run the phases without `process.exit`. If `close()` goes, could this fold back into `shutdown()` as it was? The one thing that looks like a real multi-boot need is `removeSignalHandlers()`, since handlers accumulate across boots — but if we just don't install them under the harness (see my top-level note), that need goes away too and the whole split can come out. Is there a non-test caller that would keep any of this alive?

packages/appkit/src/testing/reset.ts
```diff
@@ -0,0 +1,12 @@
+import { dropCoreSingletons } from "../core/reset-singletons";
+
+/**
+ * Drop the process-wide singletons `createApp()` initializes.
+ *
+ * A pointer drop, not teardown — close first or the old app's pools and
+ * exporters leak. `app.close()` does both, so this is only for tests that
+ * hand-roll `createApp`.
+ */
+export function resetGlobalState(): void {
```
MarioCadenas, line 10: I don't see any callers, and it looks like a thin rename over `dropCoreSingletons`. Could we drop the file/export and have hand-rolled-`createApp` tests call the teardown directly — or is it there to give external users a stable name? If the latter, is that a case we actually have yet?

packages/appkit/src/cache/index.ts
IamGalymzhan: This all are leftovers from idea that we could have multiple apps in one process, now established that we don't want it now, it can be removed

packages/appkit/src/cache/index.ts
IamGalymzhan: removed

packages/appkit/src/cache/index.ts
IamGalymzhan: removed

packages/appkit/src/testing/reset-singletons.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Drop the process-wide singletons `AppKit._createApp` initializes — called on
+ * boot to clear a previous test's leakage, and on close.
+ *
+ * A pointer drop, not teardown — close first or the old app's storage and
+ * exporters leak. A host that closes then calls `ServiceContext.get()` gets an
+ * `InitializationError`.
+ * @internal
+ */
+export function dropCoreSingletons(): void {
```
IamGalymzhan: moved

packages/appkit/src/core/appkit.ts
IamGalymzhan: done

packages/appkit/src/core/appkit.ts
IamGalymzhan: Removed

packages/appkit/src/testing/reset.ts
IamGalymzhan: Done

packages/appkit/src/core/lifecycle-manager.ts
IamGalymzhan: Done, kept `runPhases` as it is needed for `createTestApp().close` method

MarioCadenas (approved):

calvarjorge (approved):

PR #542, opened 2026-08-20
test(playground): smoke-test the server plugins and showcase the testing kit

IamGalymzhan:
## Stack

  - [**#540** — createTestApp and the mock client](https://github.com/databricks/appkit/pull/540)
    - [**#555** — the kit helpers (env, errors, context, cache, and suite hooks)](https://github.com/databricks/appkit/pull/555)
      - [**#542** — the playground tests](https://github.com/databricks/appkit/pull/542) ← **you are here**

Each PR targets the one above it, so the diff shown here is only the delta on top of #555. Rebased onto the current #555 tip, so the helpers it adds (`useTestApp`, the kit-faked workspace client) are available here.

---

## Summary

The playground's server plugins had **no server-side tests**. `apps/dev-playground/tests/` holds Playwright specs that intercept `/api` at the *browser* boundary (`page.route` + `route.fulfill`), so the Express server never runs in them — route handlers, `execute()` interceptors, and `executeStream` are unexercised.

Two suites cover that side with the harness from #540 — real HTTP, faked data plane, no workspace, no credentials, no network — and dogfood the kit the way the docs tell customers to test their own plugins.

### `smoke.test.ts` — the plugins boot and answer

| Test | Proves |
| --- | --- |
| all three boot together | the plugin set composes and registers under its manifest names |
| `GET /api/reconnect` | the plain JSON route answers |
| reconnect stream | opens as `text/event-stream` and emits its first message |
| `POST /api/telemetry-examples/combined` | the whole nested-span body runs against the real `TelemetryProvider` and threads `userId` through |
| `lakebase-examples` unconfigured | degrades to no routes without `PGHOST`/`LAKEBASE_ENDPOINT`, and the app still boots |

### `testing-kit.integration.test.ts` — a worked example of each recommended helper

One `describe` per helper, driven against a real plugin where the seam is clean:

| Helper | Demonstrated via |
| --- | --- |
| `getMock` (+ `responses`) | files `metadata` route → `files.getMetadata`, asserted on `app.client` |
| `createMockWorkspaceClient` | declared paths answer; undeclared resolve `undefined` |
| `createApiError` | a genuine `ApiError`, and seeded as a route failure |
| `useTestApp` | a fresh reconnect app per test |
| `withEnv` | set/restore env around a block |
| `useTestCache` / `resetTestCache` | the real `CacheManager` — stable & per-user keys, hit-then-miss |
| `createTestPlugin` | the `genie` factory → a configured instance |
| `createTestPluginContext` (+ `createMockRequest`) | cross-plugin tool dispatch, `asUser`, token-less reject |
| `expectStream` (+ `createMockResponse`) | `toEmit` / `toEmitExactly` on a bounded source |
| `useServiceContextMock` | the spied data-plane singleton |

Where no playground plugin fits hermetically, the demo mirrors the docs: `expectStream` on a bounded generator (reconnect's stream is long-lived, so `smoke.test.ts` reads it by hand), and tool dispatch straight through the context (agents needs a live model).

Adds a `dev-playground` vitest project so `pnpm test` and CI's Unit Tests job pick both suites up. The `tests/**` exclusion is load-bearing — without it vitest's default glob collects the Playwright specs, which fail on import.

---


vitest.config.ts
```diff
@@ -59,6 +59,19 @@ export default defineConfig({
           environment: "node",
         },
       },
+      {
+        plugins: [tsconfigPaths()],
+        test: {
+          name: "dev-playground",
```
calvarjorge, line 65: These suites aren't type-checked in CI: `apps/dev-playground` has no `typecheck` script (so `pnpm -r typecheck` skips it) and `pnpm test`/vitest transpiles only. A wrong kit-helper signature in these files would pass CI. Since the PR dogfoods the kit's public types, worth giving these files real type coverage.

---

apps/dev-playground/server/testing-kit.integration.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    await cache.current.getOrExecute(["report"], work, "");
+    expect(work).toHaveBeenCalledTimes(2); // cleared, so recomputed
+  });
+});
+
+describe("createTestPlugin — instantiate a factory the way production does", () => {
+  test("applies the config merge and manifest name", () => {
+    const plugin = createTestPlugin(genie, { spaces: { demo: "space-test" } });
+
+    expect(plugin.name).toBe("genie");
```
calvarjorge, line 167: Test name says "applies the config merge" but only `plugin.name` is asserted — the `spaces: { demo: "space-test" }` merge is never checked. Either assert the merged config or rename.

---

apps/dev-playground/server/testing-kit.integration.test.ts
```diff
@@ -0,0 +1,239 @@
...
+            message: "No such file",
+            errorCode: "NOT_FOUND",
+          });
+        },
+      },
+    });
+
+    const res = await app.get("/api/files/reports/metadata?path=report.csv");
+
+    expect(res.ok).toBe(false);
```
calvarjorge: Only asserts `res.ok === false`. The helper's whole point is the typed `statusCode`, so assert the route surfaces `404` here — as written it wouldn't catch a regression that collapses every error to 500.

---

apps/dev-playground/server/testing-kit.integration.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+describe("useServiceContextMock — spy the data-plane singleton in one line", () => {
+  // Spies the ServiceContext singleton directly — the seam beneath the client
+  // createTestApp injects.
+  const ctx = useServiceContextMock();
+
+  test("installs live spies over the service context", () => {
+    expect(ctx.current.getSpy).toBeDefined();
+    expect(ctx.current.createUserContextSpy).toBeDefined();
+    expect(ctx.current.createUserContextSpy).not.toHaveBeenCalled();
```
calvarjorge: This only checks the spies exist and weren't called — it never shows they actually intercept `ServiceContext.get`. Weaker than the other worked examples in this file.

---

apps/dev-playground/server/smoke.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  test("lakebase-examples degrades to no routes when Lakebase is unconfigured", async () => {
+    // Its setup() and injectRoutes() both bail on missing PGHOST/LAKEBASE_ENDPOINT.
+    // The app must still boot; the routes must simply be absent.
+    await using app = await createTestApp({
+      plugins: [lakebaseExamples()],
+      env: { PGHOST: "", LAKEBASE_ENDPOINT: "" },
+    });
+
+    expect(app.plugins["lakebase-examples"]).toBeDefined();
+    expect((await app.get("/api/lakebase-examples/raw")).status).toBe(404);
```
calvarjorge: A 404 can't distinguish "plugin degraded, route absent" from a path typo. Consider also asserting the route is present (non-404) when Lakebase is configured, to pin the meaning.

---

apps/dev-playground/server/smoke.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  test("POST /api/telemetry-examples/combined threads the userId through every span", async () => {
+    // This route really calls `fetch("https://example.com")` (its
+    // external-api span). Left alone the suite would need the internet, so
+    // non-loopback requests are stubbed — loopback must pass through, because
+    // that is how the harness reaches its own server.
+    const realFetch = globalThis.fetch;
+    vi.stubGlobal("fetch", (input: RequestInfo | URL, init?: RequestInit) => {
+      const url = String(input instanceof Request ? input.url : input);
+      return /127\.0\.0\.1|localhost/.test(url)
```
calvarjorge: Loopback detection is string-based and assumes the route uses global `fetch`. If the harness ever emits an IPv6 `[::1]` URL, the app's own loopback request gets the stubbed body and fails obscurely. Matching the app's actual base URL would be more robust.

---

apps/dev-playground/server/smoke.test.ts
```diff
@@ -0,0 +1,121 @@
...
+ * five messages three seconds apart — so asserting through it would cost ~12s
+ * for a smoke test.
+ */
+async function firstSSEPayload(res: Response): Promise<unknown> {
+  if (!res.body) throw new Error("expected a streaming body, got none");
+  const reader = res.body.getReader();
+  const decoder = new TextDecoder();
+  let buffered = "";
+  try {
+    while (!buffered.includes("\n\n")) {
```
calvarjorge, line 31: Reads only the first `\n\n` frame and expects a `data:` line in it. If the first SSE frame were ever a comment/heartbeat with no `data:`, this returns undefined and the assertion fails obliquely. Fine for the current stream — just a coupling to note.

---

apps/dev-playground/server/testing-kit.integration.test.ts
IamGalymzhan: Fixed — now asserts the typed status (`res.status === 404`), so a regression that collapses errors to 500 would be caught.

apps/dev-playground/server/testing-kit.integration.test.ts
IamGalymzhan: Strengthened — it now drives `getExecutionContext()` (which reads `ServiceContext.get()`) and asserts it returns the faked context + installed client and that the spy was invoked, so it demonstrates real interception rather than just spy presence.

apps/dev-playground/server/smoke.test.ts
IamGalymzhan: Done — added a variant that configures Lakebase via the harness env so the route actually registers, then asserts it's present (non-404) when configured and absent (404) when not, both hitting the real `/raw/health`. That also fixed a latent always-pass: the previous assertion hit `/raw`, which isn't a registered route.

apps/dev-playground/server/smoke.test.ts
IamGalymzhan: Done — the loopback passthrough now matches the app's actual base URL (`app.baseUrl`) instead of string-matching `localhost`/`127.0.0.1`, so an IPv6 `[::1]` app URL can't misroute the app's own loopback to the stub.

apps/dev-playground/server/smoke.test.ts
IamGalymzhan: Agreed — fine for the current stream; leaving as-is per your note.

vitest.config.ts
IamGalymzhan: Good catch — real gap. I'm deferring the type-check coverage to a separate follow-up rather than widening this PR: `apps/dev-playground` has no `typecheck` script today, so turning it on first surfaces some pre-existing stale-dist resolution errors to clean up. Tracked as a follow-up.

apps/dev-playground/server/testing-kit.integration.test.ts
IamGalymzhan: Fixed — the test now asserts the merged config (`spaces.demo === "space-test"`), not just `plugin.name`.

calvarjorge (approved):

PR #543, opened 2026-08-21
feat(appkit): Agent Skills v1 — SKILL.md progressive disclosure (SDK, 1/2 of #532)

MarioCadenas:
## Summary

SDK half of #532 (Agent Skills v1), split **1 of 2**. Runtime SKILL.md progressive disclosure for the `agents` plugin.

- **Skills engine** (`core/agent/skills/`): parse, load, resolve-catalog, render, read-resource, types
- **Agent-definition wiring**: `skills:` frontmatter on agent defs (`frontmatter.ts`, `load-agents.ts`, `types.ts`)
- **Plugin integration** (`plugins/agents/`): `load_skill` / `read_skill_file` built-in tools; catalog resolution (per-agent auto-visible + global opt-in + collision → qualified `agent:`/`bundle:` names); `clientConfig` skill catalog
- **Client surface**: `appkit-ui` `useAgentChat` `/skill` autocomplete
- **Docs**: `plugins/agents.md` + generated API refs

### Merge order
Merges **first** (no dependencies). PR **2/2** (playground + template fixtures) stacks on this branch and auto-retargets to `main` once this merges.

### Split from #532
Cleanly rebased onto current `main` — includes #485 (`defineManifest`), merged into `agents.ts`. Verified locally: `pnpm --filter=@databricks/appkit typecheck` → exit 0, 0 errors.

25 files: appkit + appkit-ui source/tests + docs.

PR #544, opened 2026-08-21
feat(appkit): Agent Skills v1 — SKILL.md progressive disclosure + demo (#532)

MarioCadenas:
## Summary

Playground/template half of #532 (Agent Skills v1), split **2 of 2**. Stacked on #543.

Demo skills exercising every source and case:
- **Global + opt-in**: `skills/bullet-brief` (helper opts in → multi-entry menu on a code agent)
- **Per-agent source**: `query/skills/routing-brief` (+ bundled `reference.md`, `allowed-tools`)
- **Collision**: per-agent `query/skills/haiku` vs global `haiku` → `agent:haiku` / `bundle:haiku`
- **Template**: `tracer-bullets` skill + planner opt-in
- `/agent` page picker gains `query` so its skills surface in the input

### Merge order
Depends on **#543** (2/2). Base is `split/532/sdk-feature`; GitHub auto-retargets this to `main` once #543 merges.

14 files: `apps/dev-playground/*` + `template/*`.

### DIFF
--- a/packages/appkit-ui/src/react/hooks/use-agent-chat.ts
+++ b/packages/appkit-ui/src/react/hooks/use-agent-chat.ts
@@ -85,8 +85,13 @@ export interface UseAgentChatResult {
   /**
    * Send a user turn and stream the response. Aborts any in-flight
    * stream. Resolves when the stream completes (success or error).
+   *
+   * Pass `opts.skill` to force-load a skill for this turn, or prefix the
+   * message with `/skill-name` as sugar (the leading token is parsed off and
+   * sent as the skill; the rest becomes the message). An explicit
+   * `opts.skill` wins over a `/`-prefix.
    */
-  send: (message: string) => Promise<void>;
+  send: (message: string, opts?: { skill?: string }) => Promise<void>;
   /**
    * Discard accumulated content, events, and threadId. Aborts any
    * in-flight stream. Use when switching agents or starting a fresh
@@ -158,7 +163,7 @@ export function useAgentChat({
   }, []);
 
   const send = useCallback(
-    async (message: string) => {
+    async (message: string, opts?: { skill?: string }) => {
       // Abort any previous stream — only one chat turn in flight per hook.
       abortControllerRef.current?.abort();
       const controller = new AbortController();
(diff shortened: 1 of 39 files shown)


MarioCadenas:
Collapsed into #547 (the playground + template skills fixtures are now included there). Closing to keep the skills work as one squash-mergeable PR during my vacation. Commits preserved in #547.

packages/appkit-ui/src/react/hooks/use-agent-chat.ts
```diff
@@ -95,6 +100,20 @@ export interface UseAgentChatResult {
   reset: () => void;
 }
 
+/** `forced` (opts.skill) wins; otherwise parse a leading `/skill-name` token off the message. */
+function resolveSkill(
+  message: string,
+  forced?: string,
+): { skill?: string; text: string } {
+  if (forced) return { skill: forced, text: message };
+  const match = message.match(/^\/([A-Za-z0-9][\w.:-]*)\s*/);
```
atilafassina, line 126: I'm afraid this match may be a bit naïve. 
If read this right, the agent may mangle some feedback to the user and messages with leading slashes will have the first word cut out.

Can we instead of a RegEx match, compare against the skills catalog directly?
```

┌─────────────────────────────────┬───────────────────┬─────────────────────────────┬─────────────────────────────┐
│            You type             │ Parser B extracts │ Message the model receives  │         Skill sent          │
├─────────────────────────────────┼───────────────────┼─────────────────────────────┼─────────────────────────────┤
│ /routing-brief summarize this   │ routing-brief     │ summarize this              │ routing-brief ✅ (intended) │
├─────────────────────────────────┼───────────────────┼─────────────────────────────┼─────────────────────────────┤
│ /tmp is full                    │ tmp               │ is full                     │ tmp ❌                      │
├─────────────────────────────────┼───────────────────┼─────────────────────────────┼─────────────────────────────┤
│ /usr/bin/python needs upgrading │ usr               │ /bin/python needs upgrading │ usr ❌                      │
├─────────────────────────────────┼───────────────────┼─────────────────────────────┼─────────────────────────────┤
│ /api/v1/query returns 500       │ api               │ /v1/query returns 500       │ api ❌                      │
└─────────────────────────────────┴───────────────────┴─────────────────────────────┴─────────────────────────────┘
```

MarioCadenas: Good catch — fixed in 86916dcb.

You're right that the bare regex mangles any leading `/word`. I went with your suggestion: the leading `/token` is now treated as a skill **only when it names a skill in the agent's catalog**, otherwise the message is sent verbatim. So `/tmp is full`, `/usr/bin/python needs upgrading`, and `/api/v1/query returns 500` all pass through untouched, while `/routing-brief …` (and qualified collision names like `/agent:haiku …`) still resolve.

The catalog is already shipped to the client via `getPluginClientConfig("agents").skills[agent]`, so no new data flow. Changes:
- `useAgentChat` gains an optional `skills` option; `resolveSkill` returns the message verbatim unless the token ∈ `skills`.
- The template passes `skills: activeSkills.map(s => s.name)` into the hook.
- The playground route (which has its own inline parser, not the hook) gets the same catalog guard.

Added hook tests covering the false-positive cases (`/tmp is full` verbatim, `/usr/bin/python …` verbatim, `opts.skill` still wins over an unknown `/token`).

---

atilafassina (approved):

PR #547, opened 2026-08-21
feat(appkit): split the agents plugin into focused modules (#532)

MarioCadenas:
## Summary

Behavior-preserving refactor of the ~2,500-line `agents.ts` god-file into focused modules. Stacked on **#543** (it only touches `agents.ts`, which #543 owns) — independent of #544.

No behavior changes: every commit keeps `pnpm --filter=@databricks/appkit typecheck` and the agent test suite (**394 tests**) green.

**`agents.ts`: 2,512 → 1,784 lines (−728, ~29%).** The plugin is now an orchestrator (registry wiring + route handlers + SSE plumbing) delegating cohesive concerns to focused sibling modules.

### What moved
- **Pure helpers** → `approval.ts`, `prompt.ts`, `builtin-tools.ts`, `adapter-extensions.ts`.
- **Skill loading/dispatch** → `skill-loader.ts` (discovery, catalog resolution, `load_skill`/`read_skill_file`).
- **Registry assembly** → `registry.ts` (`loadCodeAgents`, `hasCodeAgentSources`, `resolveDefaultAgent`, `AgentSource`).
- **Tool-dispatch engine** → `tool-dispatch.ts` (`RunState` + `dispatchToolCall` + `runSubAgent`) as free functions over a `ToolDispatchDeps` object.
- **Config resolution** → `resolve-config.ts` (`resolveApprovalPolicy` / `resolveLimits` pure fns; getters keep the memo cache).
- **Stream tracking** → `ActiveStreamTracker` (active-stream map + per-user O(1) concurrency counter). Named to avoid clashing with the unrelated SSE-layer `StreamRegistry` in `src/stream/`.

### Notes
- Delegator pattern keeps call sites stable throughout; `skillWorkspaceClient()` stays as the OBO credential seam.
- Deliberately **not** extracted: `buildToolIndex` (owns the mutable `mcpClient` lifecycle) and the HTTP handlers (the `Plugin` route contract) — both need 5-8 pieces of plugin state, so extraction would widen interfaces more than it'd clarify (fails the deep-module test). The remaining ~1,784 lines are the plugin genuinely being a plugin.
- The agents-plugin concurrency tracker is intentionally separate from the SSE `StreamRegistry`: that one is a global, transport-level store (event replay buffers, global cap) with no per-user dimension; this one counts streams per user for `maxConcurrentStreamsPerUser` + ownership checks.
- Tests updated where they white-boxed moved internals (`dispatch-tool-call`, `dos-limits`, `approval-route`).

### Merge order
Stacked on #543; auto-retargets to `main` once #543 merges. Merge after #543.


packages/appkit/src/plugins/agents/skill-loader.ts
```diff
@@ -0,0 +1,244 @@
...
+      "Failed to list catalog skills volume '%s': %s",
+      volume,
+      err instanceof Error ? err.message : String(err),
+    );
+    return [];
+  }
+
+  // Read every skill concurrently — each is a network round-trip to the
+  // volume, so serial reads would add up. A per-skill failure skips only
+  // that skill; Promise.all preserves the (sorted) entry order.
```
atilafassina, line 91: I see. So this is why `readVolumeSkill` is not recursive while local skills are?

I'm fine with that. Just double-checking. ✅

atilafassina (approved):
it looks like this PR adds new behavior. I think you should consider changing from `refactor` to `feat`

packages/appkit/src/plugins/agents/skill-loader.ts
MarioCadenas: yeah, I think for now we can keep it as is and see

PR #555, opened 2026-08-27
feat(appkit): add testing-kit helpers for env, errors, context, and cache

IamGalymzhan:
## Stack

  - [**#540** — createTestApp and the mock client](https://github.com/databricks/appkit/pull/540)
    - [**#555** — env, error, context, and cache test helpers](https://github.com/databricks/appkit/pull/555) ← **you are here**
      - [**#542** — the playground smoke tests](https://github.com/databricks/appkit/pull/542)

Each PR targets the one above it, so the diff here is only the delta on top of #540. Review in order.

---

## Summary

Adds customer-facing helpers to `@databricks/appkit/testing` so plugin tests assert business logic instead of hand-rolled boilerplate, then adopts them across the existing suites.

**New helpers**

- **`withEnv(vars, fn)`** — set env for the duration of `fn`, then restore each key's prior state (deleting keys that were unset before). Sync and async forms; nested calls restore LIFO. Replaces the `process.env.X = …; try/finally; delete` pattern, which wrongly drops pre-existing env.
- **`createTestPluginContext(fakes, { responses, env, strict })`** — a synchronous options parameter that composes a plugin unit-test in one call: a mock workspace client from `responses`, an auto-restored service context wired to it, and scoped `env`. Non-breaking — calls with no second argument are unchanged.
- **`createApiError({ statusCode, message, errorCode })`** — a genuine `ApiError` (so `error instanceof ApiError` holds) for error-path tests.
- **`useTestCache()`** — boots AppKit's real in-memory cache, clears it before each test, and returns the real `CacheManager`, so a plugin's caching path runs through production's own `getOrExecute`/`generateKey` instead of a mocked `cache` module. `CacheManager` stays a process singleton.

**Adoption**

- The analytics, metric, and ai-search suites drop their `vi.mock("cache")` fakes and assert against the real cache via `useTestCache()`.
- 17 passthrough suites share a new internal `createCacheMock()` in place of a copy-pasted fake.
- `withEnv` replaces 11 hand-rolled env blocks, and `createMockRouter` is surfaced in the testing guide.

The internal-mock migration and the singleton-reset changes stay deferred to a later PR.

**Stacked on #540** (`feat/testing-kit-harness`) — these helpers extend testing-kit code on that branch, so this targets it rather than `main`.


packages/appkit/src/testing/fixtures.ts
```diff
@@ -340,6 +341,102 @@ export function setupDatabricksEnv(overrides: Record<string, string> = {}) {
   Object.assign(process.env, overrides);
 }
 
+/**
+ * Sets environment variables for the duration of `fn`, then restores them to
```
calvarjorge: This `withEnv` JSDoc (the block with the sync/async `@example`) is stranded above `applyEnv`. Two doc comments stack here, so `applyEnv` binds the closer second block and this first one is ignored, while the public `withEnv` export below (preceded by `restoreQuietly`) is left with **no doc comment** — IDE hover and TypeDoc will show nothing for it.

Suggest moving this block down to directly above `export function withEnv` so the public helper is the one documented.

IamGalymzhan: Fixed

calvarjorge (approved):

PR #572, opened 2026-09-03
feat(appkit): thread the app cache through PluginContext

IamGalymzhan:
## Stack

- [**#572** — thread the cache through PluginContext](https://github.com/databricks/appkit/pull/572) ← **you are here**
  - [**#573** — give each app its own CacheManager](https://github.com/databricks/appkit/pull/573)

#573 stacks on top of this one and holds the behavioural change — review this first.

## Summary

Threads the app's `CacheManager` through `PluginContext`. `createApp` injects the manager it builds, and a plugin reads it from its context as `this.cache`, falling back to the process-wide manager when a context carries none.

Behaviour is unchanged — the injected cache is still the singleton, so this is a no-op for existing apps. It's the groundwork for #573, which gives each app its own manager and removes the singleton; kept separate so that change is a focused review rather than a 60-file one.

## What changes

- `PluginContext` gains an optional, `readonly` `cache`; `createApp` passes the manager into it.
- `Plugin` binds `this.cache` from its context first, then the singleton.
- `CacheManager.getInstance` / `getInstanceSync` are untouched; no consumer or test changes.

## Verification

- `pnpm test` — 4659 passing, 1 skipped
- `pnpm -r typecheck`, `pnpm check` — clean

### DIFF
--- a/packages/appkit/src/core/appkit.ts
+++ b/packages/appkit/src/core/appkit.ts
@@ -32,10 +32,16 @@ export class AppKit<TPlugins extends InputPluginMap> {
   #setupPromises: Promise<void>[] = [];
   #context: PluginContext;
 
-  private constructor(config: { plugins: TPlugins }) {
+  /**
+   * @param context - The app's plugin context, carrying this app's cache. Built
+   *   in `_createApp` after the cache so it can be injected here, rather than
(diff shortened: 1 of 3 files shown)


IamGalymzhan:
Closing — reworking the split.

PR #573, opened 2026-09-03
feat(appkit): give each app its own CacheManager

IamGalymzhan:
## Stack

- [**#572** — thread the cache through PluginContext](https://github.com/databricks/appkit/pull/572)
  - [**#573** — give each app its own CacheManager](https://github.com/databricks/appkit/pull/573) ← **you are here**

This targets #572, so the diff shown here is only the delta on top of it. Review #572 first.

## What

Each app now owns its `CacheManager`. Building on the injection from #572, `createApp` builds a manager per app with `create()`, and the process-wide singleton is gone. Two apps in one process hold independent managers, each honouring its own `cache` config.

`getInstance` / `getInstanceSync` are removed; the constructor is private behind the `create` (async, the app's path) and `forStorage` (sync, the testing kit) factories; `PluginContext.cache` is required; and `Plugin.cache` is a read-only getter that throws when a plugin was never attached to an app. `LifecycleManager` takes the app's manager by injection so shutdown closes the right one, and `close()` only closes storage the manager built.

Ships as a minor by explicit agreement (no `!`, no `BREAKING CHANGE:`).

## What changes for a consumer

Most apps need nothing — if your plugins read `this.cache` and you build with `createApp`, this is a no-op.

- `CacheManager.getInstance()` / `getInstanceSync()` are gone. Inside a plugin, use `this.cache`.
- `this.cache` is read-only — set a per-plugin `cache: { enabled, ttl }` config instead of assigning a manager.
- An unattached plugin has no cache: reading `this.cache` before the app binds it throws a named `InitializationError`.

Upgrade notes are in `docs/plugins/caching.md`.

## Test migration

The 26 suites that mocked the cache module — to keep one test's cached value out of the next under the singleton — now run on a real cache from the testing kit. Each migration was verified by mutation (delete the behaviour under test, confirm the assertion fails), so a detached handle can't leave a suite green while asserting nothing.

## Deferred

Two pool-leak fixes surfaced along the way — close the cache on a failed boot, and end the Lakebase pool when a healthy cache fails to initialize — are held for their own follow-up, to keep this PR to the refactor.

## Verification

- `pnpm test` — 4690 passing, 1 skipped
- `pnpm -r typecheck`, `pnpm check` — clean

### DIFF
--- a/apps/dev-playground/package.json
+++ b/apps/dev-playground/package.json
@@ -13,6 +13,7 @@
     "install": "cd client && npm install && cd ..",
     "preview": "vite preview",
     "check": "tsc",
+    "typecheck": "tsc --noEmit",
     "clean": "rm -rf build && cd client && rm -rf dist",
     "clean:full": "rm -rf build node_modules && cd client && rm -rf dist node_modules",
     "test:integration": "playwright test",
--- a/apps/dev-playground/server/index.ts
+++ b/apps/dev-playground/server/index.ts
@@ -375,10 +375,13 @@ createApp({
           const pngs = new Map<string, (typeof entries)[number]>();
           const metas = new Map<string, (typeof entries)[number]>();
           for (const e of entries) {
-            if (e.path.endsWith(".png")) {
-              pngs.set(e.path.replace(/\.png$/, ""), e);
-            } else if (e.path.endsWith(".json")) {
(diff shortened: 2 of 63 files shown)


IamGalymzhan:
Closing — reworking the split.

PR #582, opened 2026-09-11
fix(appkit): allow database initialization without arguments

ditadi:
## Summary

Allow `database()` and `database({})` to initialize without explicit configuration.

- Load the named `schema` export from `config/database/schema.ts`; ship a valid empty starter in newly scaffolded apps.
- Resolve the PostgreSQL username from the application's Databricks credentials when `PGUSER` and `DATABRICKS_CLIENT_ID` are absent, using the same client for identity lookup and pool authentication.
- Preserve explicit overrides and validate the schema before pool creation.
- Check the packaged template's database starter before uploading the PR artifact.

## Validation

- 4,983 tests passed, 1 skipped; build, docs, lint and typecheck passed.
- Created a fresh app with the actual Databricks CLI and prepared SDK tarballs.
- Started the unmodified generated server and received HTTP 200.
- Initialized `database()` and queried real local PostgreSQL without `PGUSER` or `DATABRICKS_CLIENT_ID`; the workspace identity/credential API was a local test double, not a live Lakebase workspace.


template/config/database/schema.ts
```diff
@@ -0,0 +1,5 @@
+import { defineSchema } from '@databricks/appkit/beta';
```
MarioCadenas, line 2: at the top of this we should put 
{{if .plugins.database -}}


and 

{{- end}} at the end of the file so it gets removed in case the plugin is not chosen

MarioCadenas (approved):

template/config/database/schema.ts
ditadi: Fixed in 231c01bd. The entire schema file is now guarded by `.plugins.database`. Updated the unit and artifact checks, verified both branches with Go templates, and confirmed that the actual Databricks CLI omits the file when database is not selected.

Issue #549, opened 2026-08-24
Support per-query cache policies for file-based analytics queries


jefbarn:
## Problem

AppKit's analytics plugin caches identical file-based SQL queries for one hour by default. Some queries are intentionally non-deterministic or calculate against current data and must execute every time the user submits them.

In AppKit `0.65.0`, `useAnalyticsQuery` supports:

- `format`
- `maxParametersSize`
- `autoStart`

There is no per-query cache option. Caching can be disabled globally with:

```ts
createApp({
  cache: {
    enabled: false,
  },
});
```

However, this disables caching for every AppKit plugin and analytics query in the application.

## Example

```sql
-- config/queries/fresh_value.sql
SELECT current_timestamp() AS calculated_at
```

```tsx
const { data } = useAnalyticsQuery('fresh_value');
```

After unmounting and remounting the component, the same cached timestamp is returned for up to an hour instead of executing the SQL again.

A more significant use case is a parameterized, non-deterministic table-valued function:

```sql
SELECT result_json
FROM IDENTIFIER(:functionName)(
  :input,
  :assumption
)
```

Submitting the same parameters again should deliberately recalculate against current source data.

## Current workaround

We add a unique nonce to every submission:

```sql
-- @param queryNonce STRING
SELECT ...
WHERE length(:queryNonce) > 0
```

```tsx
const [queryNonce] = useState(() => crypto.randomUUID());

const parameters = useMemo(
  () => ({
    input: sql.string(input),
    queryNonce: sql.string(queryNonce),
  }),
  [input, queryNonce]
);
```

This works because parameters are part of the cache key, but it pollutes the SQL contract and creates cache entries that can never be reused.

## Requested capability

Please support a server-controlled cache policy for individual file-based queries.

For example, a query directive:

```sql
-- @cache disabled
SELECT ...
```

Or configurable TTL:

```sql
-- @cache ttl=60
SELECT ...
```

Alternatively, plugin configuration could define the policy:

```ts
analytics({
  queries: {
    fresh_value: {
      cache: false,
    },
    summary: {
      cache: { ttl: 60 },
    },
  },
});
```

A client-side option such as the following would also be useful, although a server-owned policy would prevent clients from bypassing caching for arbitrary expensive queries:

```tsx
useAnalyticsQuery('fresh_value', parameters, {
  cache: 'no-store',
});
```

## Expected behavior

- Existing queries remain cached by default.
- An individual query can disable caching or select a custom TTL.
[...]

PR #584, opened 2026-09-14
feat(analytics): internal refetch + uncached request mode (poll 1/4)

atilafassina:
**Stacked PR 1 of 4** — analytics query polling (`useAnalyticsQuery({ poll })`).
Base: `main`. This is the bottom of the stack; review/merge first.

## What
Surface the analytics request store's existing settled-key re-run as an internal `refetch(cacheKey)`, and add a force-uncached execution mode so a re-run with unchanged params genuinely re-hits the warehouse.

- `analytics-request-store.ts`: internal `refetch(cacheKey)` marks the key for uncached execution (per-run, consumed once) then calls the generic `store.start(cacheKey)`.
- Uncached path adds a `skipCache: true` field to `IAnalyticsQueryRequest`; the analytics route honors it to bypass its TTL cache (works for JSON_ARRAY and ARROW_STREAM). Backward-compatible (optional, defaults false).

## Dormant
No public API, no scheduler. Consumed only by store unit tests. Nothing observable changes yet.

## Stack
1. **← this PR** internal refetch + uncached mode
2. internal `usePoll` scheduler
3. `useAnalyticsQuery({ poll })` binding + public types (behavior-adding)
4. docs

## Tests
Full suite green; store unit tests assert deferred-start + `refetch` fresh settle on unchanged params, and `skipCache` on the outgoing request.


packages/appkit/src/plugins/analytics/analytics.ts
```diff
@@ -1002,12 +1023,13 @@ export class AnalyticsPlugin extends Plugin implements ToolProvider {
...
     parameters: IAnalyticsQueryRequest["parameters"],
     executorKey: string,
+    skipCache: boolean = false,
   ): QueryExecutor {
     const hashedQuery = this.queryProcessor.hashQuery(query);
     const cache = this.cache;
     const ttl = queryDefaults.cache?.ttl;
     return {
-      query: (q, params, formatParameters, signal) => {
+      query: async (q, params, formatParameters, signal) => {
```
MarioCadenas, line 1032: why is this async now? we are not doing any async operation inside no?

packages/appkit/src/plugins/analytics/types.ts
```diff
@@ -125,6 +125,13 @@ export function normalizeAnalyticsFormat(
 export interface IAnalyticsQueryRequest {
   parameters?: Record<string, any>;
   format?: AnalyticsFormat;
+  /**
+   * Internal flag to bypass server-side TTL cache for this request.
+   * When true, the request re-hits the warehouse even with identical params.
+   * Used by polling/refetch operations to force fresh execution.
+   * @internal
+   */
+  skipCache?: boolean;
```
MarioCadenas, line 134: given that we have this issue https://github.com/databricks/appkit/issues/549 should we make this not internal? and also allow the hook in the UI to pass the skip?

packages/appkit-ui/src/react/hooks/analytics-request-store.ts
```diff
@@ -162,15 +164,34 @@ async function fetchArrowDirect(
...
+    if (uncachedKeys.has(cacheKey)) {
+      uncachedKeys.delete(cacheKey);
+    }
+
+    // Build the request payload, injecting skipCache when needed.
+    let requestPayload = options.payload;
+    if (shouldSkipCache) {
+      try {
+        const parsed = JSON.parse(options.payload);
+        requestPayload = JSON.stringify({ ...parsed, skipCache: true });
```
MarioCadenas, line 184: this exploded my mind a bit, don't we have skipCache in the options? why do we need to parse and stringify again?

